PR #12, opened 2017-12-01
Docker container updates

childsb:


deploy/kubernetes/statefulset.yaml
```diff
@@ -0,0 +1,93 @@
...
+  name: csi-provisioner
+
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-provisioner-runner
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
```
jsafrane: should the provisioner be able to get/list/watch PVCs?

deploy/kubernetes/statefulset.yaml
```diff
@@ -0,0 +1,93 @@
...
+    verbs: ["get", "list", "watch", "update"]
```
jsafrane: And create and delete PVs?

PR #25, opened 2018-01-11
Improved build process and consistency

davidz627:
Improved build to only rebuild when source files changed. Changed output directory to bin/ for consistency with external-attacher and driver-registrar.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ COPY ./examples/flex-debug.sh /opt/csi-provisioner/
 
 # install the go kube piece of provisioner
 
-COPY ./_output/csi-provisioner /opt/csi-provisioner/
+COPY ./bin/csi-provisioner /opt/csi-provisioner/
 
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -12,7 +12,7 @@ COPY ./examples/flex-debug.sh /opt/csi-provisioner/
 
 # install the go kube piece of provisioner
 
-COPY ./_output/csi-provisioner /opt/csi-provisioner/
+COPY ./bin/csi-provisioner /opt/csi-provisioner/
```
rootfs: let's make _output/bin for binaries. Change here and in Makefile and README too.

jsafrane: It should be the same as attacher, which uses `bin/` too.

childsb: I do not like bin/ for build artifacts.... It's a common directory for scripts or other real binaries that might be part of the project..  _output/bin would be fine.

davidz627:
Closing due to inactivity and staleness.

PR #33, opened 2018-01-26
Volume Naming Prefix and UUID Length Parameters

cduchesne:
This patch makes change to volume naming so that both prefix and uuid length are configurable. Some storage systems have max limits to volume naming, so making these fields configurable allows much shorter names if necessary. It is also helpful to use different prefixes to make storage administration easier.

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -35,11 +35,13 @@ import (
 )
 
 var (
-	provisioner       = flag.String("provisioner", "", "Name of the provisioner. The provisioner will only provision volumes for claims that request a StorageClass with a provisioner field set equal to this name.")
-	master            = flag.String("master", "", "Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.")
-	kubeconfig        = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Either this or master needs to be set if the provisioner is being run out of cluster.")
-	csiEndpoint       = flag.String("csi-address", "", "The gRPC endpoint for Target CSI Volume")
-	connectionTimeout = flag.Duration("connection-timeout", 10*time.Second, "Timeout for waiting for CSI driver socket.")
+	provisioner          = flag.String("provisioner", "", "Name of the provisioner. The provisioner will only provision volumes for claims that request a StorageClass with a provisioner field set equal to this name.")
+	master               = flag.String("master", "", "Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.")
+	kubeconfig           = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Either this or master needs to be set if the provisioner is being run out of cluster.")
+	csiEndpoint          = flag.String("csi-address", "", "The gRPC endpoint for Target CSI Volume")
+	connectionTimeout    = flag.Duration("connection-timeout", 10*time.Second, "Timeout for waiting for CSI driver socket.")
+	volumeNamePrefix     = flag.String("volume-name-prefix", "kubernetes-dynamic-pv", "Prefix to apply to the name of a created volume")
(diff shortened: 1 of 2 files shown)


cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -35,11 +35,13 @@ import (
...
-	connectionTimeout = flag.Duration("connection-timeout", 10*time.Second, "Timeout for waiting for CSI driver socket.")
+	provisioner          = flag.String("provisioner", "", "Name of the provisioner. The provisioner will only provision volumes for claims that request a StorageClass with a provisioner field set equal to this name.")
+	master               = flag.String("master", "", "Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.")
+	kubeconfig           = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Either this or master needs to be set if the provisioner is being run out of cluster.")
+	csiEndpoint          = flag.String("csi-address", "", "The gRPC endpoint for Target CSI Volume")
+	connectionTimeout    = flag.Duration("connection-timeout", 10*time.Second, "Timeout for waiting for CSI driver socket.")
+	volumeNamePrefix     = flag.String("volume-name-prefix", "kubernetes-dynamic-pv", "Prefix to apply to the name of a created volume")
+	volumeNameUUIDLength = flag.Int64("volume-name-uuid-length", 16, "Length in characters for the generated uuid of a created volume")
```
vladimirvivien: @cduchesne consider just calling this variable `VolumeNameLength` and also set it to just a type `int`

cduchesne: Changed to int. For some reason I thought there was a dependency on int64 but I was wrong.
As for VolumeNameLength, it wouldn't be the actual volume name length. It is the length of the UUID portion of the VolumeName. I could change the logic to subtract the length of whatever prefix is provided + the dash. This might still be confusing though.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -35,11 +35,13 @@ import (
...
-	csiEndpoint       = flag.String("csi-address", "", "The gRPC endpoint for Target CSI Volume")
-	connectionTimeout = flag.Duration("connection-timeout", 10*time.Second, "Timeout for waiting for CSI driver socket.")
+	provisioner          = flag.String("provisioner", "", "Name of the provisioner. The provisioner will only provision volumes for claims that request a StorageClass with a provisioner field set equal to this name.")
+	master               = flag.String("master", "", "Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.")
+	kubeconfig           = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Either this or master needs to be set if the provisioner is being run out of cluster.")
+	csiEndpoint          = flag.String("csi-address", "", "The gRPC endpoint for Target CSI Volume")
+	connectionTimeout    = flag.Duration("connection-timeout", 10*time.Second, "Timeout for waiting for CSI driver socket.")
+	volumeNamePrefix     = flag.String("volume-name-prefix", "kubernetes-dynamic-pv", "Prefix to apply to the name of a created volume")
```
childsb: I think the default should be blank (no prefix)

cduchesne: This would result in a volume being named -<uuid>. I simply set the default to match what was existing already today for the most part.

cduchesne: Personally I would make the default something like `k8s-csi-pv` to keep it shorter. This would result in a volume `k8s-csi-pv-[uuid]`

childsb: Ok, I see the default prefix now and this is using the same default as before, so this is fine.

cduchesne: Ok cool

childsb:
This LGTM.  If you rebase I will merge.

cduchesne:
Done

cmd/csi-provisioner/csi-provisioner.go
vladimirvivien: cool thanks for clarifying.

PR #49, opened 2018-03-02
Bugfix: PV should use capacity bytes returned by plugin, not PVC bytes

davidz627:
PV currently just passes capacity bytes through from the PVC. This might differ from the number of bytes actually provisioned by the plugin. The Plugin response capacity should be used as the source of truth for the PV Capacity.

Fixes: #48

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -23,6 +23,9 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+	_ "k8s.io/apimachinery/pkg/util/json"
+
 	"github.com/golang/glog"
 
 	"github.com/kubernetes-incubator/external-storage/lib/controller"
@@ -289,7 +292,6 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 		},
 		CapacityRange: &csi.CapacityRange{
 			RequiredBytes: int64(volSizeBytes),
-			LimitBytes:    int64(volSizeBytes),
 		},
 	}
 	rep := &csi.CreateVolumeResponse{}
@@ -336,6 +338,28 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 	for k, v := range rep.Volume.Attributes {
 		volumeAttributes[k] = v
 	}
+	respCap := rep.GetVolume().GetCapacityBytes()
+	if respCap < volSizeBytes {
+		capErr := fmt.Errorf("created volume capacity %v less than requested capacity %v", respCap, volSizeBytes)
+		delReq := &csi.DeleteVolumeRequest{
+			VolumeId: rep.GetVolume().GetId(),
+		}
+		if options.Parameters != nil {
+			credentials, err := getCredentialsFromParameters(p.client, options.Parameters)
+			if err != nil {
+				return nil, err
+			}
+			delReq.ControllerDeleteSecrets = credentials
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+		defer cancel()
+		_, err := p.csiClient.DeleteVolume(ctx, delReq)
+		if err != nil {
+			capErr = fmt.Errorf("%v. Cleanup of volume %s failed, volume is orphaned: %v", capErr, share, err)
+		}
+		return nil, capErr
+	}
+	repBytesString := fmt.Sprintf("%v", respCap)
 	pv := &v1.PersistentVolume{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: share,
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -250,7 +253,7 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 			PersistentVolumeReclaimPolicy: options.PersistentVolumeReclaimPolicy,
 			AccessModes:                   options.PVC.Spec.AccessModes,
 			Capacity: v1.ResourceList{
-				v1.ResourceName(v1.ResourceStorage): options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)],
+				v1.ResourceName(v1.ResourceStorage): resource.MustParse(repBytesString),
```
sbezverk: I think we should not provide this value in bytes. I checked already existing PV and it displays the value rounded up to Gigabytes.

[root@kube-3 ~]# kubectl get pv -o yaml kubernetes-dynamic-pv-8ecdbbbd1e2311e8
apiVersion: v1
kind: PersistentVolume
  capacity:
    storage: 1Gi

davidz627:
It is a quantity type so I am just saving it at the finest granularity.
When you view it on the pv if it is 5gb it will say 5Gi not 50000000bytes.

pkg/controller/controller.go
sbezverk: Ok,  sounds good.

sbezverk:
@davidz627 I agree with what you propose, the only thing bugs me is that all other "in-tree" plugins are assigning ResourceStorage from PVC and not based on what plugin returns. It would be great to get it reviewed by @rootfs.

pkg/controller/controller.go
rootfs: at least we need to validate the bytes are no less than what is claimed in the PVC.

sbezverk: @rootfs If number of bytes returned by the driver is less than in PVC's request, we fail, right?

lpabon:
@davidz627 What is the status of this PR? Is there a change needed?

pkg/controller/controller.go
davidz627: I believe we should throw an error, I will add that in

davidz627:
Small change, rebase, then should be ready to merge

davidz627:
@lpabon @sbezverk @rootfs PTAL, hope it works for you

pkg/controller/controller.go
```diff
@@ -265,7 +268,6 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 		},
 		CapacityRange: &csi.CapacityRange{
 			RequiredBytes: int64(volSizeBytes),
-			LimitBytes:    int64(volSizeBytes),
```
davidz627: I've also removed `LimitBytes`. Rationale is that some storage plugins may have a min volume size that is larger than `RequiredBytes` and current Kubernetes behavior is to allow the greater sized volume to be created and bound to the requested PVC. This will align the behavior with existing Kubernetes behavior. Additionally, very few storage systems are able to give back a volume with the *exact* amount of bytes requested, it might have to round up a little (or a lot), and we should not fail in those use cases.

pkg/controller/controller.go
```diff
@@ -291,6 +293,11 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 	for k, v := range rep.Volume.Attributes {
 		volumeAttributes[k] = v
 	}
+	respCap := rep.GetVolume().GetCapacityBytes()
+	if respCap < volSizeBytes {
+		return nil, fmt.Errorf("created volume capacity %v less than requested capacity %v", respCap, volSizeBytes)
```
lpabon: This leaves a zombied volume. All returns after line 289 must delete the volume on error to clean up.

davidz627: Added logic to delete volume as well as unit test to check this behavior. We should be able to use the method I used to unit test this functionality for more unit tests on the Provisioner in the future (there are currently none)

davidz627:
@lpabon PTAL

davidz627:
@lpabon I have made the requested change and wrote a test for this case

pkg/controller/controller.go
```diff
@@ -336,6 +338,28 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
+			if err != nil {
+				return nil, err
+			}
+			delReq.ControllerDeleteSecrets = credentials
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+		defer cancel()
+		_, err := p.csiClient.DeleteVolume(ctx, delReq)
+		if err != nil {
+			capErr = fmt.Errorf("%v. Cleanup of volume %s failed, volume is orphaned: %v", capErr, share, err)
```
lpabon: 👍

PR #56, opened 2018-03-13
PluginCapability update for external-provisioner

NickrenREN:
PluginCapability update for external-provisioner

@jsafrane @saad-ali


saad-ali:
Thanks @NickrenREN!

Issue #55, opened 2018-03-12
Errors from fetching secrets should be reported


jsafrane:
The provisioner only logs error when it fails to fetch a secret configured in StorageClass and continues with provisioning. It should return error to the user instead.

Secrets are optional in CSI, however admin really wants them to be used once he configures them in StorageClass. If there is a typo in secret name, admin should know and provisioning should fail. Nobody knows what a CSI driver will do with no secret, it may provision a volume that's accounted to another team / department than the admin wants to.

saad-ali:
Seems reasonable to me.
This will need to be changed on kubernetes as well. @jsafrane could you open an issue to track this.

PR #58, opened 2018-03-14
Errors from fetching secrets should be reported

wackxu:
fix https://github.com/kubernetes-csi/external-provisioner/issues/55
/assign @jsafrane

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -249,7 +249,11 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 	}
 	secret := v1.SecretReference{}
 	if options.Parameters != nil {
-		req.ControllerCreateSecrets = getCredentialsFromParameters(p.client, options.Parameters)
+		credentials, err := getCredentialsFromParameters(p.client, options.Parameters)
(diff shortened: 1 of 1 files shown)


pohly:
How does the report look like? I.e., can a developer who has misconfigured the secret understand what he did wrong?

Should this have a corresponding test somewhere? That might be a good place to show the expected error message and test for it at the same time.

jsafrane:
lgtm, can you please rebase?

@pohly, error is going to be reported as event on the PVC, visible e.g. in `kubectl describe pvc`.

pohly:
On Wed, 2018-03-14 at 15:55 +0000, Jan Šafránek wrote:
> @pohly, error is going to be reported as event on the PVC, visible
> e.g. in kubectl describe pvc.

That's where I'd expect it to show up. I was more wondering whether the
 error text is something that can be understood and traced back to the
root cause or whether it is something obscure.

jsafrane:
the PR uses `failed to find the secret %s in the namespace %s with error: %v\n", secretName, nameSpace, err`

wackxu:
@jsafrane Rebased, PTAL

jsafrane:
thanks for the PR!

PR #59, opened 2018-03-14
add UT for getDriverName

wackxu:
fix https://github.com/kubernetes-csi/external-provisioner/issues/42
/assign @lpabon


lpabon:
@wackxu Looks like the CLA still needs to be setup for your account. I will merge this change as soon as it is setup.

PR #62, opened 2018-03-16
Fix mismatch between the codes and README.md

arm64b:
This PR aligns the actual code with the described one in README.md file

Signed-off-by: Dennis Chen <dennis.chen@arm.com>


arm64b:
> Let's fix the README instead. Do you mind sending a fix for the read me to match that of the Makefile? Thanks.

Done!

PR #85, opened 2018-05-13
Return err when SC not found at Delete & add tests

### LINKED ISSUE #39: UT: controller/Delete

darkowlzz:
While deleting, when a StorageClass is not found, maybe due to permission issues, an error should be returned.

Also, while writing the tests, I found that even if I set `StorageClassName` in the `PVC.Spec`, the PV returned by `Provision()` had empty `StorageClassName`. Added that as well.

Fixes #39

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -237,7 +237,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 
 // This function get called before any attepmt to communicate with the driver.
 // Before initiating Create/Delete API calls provisioner checks if Capabilities:
-// PluginControllerService,  ControllerCreateVolume sre supported and gets the  driver name.
+// PluginControllerService,  ControllerCreateVolume are supported and gets the driver name.
 func checkDriverState(grpcClient *grpc.ClientConn, timeout time.Duration) (string, error) {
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -443,20 +447,22 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
...
-			provisionerSecretRef, err := getSecretReference(provisionerSecretNameKey, provisionerSecretNamespaceKey, storageClass.Parameters, volume.Name, nil)
-			if err != nil {
-				return err
-			}
-			credentials, err := getCredentials(p.client, provisionerSecretRef)
-			if err != nil {
-				return err
-			}
-			req.ControllerDeleteSecrets = credentials
+		storageClass, err := p.client.StorageV1beta1().StorageClasses().Get(storageClassName, metav1.GetOptions{})
```
darkowlzz: I'm not sure about this change. The test needed creating `StorageClass` objects and adding them to fake clientSet. But since I couldn't figure out any way to do that with v1 API, I ended up used v1beta1 API. Some help in doing this properly without any side effects would be great.

msau42:
Do we still want to continue this PR?

PR #88, opened 2018-05-15
Remove line logging a secret

### LINKED ISSUE #82: Do not log secret in log
https://github.com/kubernetes-csi/external-provisioner/blob/v0.2.1/pkg/controller/controller.go#L89

This line will log 'controller_create_secrets' if present

shubheksha:
Fixes #82

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -101,7 +101,6 @@ var (
 //TODO consolidate ane librarize
 func logGRPC(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 	glog.V(5).Infof("GRPC call: %s", method)
-	glog.V(5).Infof("GRPC request: %+v", req)
 	err := invoker(ctx, method, req, reply, cc, opts...)
 	glog.V(5).Infof("GRPC response: %+v", reply)
 	glog.V(5).Infof("GRPC error: %v", err)


shubheksha:
@saad-ali, I've removed the line itself as of now. I tried getting only the secret to not be logged, but there was no info available for the `req` param. Please let me know if this is not the correct approach. 😄

pkg/controller/controller.go
```diff
@@ -101,7 +101,6 @@ var (
 //TODO consolidate ane librarize
 func logGRPC(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 	glog.V(5).Infof("GRPC call: %s", method)
-	glog.V(5).Infof("GRPC request: %+v", req)
```
pohly: I was wondering about the exact same issue today, independently of you.

Removing the entire line is too drastic. It also removes less sensitive information.

My conclusion was that logging `req` entirely is probably okay and may be desired when someone explicitly chooses a high log level. However, such a high log level should not be used in production, which implies adding warnings about it somewhere.

Alternatively, formatting of req could be changed so that it masks the actual secret values.

Note that a similar logGRPC implementation also exists elsewhere (other sidecar apps, the hostpath driver itself). Changing just this one line here would not be enough. I am working on refactoring the code so that logGRPC lives in csi-common.

shubheksha: That seems like the way to go to me too. The TODO suggests the same thing. Let me know if I can help in any way with the refactor 😄

pohly: Let's discuss the code refactoring in https://github.com/kubernetes-csi/drivers/issues/81

lpabon:
@shubheksha @pohly I'd like to remove this line asap and add logging work later. What do you think?

shubheksha:
@lpabon, I think that'd serve as a good temporary workaround - better than logging secrets imo.

msau42:
This is no longer needed and is addressed with csi-lib-utils/protosanitizer

Issue #68, opened 2018-04-03
Scaling issue: kube apiserver throttles external-provisioner when 100 PVCs created at same time


saad-ali:
Reported by @cduchesne 

The CSI external-provisioner has scaling issues.

When 100 PVCs are created at same time, the CSI external-provisioner hammers the kube apiserver with requests and gets throttled, causing all sorts of issues.

orainxiong:
@sbezverk, I have hit the same issue. After getting throttled, it will take over 30mins to create a Volume. I will try your PR.

orainxiong:
@sbezverk 

The problem is still not resolved. I am not sure if I understand correctly, the root cause is that `external-provisioner` excessively talk against `Kube-API-Server`.

Every time `external-provisioner` perform the operations of `provisioning` and `deleting` Volume, a sudden surge of requests comes in `Kube-API-Server`, and then the `external-provisioner` get throttled. As a result, all operations will takes unpredictable times and cause errors. 

We can go through the logic of `provisioning` to figure out more details.

* Watch both `Events` and `PersistentVolumeClaim` to detect whether the operation of provisioning is successful or failure:

```
	successCh, err := ctrl.watchProvisioning(claim, stopCh)

```

```
	pvcWatch, err := ctrl.claimSource.Watch(options)
	if err != nil {
		return nil, err
	}

	failWatch, err := ctrl.getPVCEventWatch(claim, v1.EventTypeWarning, "ProvisioningFailed")
	if err != nil {
		pvcWatch.Stop()
		return nil, err
	}

	successWatch, err := ctrl.getPVCEventWatch(claim, v1.EventTypeNormal, "ProvisioningSucceeded")
	if err != nil {
		failWatch.Stop()
		pvcWatch.Stop()
		return nil, err
	}
```

* List `PersistentVolumeClaim` to get leadership information to determine whether
```
	oldLeaderElectionRecord, err := le.config.Lock.Get()
```

* Update `PersistentVolumeClaim`'s annotation to acquire an exclusive lock before perform the operation of provisioning 

```
	if err = le.config.Lock.Update(leaderElectionRecord); err != nil {
		glog.Errorf("Failed to update lock %s: %v", le.config.Lock.Describe(), err)
		return false
	}
```

Once external-provisioner gets throttled, the operation of watching `Events` will take over 10 seconds.
```
[...]

PR #104, opened 2018-06-19
fix apiserver throttling issue

orainxiong:
* modify leadership election
* the abuse of API Server

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -31,10 +31,19 @@ import (
 
 	"google.golang.org/grpc"
 
-	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/api/core/v1"
+
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"k8s.io/client-go/tools/record"
+	utilversion "k8s.io/kubernetes/pkg/util/version"
 )
 
 var (
@@ -47,8 +56,18 @@ var (
 	volumeNameUUIDLength = flag.Int("volume-name-uuid-length", 16, "Length in characters for the generated uuid of a created volume")
 	showVersion          = flag.Bool("version", false, "Show version.")
 
+	version   = "unknown"
+	Namespace = "default"
+
+	informerFactory     informers.SharedInformerFactory
+	resourceLock        resourcelock.Interface
 	provisionController *controller.ProvisionController
-	version             = "unknown"
+)
+
+const (
+	resyncPeriod         = 1 * time.Millisecond
+	sharedResyncPeriod   = 1 * time.Second
+	defaultServerVersion = "v1.6.0"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


lpabon:
@orainxiong So I'm guessing that by using the informers, some of the metadata is cached locally allowing for faster responses?

orainxiong:
@lpabon Thanks for the reply.

more details: #68

wongma7:
To be clear if you do not provide sharedinformers, the controller will still initialize its own informers which have their own caches & to my knowledge there should be no performance difference between sharedinformers and informers (they both need to have an internal cache) since there is nothing to be gained by "sharing" because there are no other controllers "sharing" the informers. It's still not a bad idea to use sharedinformers since we get those factories initializing everything for free and they are what kube uses everywhere as well (better readability). 

All this to say I think the leader election change is the more important/substantial, with the n provisioners * m pvcs problem.

But I am not the one with any data on scalability/stress tests so if sharedinformers alone resulted in  improvement then I would like to know!

The other problem that could cause apiserver problems was too frequent resync (i.e. list all PVC every 15s instead of using the cache) which should be fixed by https://github.com/kubernetes-incubator/external-storage/pull/825

orainxiong:
@wongma7 thanks for your reply.

From my point of view, `external-provisioner` is supposed to leverage informer cache rather than talking directly against `API Server`. I think that is one of the main reason that causes `external-provisioner` getting throttling.

As you mentioned, `sharedinformer` is not the necessity. In my scenario, it significantly improves `external-provisioner`'s efficient and avoid `API Server` throttling.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -110,17 +129,104 @@ func init() {
...
+		RetryPeriod:   2 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: run,
+			OnStoppedLeading: func() {
+				glog.Fatalf("leader election lost")
+			},
+		},
+	})
+
+	panic("unreached")
```
vladimirvivien: Is this for testing ? If so, make sure to remove when done.

pkg/controller/controller.go
```diff
@@ -571,14 +585,21 @@ func resolveTemplate(template string, params map[string]string) (string, error)
...
 	if ref == nil {
 		return nil, nil
 	}
 
-	secret, err := k8s.CoreV1().Secrets(ref.Namespace).Get(ref.Name, metav1.GetOptions{})
+	secretObject, exists, err := secrectsStore.GetByKey(fmt.Sprintf("%s/%s", ref.Namespace, ref.Name))
 	if err != nil {
 		return nil, fmt.Errorf("error getting secret %s in namespace %s: %v", ref.Name, ref.Namespace, err)
+	} else if !exists {
+		return nil, fmt.Errorf("secret %s in namespace %s is not exists", ref.Name, ref.Namespace)
```
vladimirvivien: Should read `secret %s in namespace %s does not exist`

cmd/csi-provisioner/csi-provisioner.go
vladimirvivien: If this panic should never be reached, update the string to say so i.e. `panic("panic reached, leader election failed")`.

pkg/controller/controller.go
```diff
@@ -443,14 +451,20 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
 	// get secrets if StorageClass specifies it
 	storageClassName := volume.Spec.StorageClassName
 	if len(storageClassName) != 0 {
-		if storageClass, err := p.client.StorageV1().StorageClasses().Get(storageClassName, metav1.GetOptions{}); err == nil {
+		storageClassObject, exists, err := p.classes.GetByKey(storageClassName)
+		glog.Infof("%+v %+v", exists, err)
+		if err == nil && exists {
```
vladimirvivien: What happens when `err != nil`?  Add code to handle situations when error occurs.

cmd/csi-provisioner/csi-provisioner.go
orainxiong: Yes. I delete test code.

pkg/controller/controller.go
orainxiong: Ok.

cmd/csi-provisioner/csi-provisioner.go
orainxiong: I have deleted the test code.

pkg/controller/controller.go
orainxiong: Ok.

Issue #108, opened 2018-06-29
Use driver name as provisioner name


msau42:
That way, you don't need to pass in a separate "provisioner" argument, and the StorageClass provisioner will also match the driver name.

jsafrane:
`StorageClass.Provisioner` must pass `IsQualifiedName`, which `"must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character"`, optionally with `<DNS1123Subdomain>/` prefix. It MUST be 63 characters or less.

https://github.com/kubernetes/kubernetes/blob/09784ccf83dec10b0ca6757be4ec143ec44bc993/pkg/apis/storage/validation/validation.go#L81


CSI driver name `"must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character"`. It MUST be 63 characters or less.

https://github.com/container-storage-interface/spec/blob/master/csi.proto#L89-L96

So each CSI driver is valid Provisioner name. All we need is a code that makes `--provisioner` optional.

msau42:
/reopen

@jsafrane it appears that your PR submitted is empty

PR #109, opened 2018-07-03
Make the provisioner name optional.

jsafrane:
Driver name is used as provisioner when no provisioner is specified by the
user.

Fixes: #108


davidz627:
@jsafrane This is a 0 line change. Did the real commit get lost somewhere?

jsafrane:
Hmm, indeed. My git fork is has empty commit too, where did the code go?

I'll re-implement it shortly.

PR #111, opened 2018-07-10
Add support for RWX and ROX access mode in volume provision

wnxn:
Currently, the external provisioner plugin transfer all kinds of Kubernetes access modes, such as RWO, RWX or ROX, to single node writer CSI access mode. 
Therefore, we can create a PVC with RWX Kubernetes access mode through a CSI driver which only supports single node writer CSI access mode. This might cause a strange situtation.


thelinuxfoundation:
Thank you for your pull request. Before we can look at your contribution, we need to ensure all contributors are covered by a Contributor License Agreement.

After the following items are addressed, please respond with a new comment here, and the automated system will re-verify.

* User @wnxn isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

PR #113, opened 2018-07-10
Add Block volume support for CSI provisioner 

### LINKED ISSUE #110: Add Block volume support for CSI provisioner
CSI provisioner should provision Block volumes properly.

In CSI provisioner, below three logics will need to be implemented, to add Block volume support to CSI provisioner:
1. Add ```SupportsBlock``` that properly returns whether Storage Provider's plugin supports block (this will be checked by using ```ValidateVolumeCapabilities```),
2. Pass ```BlockVolume``` instead of ```MountVolume``` to ```CreateVolume``` if ```volumeMode``` is set to be ```Block``` on ```Provision```,
3. Set ```volumeMode``` to PV returned by ```Provision```.

In Storage Provider's plugin, below two logics should be implemented:
1. Return ```true``` to ```ValidateVolumeCapabilities``` with ```BlockVolume``` specified if it supports block, and return ```false``` if it doesn't,
2. Properly handle ```CreateVolume``` with ```BlockVolume``` specified, if ```ValidateVolumeCapabilities``` with ```BlockVolume``` returns ```true```.
    (In most case, no special handling would be needed, as far as I see existing provisioner/attacher codes.)

mkimuram:
Below three logics are implemented to add Block volume support to CSI provisioner:

  1. Add ```SupportsBlock``` that properly returns whether Storage
     Provider's plugin supports block (this is checked
     by using ```ValidateVolumeCapabilities```),
  2. Pass ```BlockVolume``` instead of ```MountVolume``` to ```CreateVolume```
     if ```volumeMode``` is set to be Block on ```Provision```,
  3. Set ```volumeMode``` to PV returned by ```Provision```.

Also, below 4 test cases for TestSupportsBlock and 2 test cases for TestProvision
are added.

```TestSupportsBlock```:
 1. ```ValidateVolumeCapabilities``` return ```(true, nil)``` case: return ```true``` expected
 2. ```ValidateVolumeCapabilities``` return ```(false, nil)``` case: return ```false``` expected
 3. ```ValidateVolumeCapabilities``` return ```(true, err)``` case: return ```false``` expected
 4. ```ValidateVolumeCapabilities``` return ```(false, err)``` case: return ```false``` expected

```TestProvision```:
 1. ```volumeMode```=```Filesystem``` PVC case: return Filesystem PV expected
 2. ```volumeMode```=```Block``` PVC case: return Block PV expected

Fixes #110


mkimuram:
Close this PR and continue working on this in https://github.com/kubernetes-csi/external-provisioner/pull/135

Issue #115, opened 2018-08-02
pvName UUID should not be truncated by default


davidz627:
There was a change introduced that truncates the `pvName` UUID to `volumeNameUUIDLength` which is an external-provisioner parameter.

The default behavior of the driver should be to NOT truncate the UUID as that can cause naming collisions depending on the method of UUID generation. 

The flag should remain for optional truncation.

This fix should be cherrypicked to v0.3.0 so that behavior does not double-regress.
v0.2.0 -> v0.3.0 -> v0.4.0 should not all have different naming behavior.

PR #116, opened 2018-08-02
Change default volume name behavior to non-truncating

davidz627:
Fixes: #115

Changes default behavior of volume name to non-truncating. If the flag is set we will remove all dashes and truncate volume name.

/assign @cduchesne @saad-ali 

I believe this should also be cherry-picked to v0.3.0 container as we don't want default behavior of naming to jitter between versions.


cduchesne:
lgtm

PR #123, opened 2018-08-10
WIP: Restore Volume from Snapshot

xing-yang:
This PR adds changes to support restoring a volume from a snapshot.

Note: Mark it WIP until the DataSource addition in core API (kubernetes/kubernetes#67087) and snapshot API/controller changes (kubernetes-csi/external-snapshotter#7) are merged.


pkg/controller/controller.go
```diff
@@ -289,7 +339,21 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 		return nil, fmt.Errorf("claim Selector is not supported")
 	}
 
-	driverName, err := checkDriverState(p.grpcClient, p.timeout)
+	var checkSnapshotSupport bool
+	if options.PVC.Spec.DataSource != nil {
+		// PVC.Spec.DataSource.Name is the name of the VolumeSnapshot API object
+		if options.PVC.Spec.DataSource.Name == "" {
```
jingxu97: Since validation code already check this, do we need these checks here?

pkg/controller/controller.go
```diff
@@ -317,6 +381,46 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 			RequiredBytes: int64(volSizeBytes),
 		},
 	}
+
+	if checkSnapshotSupport {
+		snapshotObj, err := p.snapshotClient.VolumesnapshotV1alpha1().VolumeSnapshots(options.PVC.Namespace).Get(options.PVC.Spec.DataSource.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("error get snapshot %s from api server: %v", options.PVC.Spec.DataSource.Name, err)
```
jingxu97: error getting ..?

pkg/controller/controller.go
```diff
@@ -317,6 +381,46 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
+		}
+		glog.V(5).Infof("VolumeSnapshot %+v", snapshotObj)
+
+		snapContentObj, err := p.snapshotClient.VolumesnapshotV1alpha1().VolumeSnapshotContents().Get(snapshotObj.Spec.SnapshotContentName, metav1.GetOptions{})
```
jingxu97: Maybe we can check VolumeSnapshot status to see whether it is ready?

pkg/controller/controller.go
xing-yang: Actually these can be removed now.  I'll remove them.

pkg/controller/controller.go
xing-yang: Sure.

pkg/controller/controller.go
xing-yang: Good point

pkg/controller/controller.go
xing-yang: Maybe we should keep these checks here?  When we add support for more Data Source types, the validation check will be changed to allow other Data Sources.

pkg/controller/controller.go
```diff
@@ -289,7 +339,21 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 		return nil, fmt.Errorf("claim Selector is not supported")
 	}
 
-	driverName, err := checkDriverState(p.grpcClient, p.timeout)
+	var checkSnapshotSupport bool
```
jingxu97: call it needSnapshotSupport?

pkg/controller/controller.go
```diff
@@ -317,6 +381,46 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 			RequiredBytes: int64(volSizeBytes),
 		},
 	}
+
+	if checkSnapshotSupport {
```
jingxu97: maybe to put this part into a separate function to get snapshot handler?

pkg/controller/controller.go
```diff
@@ -289,7 +339,21 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
+		}
+		if options.PVC.Spec.DataSource.Kind != snapshotKind {
+			return nil, fmt.Errorf("the PVC source is not the right type. Expected %s, Got %s", snapshotKind, options.PVC.Spec.DataSource.Kind)
+		}
+		if options.PVC.Spec.DataSource.APIGroup != snapshotAPIGroup {
+			return nil, fmt.Errorf("the PVC source does not belong to the right APIGroup. Expected %s, Got %s", snapshotAPIGroup, options.PVC.Spec.DataSource.APIGroup)
+		}
+		checkSnapshotSupport = true
+	}
+	driverName, err := checkDriverState(p.grpcClient, p.timeout, checkSnapshotSupport)
```
jingxu97: Consider we will keep adding more types of data sources, a boolean value added here might not be convenient. But we could chance it later.

pkg/controller/controller.go
```diff
@@ -69,13 +71,17 @@ const (
 	backoffSteps    = 10
 
 	defaultFSType = "ext4"
+
+	snapshotKind     = "VolumeSnapshot"
+	snapshotAPIGroup = "snapshot.storage.k8s.io"
```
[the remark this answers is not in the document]
xing-yang: Looks like I can get GroupName from snapshot api package.  It is a constant there.

As for "VolumeSnapshot", it is a struct defined in types.go but there is no constant.  I think I can get it by instantiating it first and then use "reflect" to get it, but it seems weird to get it that way.  I haven't found an easier way.

pkg/controller/controller.go

pkg/controller/controller.go
xing-yang: Sure

pkg/controller/controller.go

pkg/controller/controller.go
```diff
@@ -257,6 +292,21 @@ func checkDriverState(grpcClient *grpc.ClientConn, timeout time.Duration) (strin
 		glog.Error("no create/delete volume support detected")
 		return "", fmt.Errorf("no create/delete volume support detected")
 	}
+
+	if checkSnapshotSupport {
```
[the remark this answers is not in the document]

pkg/controller/controller.go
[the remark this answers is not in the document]
xing-yang: Changed snapshotAPIGroup to use GroupName from external-snapshotter lib.

pkg/controller/controller.go

pkg/controller/controller.go
```diff
@@ -537,6 +512,51 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 	return pv, nil
 }
 
+func (p *csiProvisioner) getSnapshotHandle(options controller.VolumeOptions) (*csi.VolumeContentSource, error) {
```
jingxu97: maybe call it getVolumeContentSource?

Gopkg.toml
```diff
@@ -40,14 +44,19 @@
   name = "google.golang.org/grpc"
   version = "1.9.2"
 
-[[constraint]]
+[[override]]
```
davidz627: these probably do not have to be overrides, constraint should be better for most of these as they will behave correct if dependencies are removed/changed. We only need to override if we transitively depend on a certain version of a package, but I believe most of these are direct dependencies so we should just use `[[constraint]]`

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -86,6 +87,10 @@ func init() {
 	if err != nil {
 		glog.Fatalf("Failed to create client: %v", err)
 	}
+	snapClient, err := snapclientset.NewForConfig(config)
```
davidz627: What does `NewForConfig` mean? is it `NewSnapClientFromConfig`?

pkg/controller/controller.go
```diff
@@ -289,7 +343,21 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 		return nil, fmt.Errorf("claim Selector is not supported")
 	}
 
-	driverName, err := checkDriverState(p.grpcClient, p.timeout)
+	var needSnapshotSupport bool
```
davidz627: nit: explicitly set `needSnapshotSupport := false` here

pkg/controller/controller.go
```diff
@@ -317,6 +385,17 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
+
+	if needSnapshotSupport {
+		volumeContentSource, err := p.getVolumeContentSource(options)
+		if err != nil {
+			return nil, fmt.Errorf("error getting snapshot handle for snapshot %s: %v", options.PVC.Spec.DataSource.Name, err)
+		}
+		req.VolumeContentSource = volumeContentSource
+	}
+
+	glog.Infof("CreateVolumeRequest %+v", req)
```
davidz627: remove debug logs or decrease verbosity

pkg/controller/controller.go
```diff
@@ -433,13 +512,58 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
+
+	snapContentObj, err := p.snapshotClient.VolumesnapshotV1alpha1().VolumeSnapshotContents().Get(snapshotObj.Spec.SnapshotContentName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("error getting snapshot:snapshotcontent %s:%s from api server: %v", snapshotObj.Name, snapshotObj.Spec.SnapshotContentName, err)
+	}
+	glog.V(5).Infof("VolumeSnapshotContent %+v", snapContentObj)
+
+	snapshotSource := csi.VolumeContentSource_Snapshot{
+		Snapshot: &csi.VolumeContentSource_SnapshotSource{
+			Id: snapContentObj.Spec.VolumeSnapshotSource.CSI.SnapshotHandle,
```
davidz627: nit: nil check on this `SnapshotHandle`?

pkg/controller/controller.go
```diff
@@ -433,13 +512,58 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
+		},
+	}
+	glog.V(5).Infof("VolumeContentSource_Snapshot %+v", snapshotSource)
+
+	if snapshotObj.Status.RestoreSize != nil {
+		capacity := options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
```
davidz627: nit: exists check

pkg/controller/controller_test.go
```diff
@@ -931,3 +1076,343 @@ func TestProvision(t *testing.T) {
...
+		AccessModes   []v1.PersistentVolumeAccessMode
+		Capacity      v1.ResourceList
+		CSIPVS        *v1.CSIPersistentVolumeSource
+	}
+
+	testcases := map[string]struct {
+		volOpts              controller.VolumeOptions
+		restoredVolSizeSmall bool
+		wrongDataSource      bool
+		validSnapshotStatus  bool
```
davidz627: nit: rename to `snapshotStatusReady`

pkg/controller/controller_test.go
```diff
@@ -511,6 +620,42 @@ func provisionMockServerSetupExpectations(identityServer *driver.MockIdentitySer
 	}, nil).Times(1)
 }
 
+func provisionFromSnapshotMockServerSetupExpectations(identityServer *driver.MockIdentityServer, controllerServer *driver.MockControllerServer) {
```
davidz627: Its not incredibly clear what expectations we have here, could you add a comment explaining what expectations these are and what we use them for?

pkg/controller/controller_test.go
```diff
@@ -931,3 +1076,343 @@ func TestProvision(t *testing.T) {
...
+
+		out := &csi.CreateVolumeResponse{
+			Volume: &csi.Volume{
+				CapacityBytes: requestedBytes,
+				Id:            "test-volume-id",
+			},
+		}
+
+		// Setup mock call expectations.
+		if tc.wrongDataSource == false {
```
davidz627: its not clear to me what this means and why we do this

pkg/controller/controller_test.go
```diff
@@ -931,3 +1076,343 @@ func TestProvision(t *testing.T) {
...
+			provisionFromSnapshotMockServerSetupExpectations(identityServer, controllerServer)
+		}
+		if tc.restoredVolSizeSmall == false && tc.wrongDataSource == false && tc.validSnapshotStatus {
```
davidz627: what happens if any of these are true? Specifically if `restoredVolSizeSmall` is true it doesn't look like we do anything special

Gopkg.toml
xing-yang: I added "override" because I ran into errors earlier without them.  I just removed them and re-ran them.  It works now.

cmd/csi-provisioner/csi-provisioner.go
xing-yang: Added a comment for this: "NewForConfig creates a new Clientset for VolumesnapshotV1alpha1Client"

pkg/controller/controller.go

pkg/controller/controller.go
xing-yang: Changed to glog.V(5).Infof

pkg/controller/controller_test.go
xing-yang: Added the following comments: 
                // If tc.wrongDataSource is false, DataSource is valid
                // and the controller will proceed to check whether the plugin supports snapshot.
                // So in this case, we need the plugin to report snapshot support capabilities;
                // Otherwise, the controller will fail the operation so it won't check the capabilities.

pkg/controller/controller_test.go
xing-yang: Added the following comments:
                // If tc.restoredVolSizeSmall is true, or tc.wrongDataSource is true, or
                // tc.snapshotStatusReady is false,  create volume from snapshot operation will fail
                // early and therefore CreateVolume is not expected to be called.
                // When the if condition is met, it is a valid create volume from snapshot
                // operation and CreateVolume is expected to be called.

pkg/controller/controller_test.go
xing-yang: Added comments

pkg/controller/controller_test.go
xing-yang: Done

pkg/controller/controller.go

pkg/controller/controller.go
```diff
@@ -529,6 +529,10 @@ func (p *csiProvisioner) getVolumeContentSource(options controller.VolumeOptions
 	}
 	glog.V(5).Infof("VolumeSnapshotContent %+v", snapContentObj)
 
+	if snapContentObj.Spec.VolumeSnapshotSource.CSI == nil || len(snapContentObj.Spec.VolumeSnapshotSource.CSI.SnapshotHandle) == 0 {
```
jingxu97: Just to confirm we don't need to check Spec.VolumeSnapshotSource because it is required field?

xing-yang: So I removed the check for SnapshotHandle but kept the check for CSI is nil.  Let me know if this is okay.

xing-yang:
@jingxu97 @davidz627 @lpabon Review comments are addressed.  Can you please review again?  Thanks!

PR #126, opened 2018-08-13
Support mount options

### LINKED ISSUE #125: Mount options are not passed from the storage class to the CSI plugin
Storage drivers that use NFS typically need to provide more flexibility for mounting than block-based storage drivers. Before CSI, the mountOptions field in the storage class spec was used to supply these.

Currently, however, the kubernetes CSI sidecar containers does not pass through any mount options from the storage class to the CSI plugin.

A few things need to change to allow this use case.
1) The CSI driver within the kubernetes project needs to report that mount options are supported (currently support is hardcoded to false)
2) The external provisioner needs to copy the mount options from the storage class to the PV spec at creation time.
3) All of the the RPC calls to the CSI plugin where mount options are an option should include the mount options where appropriate.

bswartz:
Copy the mount options from the storage class to the PV spec and pass mount option to the CSI driver when creating volumes.

Fixes #125

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -107,9 +107,6 @@ const (
 var _ controller.Provisioner = &csiProvisioner{}
 
 var (
-	accessType = &csi.VolumeCapability_Mount{
-		Mount: &csi.VolumeCapability_MountVolume{},
-	}
 	// Each provisioner have a identify string to distinguish with others. This
 	// identify string will be added in PV annoations under this key.
 	provisionerIDKey = "storage.kubernetes.io/csiProvisionerIdentity"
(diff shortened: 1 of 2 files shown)


bswartz:
The first commit in this PR is just a vendor update to get a newer version of kubernetes-incubator/external-storage. The second commit is my code.

Gopkg.toml
```diff
@@ -24,16 +24,12 @@
...
-[[constraint]]
-  name = "github.com/golang/protobuf"
-  version = "v1.1.0"
-
 [[constraint]]
   branch = "master"
   name = "github.com/golang/glog"
 
 [[constraint]]
-  branch = "master"
```
saad-ali: dep ensure -upgrade ?

saad-ali:
High level concept makes sense to me.

wongma7:
enabling dep's pruning may reduce filechange churn https://golang.github.io/dep/docs/Gopkg.toml.html#prune

wongma7:
then it's unavoidable...welcome to golang dependency management :D

bswartz:
Rebased and added unit tests

PR #133, opened 2018-09-13
Support all Kubernetes access modes

lpabon:
Currently, the code is hardcoded to always send the AccessMode of SINGLE_WRITER to the CSI plugin even though the request from the PVC is to have a ReadWriteMany.

Closes #132 
Signed-off-by: Luis Pabón <luis@portworx.com>

PR #134, opened 2018-09-13
Backport: Support all Kubernetes access modes

lpabon:
Backport of #133 to release 0.3.0
Signed-off-by: Luis Pabón <luis@portworx.com>

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -86,9 +86,6 @@ type csiProvisioner struct {
 var _ controller.Provisioner = &csiProvisioner{}
 
 var (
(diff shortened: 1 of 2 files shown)


saad-ali:
It is unfortunate that the branch was called 0.3.0 it should have been 0.3. And we can cut 0.3.0, 0.3.1, etc. off that branch. Regardless, after this is merged, please create a new 0.3.1 tag instead of overwriting existing released 0.3.0 version tag.

saad-ali:
```
 command "go fmt $(go list ./... | grep -v vendor) | wc -l | grep 0" exited with 1.
```

PR #138, opened 2018-09-17
Add the check for deleting snapshot while creating volume

jingxu97:
If the snapshot is currently being deleted, the deletionTimestamp is
set, and provisioner should not try to create the volume in this case.


jingxu97:
cc @xing-yang

xing-yang:
lgtm

jingxu97:
cc @jsafrane

jingxu97:
cc @davidz627

PR #142, opened 2018-10-01
Make the provisioner name optional

jsafrane:
Driver name is used as provisioner when no provisioner is specified by the
user.

Fixes: #108

@davidz627

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -40,7 +40,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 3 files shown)


davidz627:
@jsafrane @lpabon did this ever make it into the v0.4.x release? I see it on the release notes but if I check the 0.4 release branch the PR and code are not there.
Also I'm using 0.4.1 and not seeing the log messages that should be there when provisioner is not specified.

/cc @msau42 
xref: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/126#discussion_r226146749

msau42:
Cherrypick: https://github.com/kubernetes-csi/external-provisioner/pull/161

Issue #147, opened 2018-10-05
Provisioner should not set node affinity if k8s cluster has CSINodeInfo feature disabled.


verult:
If k8s is running on version 1.12 with CSINodeInfo disabled and external provisioner 0.4.0, volume mount fails with mismatch nodeaffinity, because when `CSINodeInfo` feature is disabled, Node objects are not populated with topology labels, while provisioner creates a NodeAffinity for the volume regardless.

Provisioner should have a feature flag to enable/disable topology

This is a blocker for 0.4.0 external-provisioner release

/priority critical-urgent
/kind bug

verult:
Fixed in https://github.com/kubernetes-csi/external-provisioner/pull/148

PR #148, opened 2018-10-08
Flag gate for topology and leader election

### LINKED ISSUE #146: Add a flag to enable/disable leader election
Piggyback on the mechanism introduced here: https://github.com/kubernetes-incubator/external-storage/pull/957

/assign

/kind bug

verult:
/kind bug
/priority critical-urgent

Fixes #146 and #147 

I'm not a fan of how the topology flag is set up, but I'm hitting some hurdles integrating the feature gate mechanism from k8s/k8s here (namely how to properly setup spf13/pflags). Going to try to resolve those issues


verult:
Depends on https://github.com/kubernetes-csi/external-provisioner/pull/145 to resolve external-snapshotter dependency issue in Travis

davidz627:
Not sure it's necessary to add this whole new flag gating mechanism. Could we just have it as a normal command line flag or is there a compelling reason not to do that?

Currently v0.4.0 of the provisioner used with a topology supporting driver is not compatible with any Kubernetes cluster except for 1.12 with Alpha Topology support enabled so I would consider this high priority.

/cc @jingxu97 
for tracking as this is blocking GCE PD CSI Driver Snapshots Support

verult:
The simplest approach to adding a command-line flag would involve plumbing the flag parameter to every code path that checks the flag, which doesn't scale very well. We'd need a global source of truth for feature gates, and the library in k8s sets that up nicely so we don't have to write our own.

verult:
Status update: PR #145 depends on external-snapshotter being tagged, which will be done today

wongma7:
enableLeaderElection should default false to match external-attacher imo.

pohly:
> enableLeaderElection should default false to match external-attacher imo.

That, and it's also the right default for a `StatefulSet` (which is what many people and the example use) because then leader election isn't needed.

verult:
@pohly what kind of configuration would be used if leader election were enabled?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -37,17 +38,23 @@ import (
...
 	kubeconfig           = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Either this or master needs to be set if the provisioner is being run out of cluster.")
-	csiEndpoint          = flag.String("csi-address", "/run/csi/socket", "The gRPC endpoint for Target CSI Volume")
+	csiEndpoint          = flag.String("csi-address", "/run/csi/socket", "The gRPC endpoint for Target CSI Volume.")
 	connectionTimeout    = flag.Duration("connection-timeout", 10*time.Second, "Timeout for waiting for CSI driver socket.")
-	volumeNamePrefix     = flag.String("volume-name-prefix", "pvc", "Prefix to apply to the name of a created volume")
+	volumeNamePrefix     = flag.String("volume-name-prefix", "pvc", "Prefix to apply to the name of a created volume.")
 	volumeNameUUIDLength = flag.Int("volume-name-uuid-length", -1, "Truncates generated UUID of a created volume to this length. Defaults behavior is to NOT truncate.")
 	showVersion          = flag.Bool("version", false, "Show version.")
+	enableLeaderElection = flag.Bool("enable-leader-election", false, "Enables leader election.")
```
saad-ali: Add to comment: enabling may require additional RBAC

msau42:
Is this change ready to go?

davidz627:
@verult can a new provisioner version be cut for this?

verult:
v0.4.1 has been cut!

Issue #143, opened 2018-10-02
Topology spreading across provisioned StatefulSet volumes


verult:
/kind feature

The intention is briefly described here: 
https://github.com/kubernetes/community/pull/2034/files#diff-8fe0dac6cc0083d59d7b6dfa91462244R170

The StatefulSet spreading logic, referred to in the design doc, is here:
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/util.go#L649-L665

@ddebroy

ddebroy:
I will pick this one up. Cannot assign myself because I believe I am not a `member`.

PR #151, opened 2018-10-16
Evenly spread volumes of a StatefuleSet across nodes based on topology

ddebroy:
If immediate volume binding mode is set and the PVC follows StatefulSet naming format, then the provisioner will choose, as the first segment in preferred topology, a segment from requisite topology based on the relevant parts of the PVC name that ensures an even spread of topology across the StatefulSet's volumes.

The logic around determining the StatefulSet naming format and obtaining a hash and index is identical to the logic in pkg/volume/util/util.go Kubernetes main tree.

Fixes: https://github.com/kubernetes-csi/external-provisioner/issues/143

cc @msau42 @verult

### DIFF
--- a/pkg/controller/topology.go
+++ b/pkg/controller/topology.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"github.com/container-storage-interface/spec/lib/go/csi/v0"
 	"github.com/golang/glog"
+	"hash/fnv"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -27,6 +28,7 @@ import (
 	csiclientset "k8s.io/csi-api/pkg/client/clientset/versioned"
 	"math/rand"
 	"sort"
+	"strconv"
 	"strings"
 )
 
@@ -68,6 +70,7 @@ func GenerateAccessibilityRequirements(
 	kubeClient kubernetes.Interface,
 	csiAPIClient csiclientset.Interface,
 	driverName string,
+	pvcName string,
 	allowedTopologies []v1.TopologySelectorTerm,
 	selectedNode *v1.Node) (*csi.TopologyRequirement, error) {
 	requirement := &csi.TopologyRequirement{}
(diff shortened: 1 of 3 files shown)


pkg/controller/topology_test.go
```diff
@@ -146,6 +146,210 @@ func TestGenerateVolumeNodeAffinity(t *testing.T) {
...
+	}
+
+	testcases := map[string]struct {
+		pvcName           string
+		expectedPreferred []*csi.Topology
+	}{
+		// ordering of segments in preferred array is sensitive to statefulset name: testset
+		// if statefuleset name is changed, make sure expectedPreferred is kept in sync
+		// pvc prefix in pvcName does not have any effect on segment ordering
+		"pvc 1a": {
```
msau42: Also a test case when allowedTopologies is specified?.  Do we also need to to convert the old zone/zones to AllowedTopologies for migration?

ddebroy: Good point - I can add an additional sanity test with `allowedTopologies` specified for completeness. `TestAllowedTopologies` already covers several scenarios for trimming down the `requisiteTerms` based on `allowedTopologies` but it will be good to have a sanity case here as well.

Added point `4b` in https://docs.google.com/document/d/15Q4wXVqbalrVgf_jlGis4QFXN1RY1GMiY-Ek-hpCr-o/edit# for the migration work around topology.

pkg/controller/topology_test.go
```diff
@@ -146,6 +146,210 @@ func TestGenerateVolumeNodeAffinity(t *testing.T) {
...
+	keys := map[string][]string{testDriverName: {"com.example.csi/zone", "com.example.csi/rack"}}
+	for i := 0; i < len(nodeLabels); i++ {
+		topologyKeys = append(topologyKeys, keys)
+	}
+
+	testcases := map[string]struct {
+		pvcName           string
+		expectedPreferred []*csi.Topology
+	}{
+		// ordering of segments in preferred array is sensitive to statefulset name: testset
```
verult: nit: Maybe move this comment outside? At first glance I thought this comment was for a specific test case

pkg/controller/topology_test.go
```diff
@@ -146,6 +146,210 @@ func TestGenerateVolumeNodeAffinity(t *testing.T) {
...
+		// if statefuleset name is changed, make sure expectedPreferred is kept in sync
```
verult: statefuleset -> statefulset

pkg/controller/topology_test.go
verult: nit: Use a test name that describes what the case is testing

ddebroy: Extra test cases with `allowedTopologies` added in latest commit.

ddebroy:
Thanks for the detailed review @verult and @msau42. Addressed the code review comments so far.

pkg/controller/topology.go
```diff
@@ -120,6 +123,14 @@ func GenerateAccessibilityRequirements(
 			return nil, fmt.Errorf("topology %v from selected node %q is not in requisite", selectedTopology, selectedNode.Name)
 		}
 
+		requirement.Preferred = toCSITopology(preferredTerms)
+	} else {
+		hash, index := getPVCNameHashAndIndexOffset(pvcName)
+		sort.Slice(requisiteTerms, func(i, j int) bool {
```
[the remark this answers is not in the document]
ddebroy: Done. Added an an extra parameter to `sortAndShift()` - `shiftIndex` - which will be considered if the `topologyTerm` parameter is `nil`.

ddebroy:
@verult comments fixed.

ddebroy:
@verult assigned you so that lgtm is accepted :-)

ddebroy:
@jsafrane can you PTAL?

jsafrane:
/hold

please squash the last commit

ddebroy:
@jsafrane last commit squashed.

PR #155, opened 2018-10-19
Use sig-storage-lib-external-provisioner instead of external-storage

wongma7:
Use v1.0.0 and bump to v2.0.0(-alpha) when we move to 1.13 dependencies. Please discuss here https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/issues/2


wongma7:
done: https://github.com/kubernetes-csi/external-provisioner/pull/168

PR #156, opened 2018-10-21
deploy: split out RBAC, fix leadership election permissions

pohly:
The RBAC definitions were not updated when introducing leadership
election, which depends on the permission to modify endpoints in the
current namespace.

Having the RBAC definition in a separate file makes them usable as-is
elsewhere without manual editing, for example in kubernetes/e2e.

### DIFF
--- a/deploy/kubernetes/rbac.yaml
+++ b/deploy/kubernetes/rbac.yaml
@@ -1,0 +1,89 @@
+# This YAML file contains all RBAC objects that are necessary to run external
+# CSI provisioner.
+#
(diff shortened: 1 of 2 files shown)


pohly:
This rbac.yaml file is identical to the one in the E2E PR https://github.com/kubernetes/kubernetes/pull/69868

pohly:
The testing was done with the 0.4 release. If I haven't missed something, the provisioner still uses endpoints. The change to use configmaps as in the external-attacher will have to update the rbac.yaml together with the code.

PR #159, opened 2018-10-24
Share topology keys

liggitt:
restructures how topology keys are provided to GenerateAccessibilityRequirements so they only get looked up once when a selected node is provided, and so we can unit test how GenerateAccessibilityRequirements behaves with different sets of topology keys

if we end up with topologyKeys living in the CSIDriver object, this top-level lookup will get even simpler, and we'll have one spot to swap it out

### DIFF
--- a/pkg/controller/topology.go
+++ b/pkg/controller/topology.go
@@ -65,8 +65,8 @@ func GenerateVolumeNodeAffinity(accessibleTopology []*csi.Topology) *v1.VolumeNo
 }
 
 func GenerateAccessibilityRequirements(
+	topologyKeys []string,
 	kubeClient kubernetes.Interface,
-	csiAPIClient csiclientset.Interface,
 	driverName string,
 	allowedTopologies []v1.TopologySelectorTerm,
 	selectedNode *v1.Node) (*csi.TopologyRequirement, error) {
@@ -77,7 +77,7 @@ func GenerateAccessibilityRequirements(
 	if len(allowedTopologies) == 0 {
 		// Aggregate existing topologies in nodes across the entire cluster.
 		var err error
-		requisiteTerms, err = aggregateTopologies(kubeClient, csiAPIClient, driverName, selectedNode)
+		requisiteTerms, err = aggregateTopologies(topologyKeys, kubeClient)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


pkg/controller/topology.go
```diff
@@ -219,7 +216,7 @@ func aggregateTopologies(
 //
 // This flattening is then applied to all TopologySelectorTerms in AllowedTopologies, and
 // the resulting terms are OR'd together.
-func flatten(allowedTopologies []v1.TopologySelectorTerm) []topologyTerm {
+func flatten(topologyKeys []string, allowedTopologies []v1.TopologySelectorTerm) []topologyTerm {
```
ddebroy: The `topologyKeys` parameter does not seem to be used inside `flatten()`. Can it be removed?

liggitt: sorry, came out of prep for https://github.com/kubernetes-csi/external-provisioner/pull/160

I can remove it and move the param into that PR if you prefer

pkg/controller/topology.go
```diff
@@ -65,6 +65,7 @@ func GenerateVolumeNodeAffinity(accessibleTopology []*csi.Topology) *v1.VolumeNo
 }
 
 func GenerateAccessibilityRequirements(
+	topologyKeys []string,
 	kubeClient kubernetes.Interface,
 	csiAPIClient csiclientset.Interface,
```
ddebroy: `csiAPIClient` does not appear to be used any more in `GenerateAccessibilityRequirements`. Can it be removed?

pkg/controller/topology.go
ddebroy: Ah I see. Moving it to #160 sounds great.

liggitt:
comments addressed

ddebroy:
Should this PR be merged given the changes here, as mentioned earlier, is also covered in https://github.com/kubernetes-csi/external-provisioner/pull/160/files?

liggitt:
> Should this PR be merged given the changes here, as mentioned earlier, is also covered in https://github.com/kubernetes-csi/external-provisioner/pull/160/files?

Yes, these changes are useful now. I will rebase the other, and hold it pending the discussion in https://github.com/kubernetes/community/pull/911

verult:
Related comment here: https://github.com/kubernetes-csi/external-provisioner/pull/160#issuecomment-433576451

PR #160, opened 2018-10-24
WIP - DO NOT MERGE: CSI label mapping

liggitt:
Proof of concept mapping CSI topology keys to kubernetes labels under a specified topology namespace.

Builds on https://github.com/kubernetes-csi/external-provisioner/pull/159

Related to https://github.com/kubernetes/community/pull/911

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -416,9 +416,10 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 
 	if driverState.capabilities.Has(PluginCapability_ACCESSIBILITY_CONSTRAINTS) &&
 		utilfeature.DefaultFeatureGate.Enabled(features.Topology) {
+		topologyKeys := lookupTopologyKeys(p.client, p.csiAPIClient, driverState.driverName, options.SelectedNode)
 		requirements, err := GenerateAccessibilityRequirements(
+			topologyKeys,
 			p.client,
-			p.csiAPIClient,
 			driverState.driverName,
 			options.AllowedTopologies,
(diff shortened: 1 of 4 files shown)


verult:
Should we include the namespace in `CSINodeInfo`'s `TopologyKeys` as well? That way it becomes consistent, that every k8s representation of topology keys includes the namespace, and the external-provisioner and kubelet driver registration mechanism serve as the boundaries and perform the translation.

liggitt:
> Should we include the namespace in `CSINodeInfo`'s `TopologyKeys` as well? That way it becomes consistent, that every k8s representation of topology keys includes the namespace, and the external-provisioner and kubelet driver registration mechanism serve as the boundaries and perform the translation.

No, to round trip accurately, we need the originally returned topology keys from the driver. Otherwise we can't tell the difference between a kube-aware CSI driver that uses topology keys like "acme.com.topology.kubernetes.io/zone" and one that used a topology key of "acme.com/zone"

> If this is done, then `lookupTopologyKeys()` doesn't need to be outside `GenerateTopologyRequirements()`, which is helpful to make sure `lookupTopologyKeys()` is fully unit tested

The unit tests are still exercising lookupTopologyKeys(), right?

verult:
> Otherwise we can't tell the difference between a kube-aware CSI driver that uses topology keys like "acme.com.topology.kubernetes.io/zone" and one that used a topology key of "acme.com/zone"

To clarify, are you proposing that if the driver uses `acme.com.topology.kubernetes.io/zone`, the node labels would also be `acme.com.topology.kubernetes.io/zone`? Do we need to support this use case?

PR #166, opened 2018-11-15
Update external-provisioner to use csi spec 1.0

msau42:


saad-ali:
Modify `ControllerPublishVolume` to check for new `PUBLISH_READONLY` controller capability before setting the `ControllerPublishVolume.readonly` field.

See https://github.com/container-storage-interface/spec/pull/310/files for details

PR #168, opened 2018-11-20
Bumping k8s version to 1.13.0-beta.1; Pointing to kubernetes-sigs/sig-storage-lib-external-provisioner instead of external-storage; Update CSINodeInfo schema

verult:
Note for reviewers: version bump and provisioner lib change are combined into the first commit. See Gopkg.toml and all changes in `cmd` and `pkg`.


Gopkg.toml
```diff
@@ -38,32 +38,33 @@
 
 [[constraint]]
   name = "github.com/kubernetes-csi/external-snapshotter"
-  version = "1.0.0-rc2"
+  #version = "1.0.0-rc2"
+  branch = "master"
```
msau42: why does this need to be master?

verult: Ah ok didn't realize rc3 has been cut, will update

PR #178, opened 2018-11-29
Add new  reserved prefixed parameter keys which are stripped out of parameter list, add deprecation notice for old keys and keep their behavior the same

davidz627:
Add new reserved prefixed parameter keys which are stripped out of parameter list:
```
csiParameterPrefix = "csi.storage.k8s.io/"
prefixedFsTypeKey = csiParameterPrefix + "FSType"
prefixedProvisionerSecretNameKey      = csiParameterPrefix + "ProvisionerSecretName"
prefixedProvisionerSecretNamespaceKey = csiParameterPrefix + "ProvisionerSecretNamespace"
prefixedControllerPublishSecretNameKey      = csiParameterPrefix + "ControllerPublishSecretName"
prefixedControllerPublishSecretNamespaceKey = csiParameterPrefix + "ControllerPublishSecretNamespace"
prefixedNodeStageSecretNameKey      = csiParameterPrefix + "NodeStageSecretName"
prefixedNodeStageSecretNamespaceKey = csiParameterPrefix + "NodeStageSecretNamespace"
prefixedNodePublishSecretNameKey      = csiParameterPrefix + "NodePublishSecretName"
prefixedNodePublishSecretNamespaceKey = csiParameterPrefix + "NodePublishSecretNamespace"
```

add deprecation notice for old keys, example:
`"fstype" is deprecated and will be removed in a future release, please use "csi.storage.k8s.io/FSType" instead`

the behavior will be the same for the old parameter keys, except for the fact that they cannot be specified in conjunction with their respective new prefixed parameter key

/assign @saad-ali @lpabon @jsafrane @msau42 @mkimuram @vladimirvivien

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -56,7 +56,37 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 )
 
+type deprecatedSecretParamsMap struct {
+	name                         string
+	deprecatedSecretNameKey      string
+	deprecatedSecretNamespaceKey string
+	secretNameKey                string
+	secretNamespaceKey           string
+}
+
 const (
+	// CSI Parameters prefixed with csiParameterPrefix are not passed through
+	// to the driver on CreateVolumeRequest calls. Instead they are intended
+	// to used by the CSI external-provisioner and maybe used to populate
+	// fields in subsequent CSI calls or Kubernetes API objects.
+	csiParameterPrefix = "csi.storage.k8s.io/"
+
+	prefixedFsTypeKey = csiParameterPrefix + "fstype"
+
+	prefixedProvisionerSecretNameKey      = csiParameterPrefix + "provisioner-secret-name"
+	prefixedProvisionerSecretNamespaceKey = csiParameterPrefix + "provisioner-secret-namespace"
+
+	prefixedControllerPublishSecretNameKey      = csiParameterPrefix + "controller-publish-secret-name"
+	prefixedControllerPublishSecretNamespaceKey = csiParameterPrefix + "controller-publish-secret-namespace"
+
+	prefixedNodeStageSecretNameKey      = csiParameterPrefix + "node-stage-secret-name"
+	prefixedNodeStageSecretNamespaceKey = csiParameterPrefix + "node-stage-secret-namespace"
+
+	prefixedNodePublishSecretNameKey      = csiParameterPrefix + "node-publish-secret-name"
+	prefixedNodePublishSecretNamespaceKey = csiParameterPrefix + "node-publish-secret-namespace"
+
+	// [Deprecated] CSI Parameters that are put into fields but
+	// NOT stripped from the parameters passed to CreateVolume
 	provisionerSecretNameKey      = "csiProvisionerSecretName"
 	provisionerSecretNamespaceKey = "csiProvisionerSecretNamespace"
 
@@ -80,6 +110,40 @@ const (
 
 	snapshotKind     = "VolumeSnapshot"
 	snapshotAPIGroup = snapapi.GroupName // "snapshot.storage.k8s.io"
+)
+
+var (
+	provisionerSecretParams = deprecatedSecretParamsMap{
+		name:                         "Provisioner",
+		deprecatedSecretNameKey:      provisionerSecretNameKey,
+		deprecatedSecretNamespaceKey: provisionerSecretNamespaceKey,
+		secretNameKey:                prefixedProvisionerSecretNameKey,
+		secretNamespaceKey:           prefixedProvisionerSecretNamespaceKey,
+	}
+
+	nodePublishSecretParams = deprecatedSecretParamsMap{
+		name:                         "NodePublish",
+		deprecatedSecretNameKey:      nodePublishSecretNameKey,
+		deprecatedSecretNamespaceKey: nodePublishSecretNamespaceKey,
+		secretNameKey:                prefixedNodePublishSecretNameKey,
+		secretNamespaceKey:           prefixedNodePublishSecretNamespaceKey,
+	}
+
+	controllerPublishSecretParams = deprecatedSecretParamsMap{
+		name:                         "ControllerPublish",
+		deprecatedSecretNameKey:      controllerPublishSecretNameKey,
+		deprecatedSecretNamespaceKey: controllerPublishSecretNamespaceKey,
+		secretNameKey:                prefixedControllerPublishSecretNameKey,
+		secretNamespaceKey:           prefixedControllerPublishSecretNamespaceKey,
+	}
+
+	nodeStageSecretParams = deprecatedSecretParamsMap{
+		name:                         "NodeStage",
+		deprecatedSecretNameKey:      nodeStageSecretNameKey,
+		deprecatedSecretNamespaceKey: nodeStageSecretNamespaceKey,
+		secretNameKey:                prefixedNodeStageSecretNameKey,
+		secretNamespaceKey:           prefixedNodeStageSecretNamespaceKey,
+	}
 )
 
 // CSIProvisioner struct
@@ -415,12 +479,20 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 		return nil, err
 	}
 
+	fsTypesFound := 0
 	fsType := ""
 	for k, v := range options.Parameters {
-		switch strings.ToLower(k) {
-		case "fstype":
+		if strings.ToLower(k) == "fstype" {
 			fsType = v
+			fsTypesFound += 1
+			glog.Warningf(deprecationWarning("fstype", prefixedFsTypeKey, ""))
+		} else if k == prefixedFsTypeKey {
+			fsType = v
+			fsTypesFound += 1
 		}
+	}
+	if fsTypesFound > 1 {
+		return nil, fmt.Errorf("fstype specified in parameters with both \"fstype\" and \"%s\" keys", prefixedFsTypeKey)
 	}
 	if len(fsType) == 0 {
 		fsType = defaultFSType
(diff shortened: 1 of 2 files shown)


saad-ali:
I support this change. This reserves `StorageClass` parameters with keys starting with `csi...` for use by the CSI `external-provisioner` and prevents them from being passed through to the CSI driver in the `CreateVolume` call.

mkimuram:
@davidz627 

Is my understanding right that the issue is caused by that csi external provisioner passes all the storage class parameters to csi driver's ```CreateVolume```, as a result, parameters that are not specific to csi driver, such as ```fsType``` and secrets for csi,  is passed to csi driver?

As [spec describes as below](https://github.com/container-storage-interface/spec/blame/master/spec.md#L743), I agree that csi plugins should check if invalid parameters are not passed like [this](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blame/master/pkg/gce-pd-csi-driver/controller.go#L103), and then above behavior will make csi driver regard such parameters as invalid.

```
  // Plugin specific parameters passed in as opaque key-value pairs.
  // This field is OPTIONAL. The Plugin is responsible for parsing and
  // validating these parameters. COs will treat these as opaque.
```
 
To solve this issue, we also have an option to strip ```fsType``` in addition to parameters that is prefixed with csi, instead of introducing new ```csiFsType```. Pros and Cons will be as below, so if we don't need to prioritize compatibility, implementation of this PR will be better. 

Pros:
  - Compatible with csi v0 driver on how to specify ```fstype``` in storage class parameters
  - Compatible with in-tree driver or non csi external provisioner on how to specify ```fstype``` in storage class parameters

Cons:
  - Introduce an irregular case for handling csi's storage class parameters (Assuming that this PR intends to make a rule that storage class parameters prefixed with "csi" are not for each csi driver in k8s.)

pkg/controller/controller.go
```diff
@@ -56,7 +56,35 @@ import (
...
+	secretNameKey                string
+	secretNamespaceKey           string
+}
+
 const (
+	// CSI Parameters that are put into fields and are stripped
+	// from Parameters passed to CreateVolume
+	csiParameterPrefix = "csi.storage.k8s.io/"
+
+	prefixedFsTypeKey = csiParameterPrefix + "FSType"
```
msau42: do we want these keys to be capitalized?

davidz627: @saad-ali what do you think, looks like the secret parameters were always camelcased. Whereas the old FS Type parameter was `fstype`.

Options:
- `fstype`
- `FsType`
- `fsType` [my preference]
- `FSType`
and I guess all the prefixed secret ones should be camel cased and not start with a capital.

davidz627:
@saad-ali @lpabon @jsafrane @msau42 @mkimuram @vladimirvivien 
new version based on morning CSI standup discussion pushed

pkg/controller/controller.go
```diff
@@ -82,6 +110,40 @@ const (
 	snapshotAPIGroup = snapapi.GroupName // "snapshot.storage.k8s.io"
 )
 
+var (
```
verult: What about a function
```golang
generateCSIParams(actor, key string, deprecated bool)
```
and
```golang
noActor = ""
provisionerActor = "Provisioner"
controllerPublishActor = "ControllerPublish"
...
secretNameKey = "SecretName"
fsTypeKey = "FSType"
...
actors = []string{provisionerActor, controllerPublishActor, ...}
keys = []string{secretNameKey, fsTypeKey, ...}
```

pkg/controller/controller.go
```diff
@@ -702,8 +803,52 @@ func (p *csiProvisioner) volumeHandleToId(handle string) string {
...
+	}
+	if t, ok := storageClassParams[secret.secretNameKey]; ok {
+		nameTemplate = t
+		numName += 1
+	}
+	if t, ok := storageClassParams[secret.secretNamespaceKey]; ok {
+		namespaceTemplate = t
+		numNamespace += 1
+	}
+
```
verult: What if you had a helper function
```golang
getKeyFromParams(actor, key string) (val string, isDeprecated bool, err error)
```
that
* Returns the val corresponding to either the deprecated or the existing key if only 1 of them exists (and set `isDeprecated` accordingly)
* Returns an error if both deprecated and existing keys exist

Then the check below becomes something like:
```golang
if secretNameIsDeprecated != secretNamespaceIsDeprecated {
   throwError()
} else if secretNameIsDeprecated {
   ...
} else { // !secretNameIsDeprecated && !secretNamespaceIsDeprecated
   ...
}
```

davidz627:
@saad-ali resolved. The error messages I made more generic because we can't tell if the user used the deprecated or new parameters, unless we want to add more cases :S

pkg/controller/controller.go
```diff
@@ -702,8 +805,52 @@ func (p *csiProvisioner) volumeHandleToId(handle string) string {
 	return handle
 }
 
-// getSecretReference returns a reference to the secret specified in the given nameKey and namespaceKey parameters, or an error if the parameters are not specified correctly.
-// if neither the name or namespace parameter are set, a nil reference and no error is returned.
+func getSecretNameAndNamespaceTemplate(secret secretParamsType, storageClassParams map[string]string) (nameTemplate, namespaceTemplate string, err error) {
```
verult: Going off the suggestions from above, this would take in `actor string` instead of `secret secretParamsType`

pkg/controller/controller.go
davidz627: how would we associate actors with keys? This ends up just shifting the variable definition into a different type of struct I think. It's unclear what params we're trying to generate here, whats the output of this function?

pkg/controller/controller.go
davidz627: when passing in a key we already know whether its deprecated or not as we define that so it doesnt make sense to me to have that as a return. Also given one deprecated theres no way to associate that with the "non-deprecated" version of that key without a struct that binds both together (or some hardcoding in this function) which serves the same purpose.

pkg/controller/controller.go
davidz627: this just shifts association of deprecatedness/actors from the structs that I created into hardcoded cases in the functions

pkg/controller/controller.go
verult: Sorry the func should be called `generateCSIParamKeys` and returns an actual storageclass parameter key. The logic would be the same as what you have right now, but encapsulated in a function, so it's easier to read and you don't need a giant list of variables.

verult: I'm trying to see if there's a way to not have a hard-coded list of all the various combinations.

pkg/controller/controller.go
verult: Tried to prototype this piece of logic and yeah, it doesn't look much prettier. OK to keep everything as is!

davidz627: could you LGTM?

pkg/controller/controller.go
```diff
@@ -590,6 +667,32 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 	return pv, nil
 }
 
+func stripPrefixedParameters(param map[string]string) (map[string]string, error) {
```
verult: nit: `remove` instead of `strip`? At first I was under the impression that the prefix is being stripped off of the parameter key.

pkg/controller/controller_test.go
```diff
@@ -149,6 +154,104 @@ func TestGetPluginName(t *testing.T) {
...
+			}
+		} else {
+			if tc.expectErr {
+				t.Fatalf("Did not get error when one was expected")
+			}
+		}
+
+		eq := reflect.DeepEqual(newParams, tc.expectedParams)
+		if !eq {
+			t.Fatalf("Stripped paramaters: %v not equal to expected paramaters: %v", newParams, tc.expectedParams)
```
verult: paramaters -> parameters

pkg/controller/controller.go
```diff
@@ -56,7 +56,37 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 )
 
+type secretParamsType struct {
```
jingxu97: maybe consider to change the name because it is a mapping from deprecated parameters to new ones instead of just secretParameter

pkg/controller/controller.go
```diff
@@ -590,6 +667,32 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
+			case prefixedControllerPublishSecretNamespaceKey:
+			case prefixedNodeStageSecretNameKey:
+			case prefixedNodeStageSecretNamespaceKey:
+			case prefixedNodePublishSecretNameKey:
+			case prefixedNodePublishSecretNamespaceKey:
+			default:
+				return map[string]string{}, fmt.Errorf("found unknown parameter key \"%s\" with reserved namespace %s", k, csiParameterPrefix)
+			}
+		} else {
+			// Don't strip, add this key-value to new map
```
verult: Leave a comment on why parameters previously prefixed with `csi` isn't being removed as well.

pkg/controller/controller.go
```diff
@@ -717,24 +864,19 @@ func (p *csiProvisioner) volumeHandleToId(handle string) string {
...
-		return nil, nil
+func getSecretReference(secretParams secretParamsType, storageClassParams map[string]string, pvName string, pvc *v1.PersistentVolumeClaim) (*v1.SecretReference, error) {
+	nameTemplate, namespaceTemplate, err := getSecretNameAndNamespaceTemplate(secretParams, storageClassParams)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get name and namespace template from params: %v", err)
 	}
-
-	if len(nameTemplate) == 0 || len(namespaceTemplate) == 0 {
-		return nil, fmt.Errorf("%s and %s parameters must be specified together", nameKey, namespaceKey)
+	if nameTemplate == "" && namespaceTemplate == "" {
```
jingxu97: you only need to check one empty string here, otherwise the previous function will return error.

pkg/controller/controller.go
jingxu97: maybe add comments about this function? It verify the parameters and then get them. How about rename to "verifyAndGet..." since one major part of this is verifying?

pkg/controller/controller.go
```diff
@@ -56,7 +56,37 @@ import (
...
 
+type secretParamsType struct {
+	name                         string
+	deprecatedSecretNameKey      string
+	deprecatedSecretNamespaceKey string
+	secretNameKey                string
+	secretNamespaceKey           string
+}
+
 const (
```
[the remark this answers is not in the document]
davidz627: refactor in future

pkg/controller/controller.go
verult: From my perspective it's easier to see what the check is doing this way

PR #183, opened 2018-12-05
Update Makefile to 1.0.1 version

msau42:


msau42:
This is not ideal. Let's just manually push to quay for now. @lpabon 
/close

Issue #164, opened 2018-11-05
Add rbac rules for accessing secrets


msau42:
They should be optional as not all volume drivers will need it: https://github.com/kubernetes/kubernetes/issues/69122

jarrpa:
This seems to be already resolved by #156. @msau42, is that right?

msau42:
Oh, I think actually we don't want to give secret permissions by default because the ClusterRole is very broad and not all volume plugins need it.

msau42:
I think it is ok to comment it out and add a comment with a description of when you need to enable it.

msau42:
Fixed by #188
/close

PR #188, opened 2018-12-10
Don't provide access to Secrets in default RBAC

jarrpa:
Related: #164

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

### DIFF
--- a/deploy/kubernetes/rbac.yaml
+++ b/deploy/kubernetes/rbac.yaml
@@ -21,9 +21,11 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: external-provisioner-runner
 rules:
-  - apiGroups: [""]
-    resources: ["secrets"]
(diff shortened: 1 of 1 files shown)


jarrpa:
@msau42 PTAL

deploy/kubernetes/rbac.yaml
```diff
@@ -21,9 +21,10 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: external-provisioner-runner
 rules:
-  - apiGroups: [""]
-    resources: ["secrets"]
-    verbs: ["get", "list"]
+  # Some provisioners may need to uncomment the following rule.
```
msau42: Maybe something like "The following rule should be uncommented for plugins that require secrets for provisioning"

jarrpa: Comment updated.

PR #208, opened 2019-01-06
Remove some constants from features.go

humblec:

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -19,8 +19,6 @@ package features
 import utilfeature "k8s.io/apiserver/pkg/util/feature"
 
 const (
-	// owner: @verult
-	// alpha: v0.4
 	Topology utilfeature.Feature = "Topology"
 )
 


humblec:
@verult Please take a look at this PR. I have removed an uncommon practice in kubernetes repo with this commit.

pkg/features/features.go
```diff
@@ -19,8 +19,6 @@ package features
 import utilfeature "k8s.io/apiserver/pkg/util/feature"
 
 const (
-	// owner: @verult
```
msau42: I think we should keep this so we know who are the feature owners like in k/k

humblec: @msau42 dont we record it in `OWNERS` file instead of individual source file? I thought thats the general practice.

davidz627: It's not a file-by-file ownership as much as a feature ownership that can span several files/packages

humblec: @davidz627 not sure I got your last comment . Are you saying invidual file mention  of owner is fair enough ?

davidz627: What I'm saying is
`OWNERS` file -> specifies ownership over some file or folders (related files)
`features.go` comments -> specifies ownership over some feature that may span several unrelated files, and does not imply ownership over the files in general

humblec:
@msau42 @davidz627 thanks. Closing this PR based on the review!

PR #209, opened 2019-01-08
update kubernetes-sigs/sig-storage-lib-external-provisioner to fix #195

### LINKED ISSUE #195: PV Protection breaks deleting
If a user deletes a PV while it has a PVC with a pv protection finalizer the CSI never gets alerted to delete the volume when the pvc is deleted.

rmb938:
Fixes #195

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,80 +3,103 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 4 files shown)


rmb938:
Thanks @msau42. Would it be possible to get a release and a new container image built with this change?

msau42:
Can you do a cherry-pick to release-1.0 branch, and then we can cut a new patch release

rmb938:
Sure

PR #186, opened 2018-12-09
Remove provisioner argument, update README

jarrpa:
Resolves #162 

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

PR #215, opened 2019-01-17
Cache driver name

jsafrane:
Provisioner should not ask for it every time, it should not change during driver container lifetime.

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -19,26 +19,25 @@ package main
 import (
 	goflag "flag"
 	"fmt"
-	flag "github.com/spf13/pflag"
 	"math/rand"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
+	flag "github.com/spf13/pflag"
+	"google.golang.org/grpc"
 
 	ctrl "github.com/kubernetes-csi/external-provisioner/pkg/controller"
 	snapclientset "github.com/kubernetes-csi/external-snapshotter/pkg/client/clientset/versioned"
 	"github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/controller"
 	csiclientset "k8s.io/csi-api/pkg/client/clientset/versioned"
 
-	"google.golang.org/grpc"
-
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"strings"
 
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -339,14 +336,15 @@ func NewCSIProvisioner(client kubernetes.Interface,
...
+		driverName:           driverName,
 	}
 	return provisioner
 }
 
 // This function get called before any attempt to communicate with the driver.
 // Before initiating Create/Delete API calls provisioner checks if Capabilities:
 // PluginControllerService,  ControllerCreateVolume sre supported and gets the  driver name.
-func checkDriverState(grpcClient *grpc.ClientConn, timeout time.Duration, needSnapshotSupport bool) (*driverState, error) {
+func checkDriverCapabilities(grpcClient *grpc.ClientConn, timeout time.Duration, needSnapshotSupport bool) (sets.Int, error) {
```
davidz627: Why are we representing capabilities with `sets.Int`, is there no named type from the protobuf we can use? if not, can we define one
`type PluginCapabilitiesType sets.Int`

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -145,7 +144,7 @@ func init() {
 
 	// Create the provisioner: it implements the Provisioner interface expected by
 	// the controller
-	csiProvisioner := ctrl.NewCSIProvisioner(clientset, csiAPIClient, *csiEndpoint, *connectionTimeout, identity, *volumeNamePrefix, *volumeNameUUIDLength, grpcClient, snapClient)
+	csiProvisioner := ctrl.NewCSIProvisioner(clientset, csiAPIClient, *csiEndpoint, *connectionTimeout, identity, *volumeNamePrefix, *volumeNameUUIDLength, grpcClient, snapClient, provisionerName)
```
davidz627: `provisionerName == driverName`? why do we even have these two concepts then. Can we just call it `driverName` everywhere?

jsafrane: I think it's mandatory for volume topology, where kubelet knows just driver name from driver registrar and scheduler knows only provider name from StorageClass. @msau42 knows details.

pkg/controller/controller.go
jsafrane: I tried, but then I loose all `sets.Int` methods (no `PluginCapabilitiesType.Has`). I could restore `driverState` struct (with a better name) with `sets.Int` as a field, but that looks ugly.

cmd/csi-provisioner/csi-provisioner.go
msau42: Yes, they're the same.  external-provisioner originally was implemented allowing them to be different but there's no reason they need to be so we removed it in https://github.com/kubernetes-csi/external-provisioner/pull/186

Issue #101, opened 2018-06-12
Exponential backoff for VolumeCreate issues


msau42:
1. If PVC is deleted, we don't exit the exponential backoff [loop](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L342)
2. Non timeout errors are not retried within the Provision() call, but won't Provision() be retired by the external-provisioner controller with no exponential backoff?

This probably requires change to external-provisioner library

Issue #154, opened 2018-10-19
PVC stuck in pending state after provisioner retry exceeds threshold


leakingtapan:
/kind bug

When CSI driver returns error on `CreateVolume`, the same operation will be retried by external-provisioner. However, after provisioner retry threshold is exceeded, the pvc is stuck in pending state:
```
>> kk get pvc
NAME            STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   AGE
voltype-claim   Pending                                      voltype-sc     6h17m

>> kk describe pvc
Name:          voltype-claim
Namespace:     default
StorageClass:  voltype-sc
Status:        Pending
Volume:
Labels:        <none>
Annotations:   volume.beta.kubernetes.io/storage-provisioner: com.amazon.aws.csi.ebs
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:
Access Modes:
Events:
  Type       Reason                Age                     From                         Message
  ----       ------                ----                    ----                         -------
  Normal     ExternalProvisioning  79s (x1524 over 6h21m)  persistentvolume-controller  waiting for a volume to be created, either by external provisioner "com.amazon.aws.csi.ebs" or manually created by system administrator
Mounted By:  <none>
```
provisioner log:
```
I1019 01:09:44.103840       1 controller.go:927] provision "default/voltype-claim" class "voltype-sc": started
I1019 01:09:44.116655       1 controller.go:124] GRPC call: /csi.v0.Identity/GetPluginCapabilities
I1019 01:09:44.116671       1 controller.go:125] GRPC request:
I1019 01:09:44.116828       1 event.go:221] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"voltype-claim", UID:"a8d939de-d324-11e8-81f1-0a75e9a76798", APIVersion:"v1", ResourceVersion:"539476", FieldPath:""}): type: 'Normal' reason: 'Provisioning' External provisioner is provisioning volume for claim "default/voltype-claim"
I1019 01:09:44.117364       1 controller.go:127] GRPC response: capabilities:<service:<type:CONTROLLER_SERVICE > > capabilities:<service:<type:ACCESSIBILITY_CONSTRAINTS > >
I1019 01:09:44.117398       1 controller.go:128] GRPC error: <nil>
I1019 01:09:44.117408       1 controller.go:124] GRPC call: /csi.v0.Controller/ControllerGetCapabilities
[...]

PR #216, opened 2019-01-17
Remove internal retry loop

jsafrane:
We should use exponential backoff from the provisioner library. It will react more quickly when a PVC is deleted and it won't occupy a worker thread.

I verified that #63 is still fixed, the provisioner consistently uses `pvc-<uid>` as CO-supplied volume name for idempotency.

Fixes: #101 #154

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -40,14 +40,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/connectivity"
-	"google.golang.org/grpc/status"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	csiclientset "k8s.io/csi-api/pkg/client/clientset/versioned"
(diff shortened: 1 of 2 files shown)


jsafrane:
/hold
waiting for rebase with  https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/12

humblec:
/lgtm. 
Also workqueue.Ratelimiter patch is merged, so I believe this is good to go!

jsafrane:
Rebased + added parameters for tighter retry loop.

@msau42 @pohly, PTAL

pkg/controller/controller.go
```diff
@@ -566,26 +563,9 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
-			if status.Code() == codes.DeadlineExceeded {
-				// CreateVolume timed out, give it another chance to complete
-				klog.Warningf("CreateVolume timeout: %s has expired, operation will be retried", p.timeout.String())
-				return false, nil
-			}
-		}
-		// CreateVolume failed , no reason to retry, bailing from ExponentialBackoff
-		return false, err
-	})
+	ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
```
msau42: this is still the rpc connection timeout value right?  Do we want to have a separate timeout value for the CreateVolume call?  For some drivers, the provisioning operation could take minutes.

jsafrane: No, the value is not right. I'd like to address it in a separate PR.

jsafrane: Filled https://github.com/kubernetes-csi/external-provisioner/issues/227

jsafrane:
/hold cancel

Regarding README, the current one is not very helpful. Filled https://github.com/kubernetes-csi/external-provisioner/issues/228

Issue #172, opened 2018-11-28
External provisioner misses VolumeSource support for CLONE_VOLUMES 


humblec:
CSI 1.0.0 has VolumeSource struct in VolumeContentSource for supporting cloning volumes (https://github.com/container-storage-interface/spec/commit/7a8d2d546db2b010d924708ab36f5dd43a10098d) , but the required changes are not there in the  external-provisioner . 

This is the template we used for testing, am i missing anything here?

> apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: glusterfs-pv-restore
spec:
  storageClassName: glusterfs-csi
  dataSource:
    name: gcs-example-volume
    kind: VolumeSource
    apiGroup: snapshot.storage.k8s.io 
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 5Gi
>

The error we receive.

~~~
[root@master ~]# kubectl create -f vol-csi.yaml 
The PersistentVolumeClaim "glusterfs-pv-restore" is invalid: spec.dataSource: Unsupported value: "VolumeContentSource.snapshot.storage.k8s.io": supported values: "VolumeSnapshot.snapshot.storage.k8s.io" [root@master ~]# vi vol-csi.yaml [root@master ~]# kubectl create -f vol-csi.yaml The PersistentVolumeClaim "glusterfs-pv-restore" is invalid: spec.dataSource: Unsupported value: "VolumeSource.snapshot.storage.k8s.io": supported values: "VolumeSnapshot.snapshot.storage.k8s.io">
~~~

humblec:
kubernetes used for testing
~~~
[root@master ~]# kubectl version Client Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.0-beta.2", GitCommit:"a4ff09c41589c48547e04f85391aa5610ebe0e17", GitTreeState:"clean", BuildDate:"2018-11-23T00:55:34Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"} Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.0-beta.2", GitCommit:"a4ff09c41589c48547e04f85391aa5610ebe0e17", GitTreeState:"clean", BuildDate:"2018-11-23T00:47:47Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}
~~~

saad-ali:
[...]

PR #220, opened 2019-01-29
Extend allowed DataSources

j-griffith:
This PR extends the supported DataSources handled by the
external-provisioner to include PVCs (Clone) and Populators (external
CRDs).

It removes the restriction of of the provisioner erroring when a
DataSource.Kind of anything other than VolumeSnapshot is supplied (the
Kubernetes API currently enforces this anyway via it's verifictaion
checks) and adds handling for a DataSource.Kind of
PersistentVolumeClaim.

Addresses: Issue #172

Does this PR introduce a user-facing change?:

```release-note
Adds the ability to handle PVC as a DataSource to enable cloning for plugins that support it.
 ```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -105,7 +105,8 @@ const (
 	defaultFSType = "ext4"
 
 	snapshotKind     = "VolumeSnapshot"
-	snapshotAPIGroup = snapapi.GroupName // "snapshot.storage.k8s.io"
+	snapshotAPIGroup = snapapi.GroupName       // "snapshot.storage.k8s.io"
+	pvcKind          = "PersistentVolumeClaim" // Native types don't require an API group
 
 	tokenPVNameKey       = "pv.name"
 	tokenPVCNameKey      = "pvc.name"
@@ -145,6 +146,12 @@ var (
 		secretNamespaceKey:           prefixedNodeStageSecretNamespaceKey,
 	}
 )
+
+// requiredCapabilities provides a set of extra capabilities required for special/optional features provided by a plugin
+type requiredCapabilities struct {
+	snapshot bool
+	clone    bool
+}
 
 // CSIProvisioner struct
 type csiProvisioner struct {
@@ -242,7 +249,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 // This function get called before any attempt to communicate with the driver.
 // Before initiating Create/Delete API calls provisioner checks if Capabilities:
 // PluginControllerService,  ControllerCreateVolume sre supported and gets the  driver name.
-func (p *csiProvisioner) checkDriverCapabilities(needSnapshotSupport bool) error {
+func (p *csiProvisioner) checkDriverCapabilities(rc *requiredCapabilities) error {
 	if !p.pluginCapabilities[csi.PluginCapability_Service_CONTROLLER_SERVICE] {
 		return fmt.Errorf("CSI driver does not support dynamic provisioning: plugin CONTROLLER_SERVICE capability is not reported")
 	}
@@ -251,11 +258,18 @@ func (p *csiProvisioner) checkDriverCapabilities(needSnapshotSupport bool) error
 		return fmt.Errorf("CSI driver does not support dynamic provisioning: controller CREATE_DELETE_VOLUME capability is not reported")
 	}
 
-	if needSnapshotSupport {
+	if rc.snapshot {
 		// Check whether plugin supports create snapshot
 		// If not, create volume from snapshot cannot proceed
 		if !p.controllerCapabilities[csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT] {
 			return fmt.Errorf("CSI driver does not support snapshot restore: controller CREATE_DELETE_SNAPSHOT capability is not reported")
+		}
+	}
+	if rc.clone {
+		// Check whether plugin supports clone operations
+		// If not, create volume from pvc cannot proceed
+		if !p.controllerCapabilities[csi.ControllerServiceCapability_RPC_CLONE_VOLUME] {
+			return fmt.Errorf("CSI driver does not support clone operations: controller CLONE_VOLUME capability is not reported")
 		}
 	}
 
@@ -356,22 +370,27 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
 		}
 	}
 
-	var needSnapshotSupport bool
+	// Make sure the plugin is capable of fulfilling the requested options
+	rc := &requiredCapabilities{}
 	if options.PVC.Spec.DataSource != nil {
 		// PVC.Spec.DataSource.Name is the name of the VolumeSnapshot API object
 		if options.PVC.Spec.DataSource.Name == "" {
 			return nil, fmt.Errorf("the PVC source not found for PVC %s", options.PVC.Name)
 		}
-		if options.PVC.Spec.DataSource.Kind != snapshotKind {
-			return nil, fmt.Errorf("the PVC source is not the right type. Expected %s, Got %s", snapshotKind, options.PVC.Spec.DataSource.Kind)
-		}
-		if *(options.PVC.Spec.DataSource.APIGroup) != snapshotAPIGroup {
-			return nil, fmt.Errorf("the PVC source does not belong to the right APIGroup. Expected %s, Got %s", snapshotAPIGroup, *(options.PVC.Spec.DataSource.APIGroup))
-		}
-		needSnapshotSupport = true
-	}
 
-	if err := p.checkDriverCapabilities(needSnapshotSupport); err != nil {
+		switch options.PVC.Spec.DataSource.Kind {
+		case snapshotKind:
+			if *(options.PVC.Spec.DataSource.APIGroup) != snapshotAPIGroup {
+				return nil, fmt.Errorf("the PVC source does not belong to the right APIGroup. Expected %s, Got %s", snapshotAPIGroup, *(options.PVC.Spec.DataSource.APIGroup))
+			}
+			rc.snapshot = true
+		case pvcKind:
+			rc.clone = true
+		default:
+			klog.Infof("Unsupported DataSource specified (%s), the provisioner won't act on this request", options.PVC.Spec.DataSource.Kind)
+		}
+	}
+	if err := p.checkDriverCapabilities(rc); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/controller/controller.go [hunk 1]
```diff
@@ -354,17 +365,22 @@ func checkDriverCapabilities(grpcClient *grpc.ClientConn, timeout time.Duration,
...
 		if !capabilities.Has(ControllerCapability_CREATE_DELETE_SNAPSHOT) {
 			return nil, fmt.Errorf("no create/delete snapshot support detected. Cannot create volume from snapshot")
 		}
 	}
 
+	if caps.clone {
+		// Check whether plugin supports create clone
+		// If not, return error
+		if !capabilities.Has(ControllerCapability_CLONE_VOLUME) {
+			return nil, fmt.Errorf("no create/delete snapshot support detected. Cannot create volume from snapshot")
```
rollandf: The error should be updated
"no clone volume support detected. Cannot create volume from volume"

pkg/controller/controller.go
```diff
@@ -665,7 +686,25 @@ func removePrefixedParameters(param map[string]string) (map[string]string, error
...
+		return p.getSnapshotSource(options)
+	case pvcKind:
+		return p.getPVCSource(options)
+	default:
+		// For now we shouldn't pass other things to this function, but treat it as a noop and extend as needed
+		return nil, nil
+	}
+}
+
+// getSnapshotSource verifies DataSource.Kind of tyep VolumeSnapshot, making sure that the requested Snapshot is available/ready
```
rollandf: tyep->type

pkg/controller/controller.go
```diff
@@ -720,13 +759,50 @@ func (p *csiProvisioner) getVolumeContentSource(options controller.VolumeOptions
 	return volumeContentSource, nil
 }
 
+// getPVCSource verifies DataSource.Kind of tyep PersistentVolumeClaim, making sure that the requested PVC is available/ready
```
rollandf: tyep->type

pkg/controller/controller.go
```diff
@@ -720,13 +759,50 @@ func (p *csiProvisioner) getVolumeContentSource(options controller.VolumeOptions
...
+// returns the VolumeContentSource for the requested PVC
+func (p *csiProvisioner) getPVCSource(options controller.VolumeOptions) (*csi.VolumeContentSource, error) {
+	sourcePVC, err := p.client.CoreV1().PersistentVolumeClaims(options.PVC.Namespace).Get(options.PVC.Spec.DataSource.Name, metav1.GetOptions{})
+	if err != nil {
+		fmt.Printf("JDG: Well that blows: %v\n", err)
```
rollandf: I guess this printf  should go out

j-griffith: Definitely :)

pkg/controller/controller_test.go
```diff
@@ -1891,3 +1925,137 @@ func TestProvisionWithMountOptions(t *testing.T) {
...
 }
+
+// TestProvisionFromPVC tests create volume from snapshot
+func TestProvisionFromPVC(t *testing.T) {
+	var requestedBytes int64 = 1000
+	var pvcName = "src-pvc"
+	//var timeNow = time.Now().UnixNano()
+	//var metaTimeNowUnix = &metav1.Time{
+	//		Time: time.Unix(0, timeNow),
+	//	}
```
humblec: @j-griffith can we remove this commented code ?

j-griffith:
Much has changed since I worked on this; fixing up the rebase is probable more hazardous than the change itself, I think I'll close this out and start with a clean slate.

pkg/controller/controller.go
```diff
@@ -591,7 +610,61 @@ func removePrefixedParameters(param map[string]string) (map[string]string, error
...
+}
+
+// getPVCSource verifies DataSource.Kind of type PersistentVolumeClaim, making sure that the requested PVC is available/ready
+// returns the VolumeContentSource for the requested PVC
+func (p *csiProvisioner) getPVCSource(options controller.VolumeOptions) (*csi.VolumeContentSource, error) {
+	sourcePVC, err := p.client.CoreV1().PersistentVolumeClaims(options.PVC.Namespace).Get(options.PVC.Spec.DataSource.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("error getting PVC %s from api server: %v", options.PVC.Spec.DataSource.Name, err)
+	}
+	if string(sourcePVC.Status.Phase) != "Bound" {
```
humblec: @j-griffith  I dont know whether this can land on a race , that said, by the time `getPVCSource` returns its `available/ready` , someone released it or vice versa. @jsafrane thoughts?

pkg/controller/controller.go
```diff
@@ -353,22 +367,27 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
 		if options.PVC.Spec.DataSource.Name == "" {
 			return nil, fmt.Errorf("the PVC source not found for PVC %s", options.PVC.Name)
 		}
-		if options.PVC.Spec.DataSource.Kind != snapshotKind {
-			return nil, fmt.Errorf("the PVC source is not the right type. Expected %s, Got %s", snapshotKind, options.PVC.Spec.DataSource.Kind)
-		}
-		if *(options.PVC.Spec.DataSource.APIGroup) != snapshotAPIGroup {
-			return nil, fmt.Errorf("the PVC source does not belong to the right APIGroup. Expected %s, Got %s", snapshotAPIGroup, *(options.PVC.Spec.DataSource.APIGroup))
+
+		switch options.PVC.Spec.DataSource.Kind {
```
humblec: @j-griffith  we already have a `getVolumeContentSource`, can we make use of it?

pkg/controller/controller.go
```diff
@@ -591,7 +610,61 @@ func removePrefixedParameters(param map[string]string) (map[string]string, error
...
+	if sourcePVC.ObjectMeta.DeletionTimestamp != nil {
+		return nil, fmt.Errorf("the PVC DataSource %s is currently being deleted", options.PVC.Spec.DataSource.Name)
+	}
+	if sourcePVC.Spec.StorageClassName != options.PVC.Spec.StorageClassName {
+		return nil, fmt.Errorf("the source PVC and destination PVCs must be in the same storage class for cloning.  Source is in %v, but new PVC is in %v",
+			sourcePVC.Spec.StorageClassName, options.PVC.Spec.StorageClassName)
+	}
+	capacity := options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
+	requestedSize := capacity.Value()
+	if requestedSize < int64(sourcePVC.Spec.Size()) {
```
humblec: @j-griffith  There could be scenarios where `resizer` is processing an expansion request in the backend. So, may be checking current PVC's size and capacity could help as well.

pkg/controller/controller.go
j-griffith: Hi @humblec thanks for pointing that out.  Actually you'll notice I *do* utilize that helper function, I added the fetch for PVC to it and call it when we actually pull the content.

This particular check in this section was a here to validate the options and "fail fast" if they were invalid, as well as set a variable so that the provision stage later would have a flag to "getContentSource" or not.  If we want to consolidate things into a single helper, that's certainly possible, but it seemed like following the flow we already have for Snapshots was a good idea here.

pkg/controller/controller.go
j-griffith: Yes, I think that's possible, I guess you'd want to add a shared mutex for the source PVC here?

pkg/controller/controller.go
j-griffith: Sorry, i'm not sure I follow; what do you mean the "current PVC's size and capacity"?  The destination has yet to be created at this stage, so the only object that's relevant is the sourcePVC and the requested size; do you mean a resize in progress on the specified source volume?

pkg/controller/controller_test.go
j-griffith: Oops, yes sorry!

pkg/controller/controller.go
humblec: >do you mean a resize in progress on the specified source volume?
Exactly, the resizer took the request and working with the driver to get it into desired state.

pkg/controller/controller.go
```diff
@@ -353,22 +367,27 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
...
+		switch options.PVC.Spec.DataSource.Kind {
+		case snapshotKind:
+			if *(options.PVC.Spec.DataSource.APIGroup) != snapshotAPIGroup {
+				return nil, fmt.Errorf("the PVC source does not belong to the right APIGroup. Expected %s, Got %s", snapshotAPIGroup, *(options.PVC.Spec.DataSource.APIGroup))
+			}
+			rc.snapshot = true
+		case pvcKind:
+			rc.clone = true
+		default:
+			klog.Infof("non csi DataSource specified (%s), the provisioner won't act on this request", options.PVC.Spec.DataSource.Kind)
```
saad-ali: `non csi DataSource specified` -> `Unsupported DataSource`

pkg/controller/controller.go
saad-ali: Shouldn't checking `sourcePVC.Spec` alone be sufficient since it will be the largest possible value (it will be the desired size)? Validation prevents setting that value to a smaller value right?

pkg/controller/controller.go
j-griffith: Yeah, that's better; changed.

pkg/controller/controller.go
j-griffith: > @j-griffith There could be scenarios where `resizer` is processing an expansion request in the backend. So, may be checking current PVC's size and capacity could help as well.

@humblec so I've thought about this a bit; and yes there's certainly a potential for a race here.  That being said, we have that condition across other operations already (snapshots for example) and the challenge that's been discussed is coordinating any sort of locking/checks across the different control planes.  Not to mention as @jsafrane pointed out, even across CSI provisioner nodes.

IMO this is a larger problem that needs solved independent of the cloning implementation.  Whether that's new phases on the PVC, some other even tracking, or (as I think we've done thus far) the "storage device plugins problem".  It sounds like this is something that's been brought up in other context and is being discussed, but we have some long term work to do here I think.

humblec:
@j-griffith as discussed in above review comments (https://github.com/kubernetes-csi/external-provisioner/pull/220#discussion_r289537735) i see possibilities of couple of race conditions ( pvc status/capacity..etc) , however these are not really specific to this PR rather its  a general problem which we have to address with a thorough design. Lets attack it later in a different discussion/PR.  LGTM from my side on this PR . Thanks !!

@saad-ali I am good to get this in and thanks for giving chance for final review from my side!

j-griffith:
> @j-griffith as discussed in above review comments ([#220 (comment)](https://github.com/kubernetes-csi/external-provisioner/pull/220#discussion_r289537735)) i see possibilities of couple of race conditions ( pvc status/capacity..etc) , however these are not really specific to this PR rather its a general problem which we have to address with a thorough design. Lets attack it later in a different discussion/PR. LGTM from my side on this PR . Thanks !!
> 
> @saad-ali I am good to get this in and thanks for giving chance for final review from my side!

Thanks @humblec I agree with you, we'll need to consider these in a broader scope.

j-griffith:
Had to update the call I added to NewCSIProvisioner in the test file after rebase.

PR #219, opened 2019-01-28
Migrate to k8s.io/klog from glog.

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

PR #217, opened 2019-01-18
Update to external provisioner lib v2.1.0

wongma7:
@jsafrane for your RateLimiter changes

PR #225, opened 2019-02-07
Bump provisioner lib to 2.2.0 + use klog

jsafrane:
Use new provisioner library that uses klog and replace glog with klog.


jsafrane:
Replaces https://github.com/kubernetes-csi/external-provisioner/pull/219 and https://github.com/kubernetes-csi/external-provisioner/pull/217 in one go.

humblec:
>As @humblec doesn't mind, let's merge this.

@pohly thanks for considering it ! I am fine and especially when it comes from @jsafrane :)

PR #229, opened 2019-02-08
Rename deprecatedSecretParamsMap

### LINKED ISSUE #224: Rename deprecatedSecretParamsMap to something else because it also contains non-deprecated fields
As subject says, this struct is bit oddly named and appears as if entire struct is deprecated.

j-griffith:
The struct name deprecatedSecretParamsMap is misleading, it includes the
"deprecated" prefix hinting that the struct it deprecated, however it's
not.  It also isn't made up of strictly deprecated secret keys (which
are in fact labeled with the "deprecated" prefix).

Given this a private struct, it's simple to just rename it, this change
renames it to simply "secretParamsMap" and adds a comment to the struct
to give at least a hint as to the intent.

resolves #224

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -111,6 +111,8 @@ const (
 
 	snapshotKind     = "VolumeSnapshot"
 	snapshotAPIGroup = snapapi.GroupName // "snapshot.storage.k8s.io"
+
+	pvcKind = "PersistentVolumeclaim" // Native types don't take an API group
 )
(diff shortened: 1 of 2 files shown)


msau42:
renaming lgtm, but can you remove the commit for the pvc source?

j-griffith:
Let's see if somebody new wants to grab this and submit it, otherwise I'll rebase and submit; not like this really hurts anything.

PR #234, opened 2019-02-22
Use Connection() from util package

jsafrane:
The result should be the same as in https://github.com/kubernetes-csi/external-attacher/pull/123

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -20,12 +20,12 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"net"
 	"os"
 	"strings"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/kubernetes-csi/csi-lib-utils/connection"
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
 	"github.com/kubernetes-csi/external-provisioner/pkg/features"
 	snapapi "github.com/kubernetes-csi/external-snapshotter/pkg/apis/volumesnapshot/v1alpha1"
@@ -46,7 +46,6 @@ import (
 	"k8s.io/klog"
 
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/connectivity"
 )
 
 type deprecatedSecretParamsMap struct {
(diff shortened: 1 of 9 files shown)


pkg/controller/controller.go
```diff
@@ -185,55 +184,18 @@ func logGRPC(ctx context.Context, method string, req, reply interface{}, cc *grp
...
-	}
-	name := rsp.GetName()
-	if name == "" {
-		return "", fmt.Errorf("name is empty")
-	}
-	return name, nil
+	return connection.GetDriverName(ctx, conn)
 }
 
 func getDriverCapabilities(conn *grpc.ClientConn, timeout time.Duration) (sets.Int, error) {
```
msau42: and update getcapabilities?

jsafrane: added

msau42:
cc @lpabon

pkg/controller/controller.go
```diff
@@ -248,30 +210,16 @@ func getDriverCapabilities(conn *grpc.ClientConn, timeout time.Duration) (sets.I
...
-		if cap == nil {
-			continue
-		}
-		service := cap.GetService()
-		if service == nil {
-			continue
-		}
-		switch service.GetType() {
+	for cap := range pluginCaps {
+		switch cap {
```
msau42: Is mixing plugin and controller capabilities into the same set correct?  Could you have conflicts in values?

jsafrane: `PluginCapability_*` and `ControllerCapability_*` are part of the same enum defined in the provisioner, they don't come from CSI:

https://github.com/kubernetes-csi/external-provisioner/blob/e63995ffb48736dc6275b5e8702e2dc8868c57ef/pkg/controller/controller.go#L160-L165

msau42: Thanks for pointing that out. Is there a reason why it needs to have one more level of translation from CSI capabilities?

jsafrane: Can we sort it out in a separate PR? I'd like to cache the capabilities so provisioner does not load them from the driver every time it needs them. Refactoring of the enum would be better there.

msau42: Sure do you want to use csi-lib-utils 0.4.0-rc1 here?

jsafrane: Updated to 0.4.0-rc1

Issue #231, opened 2019-02-17
Missing permission: csi-provisioner cannot update resource "persistentvolumes"


darkowlzz:
I was running the canary version of the provisioner with the hostpath driver and saw these in the logs:

```
I0217 14:13:45.598432       1 controller.go:1189] delete "pvc-881d9151-32bb-11e9-ba24-024270471d6d": started
I0217 14:13:45.650957       1 controller.go:1217] delete "pvc-881d9151-32bb-11e9-ba24-024270471d6d": volume deleted
I0217 14:13:45.670198       1 controller.go:1254] delete "pvc-881d9151-32bb-11e9-ba24-024270471d6d": failed to remove finalizer for persistentvolume: persistentvolumes "pvc-881d9151-32bb-11e9-ba24-024270471d6d" is forbidden: User "system:serviceaccount:default:csi-provisioner" cannot update resource "persistentvolumes" in API group "" at the cluster scope
W0217 14:13:45.670245       1 controller.go:793] Retrying syncing volume "pvc-881d9151-32bb-11e9-ba24-024270471d6d", failure 10
E0217 14:13:45.670305       1 controller.go:811] error syncing volume "pvc-881d9151-32bb-11e9-ba24-024270471d6d": persistentvolumes "pvc-881d9151-32bb-11e9-ba24-024270471d6d" is forbidden: User "system:serviceaccount:default:csi-provisioner" cannot update resource "persistentvolumes" in API group "" at the cluster scope
```

Got the rbac rules from [csi docs example page](https://kubernetes-csi.github.io/docs/example.html#create-rbac-rules-for-csi-provisioner).

Adding "update" verb to [persistentvolumes](https://github.com/kubernetes-csi/external-provisioner/blob/1cd1c20a6d4b2fcd25c98a008385b436d61d46a4/deploy/kubernetes/rbac.yaml#L29) resource rule under the ClusterRole fixed it.

```
I0217 14:16:07.266861       1 controller.go:1189] delete "pvc-7b475573-32bc-11e9-ba24-024270471d6d": started
I0217 14:16:07.320841       1 controller.go:1217] delete "pvc-7b475573-32bc-11e9-ba24-024270471d6d": volume deleted
I0217 14:16:07.331645       1 controller.go:1261] delete "pvc-7b475573-32bc-11e9-ba24-024270471d6d": persistentvolume deleted
[...]

PR #236, opened 2019-02-25
Fix rbac

cwdsuzhou:
This PR is for fixing #231.
The issue also existed in other versions because pkg/controller calls  the function "deleteVolumeOperation" in "github.com/kubernetes-sigs/sig-storage-lib-external-provisioner" or "github.com/kubernetes-incubator/external-storage/lib/controller" where may update PersistentVolumes.


msau42:
Thanks for the fix!  Can you also check if release-1.0 has this problem, and if so, cherry pick this fix?

Issue #221, opened 2019-01-31
Wrong AccessibilityRequirement passed in CreateVolumeRequest


avalluri:
My csi driver deals with storage local to Node, hence i would like to limit topology segment to individual node, by passing as below in NodeInfoGetRespose.
```
AccessibleTopology: &csi.Topology{
     Segments: map[string]string{
        "pmem-csi/node": nodeID,
},
```

I am using delay binding(WaitForFirstConsumer) so that volume get provisioned only after scheduler picks the node for scheduling pod that claims the PV.

My expectation was`'CreateVolumeRequest.AccessibileRequirement` is filled with the topology of selected node. But its filled with all the nodes in cluster where the driver running.

```
GRPC call: /csi.v1.Controller/CreateVolume
I0131 11:23:54.079912       1 glog.go:58] GRPC request: name:"pvc-ab1837f9-254a-11e9-8f0d-deadbeef0100" capacity_range:<required_bytes:8589934592 > volume_capabilities:<mount:<fs_type:"ext4" > access_mode:<mode:SINGLE_NODE_WRITER > > accessibility_requirements:<requisite:<segments:<key:"pmem-csi/node" value:"host-3" > > requisite:<segments:<key:"pmem-csi/node" value:"host-1" > > requisite:<segments:<key:"pmem-csi/node" value:"host-2" > > preferred:<segments:<key:"pmem-csi/node" value:"host-1" > > preferred:<segments:<key:"pmem-csi/node" value:"host-2" > > preferred:<segments:<key:"pmem-csi/node" value:"host-3" > > >
```

My observation is that the code in aggregateTopologies()  is considering only TopologyKeys provided in NodeInfo of SelectedNode to select the Nodes with similar topology, where it's omitting the keyvalue,This results in ending up  with all the nodes(where my driver runs) as all nodes has topology constraints with the same key but with unique value.

```
`		// TODO (verult) retry
		selectedNodeInfo, err := csiAPIClient.CsiV1alpha1().CSINodeInfos().Get(selectedNode.Name, metav1.GetOptions{})
		if err != nil {
			// We must support provisioning if CSINodeInfo is missing, for backward compatibility.
			glog.Warningf("error getting CSINodeInfo for selected node %q: %v; proceeding to provision without topology information", selectedNode.Name, err)
			return nil, nil
		}
		topologyKeys = getTopologyKeys(selectedNodeInfo, driverName)
	}

	if len(topologyKeys) == 0 {
[...]

Issue #144, opened 2018-10-04
Consider using Node & CSINodeInfo informers


verult:
Currently the topology code makes individual API calls to fetch nodes and CSINodeInfo objects, and in many cases a selector is used to list a subset of objects rather than all of them. Should we use informers instead?

My take is to use an informer for CSINodeInfo objects but not Node objects. CSINodeInfo objects are only updated on CSI driver installation, which is less frequent than provision calls. Node objects are updated constantly, much more frequent than provision calls.

https://github.com/kubernetes-csi/external-provisioner/blob/393375a3af0b8fd4ea3cb9491586670c55a6eb25/pkg/controller/topology.go#L102
https://github.com/kubernetes-csi/external-provisioner/blob/393375a3af0b8fd4ea3cb9491586670c55a6eb25/pkg/controller/topology.go#L138
https://github.com/kubernetes-csi/external-provisioner/blob/393375a3af0b8fd4ea3cb9491586670c55a6eb25/pkg/controller/topology.go#L158
https://github.com/kubernetes-csi/external-provisioner/blob/393375a3af0b8fd4ea3cb9491586670c55a6eb25/pkg/controller/topology.go#L179

/cc @ddebroy

ddebroy:
Absolutely ... we can use a SharedIndexInformer for the delayed binding scenarios where we don't need to list nodes. I will take this up (will start assigning to myself once I am added as a member - pinged Saad).

verult:
Thanks so much Deep!

mucahitkurt:
I'm working on this issue. cc @msau42

msau42:
> My take is to use an informer for CSINodeInfo objects but not Node objects. CSINodeInfo objects are only updated on CSI driver installation, which is less frequent than provision calls. Node objects are updated constantly, much more frequent than provision calls.

I think Nodes are updated less frequently now that the heartbeats are in a separate object. It should be fine to have a Node informer now, and could help reduce API server load in large clusters if we don't have to List all nodes for every provision operation.

PR #238, opened 2019-03-02
Topology beta

msau42:
Move from alpha CRDs to beta in-tree CSINode object.

I am temporarily copying the new API files into vendor


README.md
```diff
@@ -1,49 +1,64 @@
...
+  the pod that the `external-provisioner` container will use to issue CSI
+  operations (e.g. `/csi/csi.sock`).
 
-```bash
-bin/csi-provisioner --kubeconfig /var/run/kubernetes/admin.kubeconfig --alsologtostderr
-```
+### Common arguments
 
-### Create Storage class, PVC, and Pod
+* `--timeout`: Number of nanoseconds before a CreateVolume or DeleteVolume operation
```
msau42: @jsafrane I noticed the value is in nanoseconds. Is that what we want or would seconds be more reasonable?

jsafrane: It's `flag.Duration`, so it accepts suffixes like `s` for seconds. The description is wrong / misleading.

https://github.com/kubernetes-csi/external-provisioner/blob/e9387dc7f8079dd745c10dca185257dcf2994abb/cmd/csi-provisioner/csi-provisioner.go#L61

jsafrane: Btw, it would be probably better to rework whole README in a separate PR. I'd like to have there the same format / information as in attacher (https://github.com/kubernetes-csi/external-attacher/pull/130).

pkg/features/features.go
```diff
@@ -31,5 +32,5 @@ func init() {
 // defaultKubernetesFeatureGates consists of all known feature keys specific to external-provisioner.
 // To add a new feature, define a key for it above and add it here.
 var defaultKubernetesFeatureGates = map[utilfeature.Feature]utilfeature.FeatureSpec{
-	Topology: {Default: false, PreRelease: utilfeature.Alpha},
+	Topology: {Default: true, PreRelease: utilfeature.Beta},
```
msau42: Right now the sidecar errors if the CSINode object doesn't exist + the topology feature is enabled.

This may break backwards compatibility if you use a newer sidecar version on 1.13 where csi topology is disabled.

We could either:
* Say that this sidecar version is not supported on versions < 1.14
* Fallback to topology disabled behavior if CSINode doesn't exist

I'm leaning towards the latter, but that also means if we're on 1.14 and there's actually a problem with CSINode/driver registration, then we will continue to successfully provision a volume in a potentially bad zone.

wdyt? @jsafrane @verult @davidz627

msau42: actually now that I think about it more, is that what we want?  If the api server is temporarily having problems, then we end up provisioning in the wrong zone.

msau42: another option is that we leave topology off by default (and we offer no compatibility guarantees from alpha->beta, including k/k alpha->beta).

davidz627: The new sidecar would work on version `<1.14` with Topology enabled though right? So this is not a compatibility breaking thing. More of a "did you enable the feature on one component but not enable it on another" type problem. Maybe this is ok if we give a good error message making it very clear what is wrong in this case?

msau42: If the user on 1.13 has delayed binding enabled on the storageclass in a single zone cluster, then with the old provisioner, provisioning still succeeds.  If they upgrade the sidecar, then provisioning will fail with an error.

Now it's questionable what benefit they would get by enabling delayed binding in a single zone cluster, but there's currently no harm in doing so.

msau42: I'm going to make all the APIs call return an error if they fail (no fallback), but set topology feature gate to false.  Drivers will need to explicitly enable the topology feature gate after upgrading to 1.14.

msau42: For drivers with topology disabled, upgrading the sidecar on 1.13 should not break.

For drivers with topology enabled, upgrading the sidecar on 1.13 will break.  I think this is ok given it was an alpha feature.

msau42: Discussed with @jsafrane some more.  To simplify our documentation and testing matrix, we should just require 1.14 for this new version of the sidecar.  So then I can have topology enabled by default

pohly: Makes sense. We then need to maintain the older release branch for those people who use CSI with 1.13.

The Prow-based CI is ready for testing against two different Kubernetes releases. My proposal in https://github.com/kubernetes-csi/csi-driver-host-path/pull/16/commits is to maintain one or more deployments in the master branch, so users (and the CI) will always find what they look for there.

pkg/controller/topology_test.go
```diff
@@ -918,6 +918,7 @@ func TestTopologyAggregation(t *testing.T) {
 			nodeLabels:        []map[string]string{{}, {}, {}},
 			topologyKeys:      nil,
 			expectedRequisite: nil,
+			expectError:       true,
```
verult: I probably missed something, but shouldn't all "missing node info" cases expect an error now?

deploy/kubernetes/rbac.yaml
```diff
@@ -44,7 +44,7 @@ rules:
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotcontents"]
     verbs: ["get", "list"]
-  - apiGroups: ["csi.storage.k8s.io"]
+  - apiGroups: ["storage.k8s.io"]
     resources: ["csinodeinfos"]
```
davidz627: should be `csinode` now?

pkg/controller/topology.go
```diff
@@ -108,7 +107,6 @@ func GenerateAccessibilityRequirements(
 	} else {
 		// selectedNode is set so use topology from that node to populate preferredTerms
 		// TODO (verult) reuse selected node info from aggregateTopologies
-		// TODO (verult) retry
```
davidz627: was retry logic added? or I assume there might be a higher level retry

pkg/controller/topology.go
```diff
@@ -183,6 +177,7 @@ func aggregateTopologies(
 	if err != nil {
 		return nil, err
 	}
+	// TODO: use informers
```
davidz627: TODO(ISSUENUMBER): 

Please create an issue for tracking :)

pkg/controller/topology.go
msau42: retry is done by the lib-external-provisioner

pkg/controller/topology_test.go
```diff
@@ -1046,6 +1047,29 @@ func TestPreferredTopologies(t *testing.T) {
 				},
 			},
 		},
+		"allowedTopologies specified: no CSINode": {
```
verult: nit: change the other similar test cases to "no CSINode" too

pkg/controller/topology.go
msau42: #144 already created :-)

deploy/kubernetes/rbac.yaml
msau42: good catch fixed

pkg/controller/topology_test.go
msau42: If selected node == nil (immediate binding), and no nodes have a CSINode object, then we return no topology keys which falls back to the "no topology specified, provision wherever" case.

I think it is valid to return an error in the case where the returned topology keys are empty.  In order to get to this call, you need the driver to implement the topology capability and have the feature enabled (on by default).

pkg/controller/topology_test.go
msau42: I split it up into multiple cases, once for immediate binding and once for delayed binding.
* missing a Node object
* Node object doesn't have topology labels
* missing CSINode object
* CSINode object doesn't have topology keys

msau42:
I made substantial changes in order to fail appropriately while still preserving backwards compatibility with topology disabled behavior.  A summary of the changes:
* Immediate binding:
    * If a beta CSINode object that registers topology keys is found, at least one node has been upgraded, fail appropriately if needed
    * Otherwise, fallback to "topology disabled" behavior
* Delayed binding:
   * If selected node doesn't have beta CSINode object and the node is < 1.14, fallback to "topology disabled" behavior
   * Otherwise require that topology keys/labels are set.

Most of the test cases are under `TestTopologyAggregation`.

pkg/controller/topology.go
```diff
@@ -160,40 +183,78 @@ func aggregateTopologies(
...
 
+	// 2. Find all nodes with the topology keys and extract the topology values
 	selector, err := buildTopologyKeySelector(topologyKeys)
 	if err != nil {
 		return nil, err
 	}
+	// TODO (#144): use informers
 	nodes, err := kubeClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: selector})
 	if err != nil {
+		// If the topology feature in the external-provisioner is enabled,
```
davidz627: this comment seems to be repeated a lot. I wonder if there's a better way to capture this sentiment either in the code, error message, or whether it's even necessary in every spot. My one worry is that if we have this comment in each spot its "necessary" right now, what if we add new error conditions and forget to copy the comment, then future maintainers might be like "oh this comment is everywhere except here, maybe it doesn't apply here"...

This is not a big issue imo, but if we can think of a better solution now without too much effort that would be cool :)

pkg/controller/controller_test.go
```diff
@@ -1811,9 +1811,60 @@ func TestProvisionFromSnapshot(t *testing.T) {
...
+		},
+		"topology fail": {
+			driverSupportsTopology: true,
+			topologyKeys: []map[string][]string{
+				{driverName: []string{"com.example.csi/zone", "com.example.csi/rack"}},
+				{driverName: []string{"com.example.csi/zone", "com.example.csi/rack"}},
+			},
+			expectError: true,
+		},
+		"driver doesn't support topology": {},
```
davidz627: nit: kind of odd to just have an empty test case. maybe just explicitly set some of the important fields to false. `driverSupportsTopology` and `expectError` I guess.

pkg/controller/topology.go
```diff
@@ -134,19 +150,26 @@ func GenerateAccessibilityRequirements(
...
 		nodeInfos, err := kubeClient.StorageV1beta1().CSINodes().List(metav1.ListOptions{})
 		if err != nil {
-			// We must support provisioning if CSINode is missing, for backward compatibility.
-			klog.Warningf("error listing CSINodes: %v; proceeding to provision without topology information", err)
-			return nil, nil
+			// If the topology feature in the external-provisioner is enabled,
+			// CSINode beta feature on K8s apiserver must also be enabled.
+			// We don't want to fallback and provision in the wrong topology if there's some temporary
+			// error with the API server.
+			return nil, fmt.Errorf("error listing CSINodes: %v", err)
```
msau42: If we need to support running newer external provisioner on 1.13, then I think we need to disable topology by default.  Otherwise, provisioning on 1.13 will start to hit this error (if the driver also supports csi topology capability).

pkg/controller/topology.go
msau42: I think the "If the topology feature is enabled" part is sort of implied, and can be removed

pkg/controller/topology_test.go
```diff
@@ -908,42 +910,112 @@ func TestTopologyAggregation(t *testing.T) {
...
+			nodeLabels: []map[string]string{
+				{"com.example.csi/zone": "zone1"},
+				{"com.example.csi/zone": "zone2"},
+				{"com.example.csi/zone": "zone2"},
+			},
 			topologyKeys:      nil,
 			expectedRequisite: nil,
 		},
+		// Driver on node has not been updated to report topology keys
 		"random node: missing keys": {
```
verult: Should we have a test case where there's only a single CSINode with keys, in which there should be no errors?

msau42:
I've addressed comments and also wrapped the logic to lookup the CSINode object to reduce duplication of error handling.  I've also turned topology off by default again.

pkg/controller/topology.go
```diff
@@ -134,19 +157,51 @@ func GenerateAccessibilityRequirements(
 	return requirement, nil
 }
 
+// getSelectedNodeInfo returns the CSINode object for the given selectedNode.
+func getSelectedNodeInfo(
```
verult: nit: `getSelectedCSINode`?

pkg/controller/topology.go
```diff
@@ -160,40 +183,78 @@ func aggregateTopologies(
...
+		}
 
-	if len(topologyKeys) == 0 {
-		// Assuming the external provisioner is never running during node driver upgrades.
-		// If selectedNode != nil, the scheduler selected a node with no topology information.
-		// If selectedNode == nil, all nodes in the cluster are missing topology information.
-		// In either case, provisioning needs to be allowed to proceed.
-		return nil, nil
+		// Even though selectedNode is set, we still need to aggregate topology values across
+		// all nodes in order to find secondary topologies.
```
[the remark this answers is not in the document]
msau42: There's some desire to be able to limit the number of topologies returned, especially for the most plugins that only have a single primary topology: https://github.com/kubernetes-csi/external-provisioner/issues/221

So I wanted to add in some more explicit comments here explaining the reason why we need to return all possible topologies even though a plugin may only need one.

pkg/controller/topology.go
```diff
@@ -88,6 +91,13 @@ func GenerateAccessibilityRequirements(
 		requisiteTerms = flatten(allowedTopologies)
 	}
 
+	// It might be possible to reach here if:
```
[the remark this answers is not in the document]
msau42: I added a giant comment at the top. Let me know if this format works or you prefer something else

pkg/controller/topology_test.go
msau42: Do the existing "same keys and values across cluster" and "different values across cluster" cases address this?

msau42:
addressed latest comments

pkg/controller/topology.go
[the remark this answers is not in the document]
verult: It's super clear. Thanks!

pkg/controller/topology_test.go
verult: Not quite. I'm thinking about the case where k8s nodes are partially upgraded to 1.14, so some nodes have keys but others don't. Doesn't matter if Node objects have the right topology values.

verult: Something like:
```golang
nodeLabels: []map[string]string{
				{"com.example.csi/zone": "zone1"},
				{"com.example.csi/zone": "zone2"},
				{"com.example.csi/zone": "zone2"},
			},
			topologyKeys: []map[string][]string{
				{testDriverName: []string{"com.example.csi/zone"}},
			},
```
in which the requisite should only have "com.example.csi/zone": "zone1", right?

msau42:
/hold
debugging e2e test failures

pkg/features/features.go
```diff
@@ -31,5 +32,5 @@ func init() {
 // defaultKubernetesFeatureGates consists of all known feature keys specific to external-provisioner.
 // To add a new feature, define a key for it above and add it here.
 var defaultKubernetesFeatureGates = map[utilfeature.Feature]utilfeature.FeatureSpec{
-	Topology: {Default: false, PreRelease: utilfeature.Alpha},
+	Topology: {Default: false, PreRelease: utilfeature.Beta},
```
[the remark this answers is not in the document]
msau42: I'm unsure if we need actually need to.  It's important that kubernetes enables topology by default since it needs to be able to support all kinds of drivers, but since this flag is only controlling a specific driver implementation, I'm not sure if it actually gives much benefit being enabled by default.

pkg/controller/topology_test.go
msau42: Added 2 more test cases.

So actually, we in your case above, requisite has both zone1 and zone2. even though the other nodes don't have csiNode.  It's because csiNode is only used to determine which topology keys to use, but we just iterate over all nodes that have the topology key.

msau42:
E2E tests are passing (I had to explicitly enable topology in the driver because I it's now turned it off by default).  

I'll keep the hold to squash commits after review.

pkg/features/features.go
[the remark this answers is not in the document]
verult: ACK

msau42:
Squashed commits
/hold cancel

PR #239, opened 2019-03-04
Fix GetDriverName timeout

jsafrane:
It should be CSI call timeout, not deprecated connection timeout.

cc @pohly


jsafrane:
There is warning when the option is used and I'd like to keep it there.

https://github.com/kubernetes-csi/external-provisioner/blob/f588078fd1f33dd7e9557cceba4c05e3562683fd/cmd/csi-provisioner/csi-provisioner.go#L80-L82

PR #241, opened 2019-03-05
Cache driver capabilities

jsafrane:
We don't expect that driver capabilities change when the driver is running. In this PR:

* The provisioner checks capabilities on startup and exits, if the driver does not support dynamic provisioning.
* It caches all capabilities and check the cached values for snapshot and volume topology support when necessary.

edit: This changes behavior of external-provisioner.

Previously, it was running even if the driver did not support dynamic provisioning and created events for all PVCs that it won't provision them. Users could spot this event easily and talk to cluster admin.

Now users will see just "waiting for a volume to be created, either by external provisioner or manually". The provisioner just dies on startup and it's up to cluster admin to find out what's wrong.

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -25,8 +25,6 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/klog"
-
 	flag "github.com/spf13/pflag"
 
 	ctrl "github.com/kubernetes-csi/external-provisioner/pkg/controller"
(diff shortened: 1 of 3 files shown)


msau42:
If the provisioner just dies on startup, will kubernetes constantly try to restart it? Would that take up more cycles than just having it be idle?

jsafrane:
Yes, Kubernetes will try to restart it (with exponential backoff) and the pod should eventually end up in CrashLoopBackOff state, hoping to get admin's attention. The max backoff is [probably 5 minutes](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L131), i.e. nothing that I would worry about (for one driver).

Question is if it's the right thing to do. What is the right way how to get admin's attention? Crashing provisioner or event in PVC that this PVC cannot be provisioned?

msau42:
Conclusion:
- Keep current nice error message (don't exit the sidecar if the driver doesn't support?)
- for 1.15, we can add capabilities to CSIDriver object so that the PV controller can display the nice error message

jsafrane:
Restored sending of events `plugin CONTROLLER_SERVICE capability is not reported` instead of exiting.

PR #247, opened 2019-03-12
Use distroless as base image

jsafrane:


jsafrane:
Reason: k8s core components move its base images to distroless/static. See kubernetes/kubernetes#70249 (comment) for details.

davidz627:
have we done enough verification/testing to know that this works for us? i.e. core functionality works, can still get logs, etc.

jsafrane:
I tested node-driver-registrat + attacher + provisioner with distroless, `kubectl logs` works as usual. `kubectl exec` obviously does not work, because there is no `/bin/sh` in the container.

I haven't tested the other containers (snapshotter, resizer, cluster-registrar, ...)

PR #248, opened 2019-03-13
filter resizer related params from storageclass before passing them to the driver

mlmhl:
The external resizer introduces two StorageClass parameters used by resizing secret: https://github.com/kubernetes-csi/external-resizer/blob/25e836c8513f0d31df3604981f1312f0db376e45/pkg/resizer/csi_resizer.go#L45

The external provisioner should remove these parameters before passing to the driver.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -76,6 +76,9 @@ const (
 	prefixedNodePublishSecretNameKey      = csiParameterPrefix + "node-publish-secret-name"
 	prefixedNodePublishSecretNamespaceKey = csiParameterPrefix + "node-publish-secret-namespace"
 
+	prefixedResizerSecretNameKey      = csiParameterPrefix + "resizer-secret-name"
+	prefixedResizerSecretNamespaceKey = csiParameterPrefix + "resizer-secret-namespace"
+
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -571,6 +574,8 @@ func removePrefixedParameters(param map[string]string) (map[string]string, error
 			case prefixedNodeStageSecretNamespaceKey:
 			case prefixedNodePublishSecretNameKey:
 			case prefixedNodePublishSecretNamespaceKey:
+			case prefixedResizerSecretNameKey:
+			case prefixedResizerSecretNamespaceKey:
```
msau42: can you also update the unit test?

mlmhl: Updated, PTAL :)

.idea/external-provisioner.iml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
msau42: also remove these ide files?

mlmhl: Oh sorry, I forgot to check the commit change. Already removed, PTAL :)

PR #249, opened 2019-03-13
Update README to 1.14

jsafrane:
- Following external-attacher as example.
- Removed #Build and #Test, they were not useful, added short doc/development.md.
- Moved away from StatefulSet to Deployment. Deployment behaves better when a node gets disconnected from the rest of the cluster - new provisioner leader is elected in ~15 seconds, while it may take up to 5 minutes for StatefulSet to start a new replica. (Again, following external-attacher as sample)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,50 +1,95 @@
 [![Build Status](https://travis-ci.org/kubernetes-csi/external-provisioner.svg?branch=master)](https://travis-ci.org/kubernetes-csi/external-provisioner)
 
-# Kubernetes external provisioner that works with CSI volumes
+# CSI provisioner
 
-This is an example external provisioner for Kubernetes which provisions using CSI Volume drivers..  It's under heavy development, so at this time README.md is notes for the developers coding.  Once complete this will change to something user friendly.
+The external-provisioner is a sidecar container that dynamically provisions volumes by calling `ControllerCreateVolume` and `ControllerDeleteVolume` functions of CSI drivers. It is necessary because internal persistent volume controller running in Kubernetes controller-manager does not have any direct interfaces to CSI drivers.
 
-# Build
+## Overview
+The external-provisioner is an external controller that monitors `PersistentVolumeClaim` objects created by user and creates/deletes volumes for them. Full design can be found at Kubernetes proposal at [container-storage-interface.md](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/container-storage-interface.md)
 
-```bash
-make csi-provisioner
+## Compatibility
+
+This information reflects the head of this branch.
+
+| Compatible with CSI Version                                                                | Container Image                | Min K8s Version |
+| ------------------------------------------------------------------------------------------ | -------------------------------| --------------- |
+| [CSI Spec v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | quay.io/k8scsi/csi-provisioner | 1.14            |
+
+## Feature status
+
+Various external-provisioner releases come with different alpha / beta features. Check `--help` output for alpha/beta features in each release.
+
+Following table reflects the head of this branch.
+
+| Feature    | Status  | Default | Description                                                                                   |
+| ---------- | ------- | ------- | --------------------------------------------------------------------------------------------- |
+| Topology   | Beta    | Off     | [Topology aware dynamic provisioning](https://kubernetes-csi.github.io/docs/topology.html)  (requires kubelet 1.14 on nodes). |
+| Snapshots* | Alpha   | On      | [Snapshots and Restore](https://kubernetes-csi.github.io/docs/snapshot-restore-feature.html). |
+
+*) There is no special feature gate for snapshots, it is enabled by running [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter/) and creating its CRDs.
+
+All other external-provisioner features and the external-provisioner itself is considered GA and fully supported.
+
+## Usage
+
+It is necessary to create a new service account and give it enough privileges to run the external-provisioner, see `deploy/kubernetes/rbac.yaml`. The provisioner is then deployed as single Deployment as illustrated below:
+
+```sh
+kubectl create deploy/kubernetes/deployment.yaml
 ```
 
-# Test
+The external-provisioner may run in the same pod with other external CSI controllers such as the external-attacher, external-snapshotter and/or external-resizer.
 
-## Start Kubernetes
+Note that the external-provisioner does not scale with more replicas. Only one external-provisioner is elected as leader and running. The others are waiting for the leader to die. They re-elect a new active leader in ~15 seconds after death of the old leader.
 
-Run a local kubernetes cluster built from latest master code
+### Command line options
 
-## Run Storage Provider
+#### Recommended optional arguments"
+* `--csi-address <path to CSI socket>`: This is the path to the CSI driver socket inside the pod that the external-provisioner container will use to issue CSI operations (`/run/csi/socket` is used by default).
 
-### Use HostPath drivers
+* `--enable-leader-election`: Enables leader election. This is mandatory when there are multiple replicas of the same external-provisioner running for one CSI driver. Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.
 
-Go to drivers and run:
+* `--timeout <duration>`: Timeout of all calls to CSI driver. It should be set to value that accommodates majority of `ControllerCreateVolume` and `ControllerDeleteVolume` calls. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details. 15 seconds is used by default.
 
-```bash
-bin/hostpathplugin --drivername mydriver  --endpoint unix://tmp/csi.sock --nodeid foobar -v=5
-```
+* `--retry-interval-start <duration>` - Initial retry interval of failed provisioning or deletion. It doubles with each failure, up to `--retry-interval-max` and then it stops increasing. Default value is 1 second. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details.
 
-### Start external provisioner
+* `--retry-interval-max <duration>` - Maximum retry interval of failed provisioning or deletion. Default value is 5 minutes. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details.
 
-```bash
-bin/csi-provisioner --kubeconfig /var/run/kubernetes/admin.kubeconfig --alsologtostderr
-```
+* `--worker-threads <num>`: Number of simultaneously running `ControllerCreateVolume` and `ControllerDeleteVolume` operations. Default value is `100`.
 
-### Create Storage class, PVC, and Pod
+#### Other recognized arguments
+* `--feature-gates <gates>`: A set of comma separated `<feature-name>=<true|false>` pairs that describe feature gates for alpha/experimental features. See [list of features](#feature-status) or `--help` output for list of recognized features. Example: `--feature-gates Topology=true` to enable Topology feature that's disabled by default.
 
-```bash
-kubectl create -f examples/sc.yaml
-kubectl create -f example/pvc.yaml
-kubectl create -f example/pod.yaml
-```
+* `--kubeconfig <path>`: Path to Kubernetes client configuration that the external-provisioner uses to connect to Kubernetes API server. When omitted, default token provided by Kubernetes will be used. This option is useful only when the external-provisioner does not run as a Kubernetes pod, e.g. for debugging. Either this or `--master` needs to be set if the external-provisioner is being run out of cluster.
 
-### Delete PVC
+* `--master <url>`: Master URL to build a client config from. When omitted, default token provided by Kubernetes will be used. This option is useful only when the external-provisioner does not run as a Kubernetes pod, e.g. for debugging. Either this or `--kubeconfig` needs to be set if the external-provisioner is being run out of cluster.
 
-```bash
-kubectl delete pvc example-pvc
-```
+* `--volume-name-prefix <prefix>`: Prefix of PersistentVolume names created by the external-provisioner. Default value is "pvc", i.e. created PersistentVolume objects will have name `pvc-<uuid>`.
+
+* `--volume-name-uuid-length`: Length of UUID to be added to `--volume-name-prefix`. Default behavior is to NOT truncate the UUID.
+
+* `--version`: Prints current external-provisioner version and quits.
+
+* All glog / klog arguments are supported, such as `-v <log level>` or `-alsologtostderr`.
+
+#### Deprecated arguments
+* `--connection-timeout <duration>`: This option was used to limit establishing connection to CSI driver. Currently, the option does not have any effect and the external-provisioner tries to connect to CSI driver socket indefinitely. It is recommended to run ReadinessProbe on the driver to ensure that the driver comes up in reasonable time.
+
+
+### CSI error and timeout handling
+The external-provisioner invokes all gRPC calls to CSI driver with timeout provided by `--timeout` command line argument (15 seconds by default).
+
+Correct timeout value and number of worker threads depends on the storage backend and how quickly it is able to processes `ControllerCreateVolume` and `ControllerDeleteVolume` calls. The value should be set to accommodate majority of them. It is fine if some calls time out - such calls will be retried after exponential backoff (starting with 1s by default), however, this backoff will introduce delay when the call times out several times for a single volume.
+
+Frequency of `ControllerCreateVolume` and `ControllerDeleteVolume` retries can be configured by `--retry-interval-start` and `--retry-interval-max` parameters. The external-provisioner starts retries with `retry-interval-start` interval (1s by default) and doubles it with each failure until it reaches `retry-interval-max` (5 minutes by default). The external provisioner stops increasing the retry interval when it reaches `retry-interval-max`, however, it still retries provisioning/deletion of a volume until it's provisioned. The external-provisioner keeps its own number of provisioning/deletion failures for each volume.
+
+The external-provisioner can invoke up to `--worker-threads` (100 by default) `ControllerCreateVolume` **and** up to `--worker-threads` `ControllerDeleteVolume` calls in parallel, i.e. these two calls are counted separately. The external-provisioner assumes that the storage backend can cope with such high number of parallel requests and that the requests are handled in relatively short time (ideally sub-second). Lower value should be used for storage backends that expect slower processing related to newly created / deleted volumes or can handle lower amount of parallel calls.
+
+Details of error handling of individual CSI calls:
+* `ControllerCreateVolume`: The call might have timed out just before the driver provisioned a volume and was sending a response. From that reason, timeouts from `ControllerCreateVolume` is considered as "*volume may be provisioned*" or "*volume is being provisioned in the background*." The external-provisioner will retry calling `ControllerCreateVolume` after exponential backoff until it gets either successful response or final (non-timeout) error that the volume cannot be created.
+* `ControllerDeleteVolume`: This is similar to `ControllerCreateVolume`, The external-provisioner will retry calling `ControllerDeleteVolume` with exponential backoff after timeout until it gets either successful response or a final error that the volume cannot be deleted.
+* `Probe`: The external-provisioner retries calling Probe until the driver reports it's ready. It retries also when it receives timeout from `Probe` call. The external-provisioner has no limit of retries. It is expected that ReadinessProbe on the driver container will catch case when the driver takes too long time to get ready.
+* `GetPluginInfo`, `GetPluginCapabilitiesRequest`, `ControllerGetCapabilities`: The external-provisioner expects that these calls are quick and does not retry them on any error, including timeout. Instead, it assumes that the driver is faulty and exits. Note that Kubernetes will likely start a new provisioner container and it will start with `Probe` call.
 
 ## Community, discussion, contribution, and support
 
--- a/deploy/kubernetes/deployment.yaml
+++ b/deploy/kubernetes/deployment.yaml
@@ -17,13 +17,15 @@ spec:
       port: 12345
 
 ---
-kind: StatefulSet
-apiVersion: apps/v1beta1
+kind: Deployment
+apiVersion: apps/v1
 metadata:
   name: csi-provisioner
 spec:
-  serviceName: "csi-provisioner"
-  replicas: 1
+  replicas: 3
+  selector:
+    matchLabels:
+      app: csi-provisioner
   template:
     metadata:
       labels:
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -1,50 +1,86 @@
...
-### Use HostPath drivers
+* `--enable-leader-election`: Enables leader election. This is useful when there are multiple replicas of the same external-provisioner running for one CSI driver. Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.
 
-Go to drivers and run:
+* `--timeout <duration>`: Timeout of all calls to CSI driver. It should be set to value that accommodates majority of `ControllerCreateVolume` and `ControlerDeleteVolume` calls. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details. 15 seconds is used by default.
 
-```bash
-bin/hostpathplugin --drivername mydriver  --endpoint unix://tmp/csi.sock --nodeid foobar -v=5
-```
+* `--retry-interval-start <duration>` - Initial retry interval of failed provisioning or deletion. It doubles with each failure, up to `--retry-interval-max` and then it stops increasing. Default value is 1 second. [CSI error and timeout handling](#csi-error-and-timeout-handling) for details.
```
msau42: external attacher doesn't have these options? should it?

README.md
```diff
@@ -1,50 +1,86 @@
...
+* `--retry-interval-max <duration>` - Maximum retry interval of failed provisioning or deletion. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details.
 
-```bash
-bin/csi-provisioner --kubeconfig /var/run/kubernetes/admin.kubeconfig --alsologtostderr
-```
+* `--worker-threads <num>`: Number of simultaneously running `ControllerCreateVolume` and `ControllerDeleteVolume` operations. Default value is `100`.
 
-### Create Storage class, PVC, and Pod
+#### Other recognized arguments
+* `--provisioning-retry-count <num>`: Sets number of retries of failed provisioning of a single volume. The external-provisioner will give up provisioning after given number of retries. It should be used only with special CSI drivers whose provisioning of volumes is extremely slow or expensive so the external-provisioner does not waste resources. Set to `0` to retry forever (which is also the default value).
```
msau42: can you explain in more detail (at least to me) the use case of why you would give up rather than reconcile forever?

README.md
```diff
@@ -1,50 +1,86 @@
...
-This is an example external provisioner for Kubernetes which provisions using CSI Volume drivers..  It's under heavy development, so at this time README.md is notes for the developers coding.  Once complete this will change to something user friendly.
+The external-provisioner is a sidecar container that dynamically provisions volumes by calling `ControllerCreateVolume` and `ControlerDeleteVolume` functions of CSI drivers. It is necessary because internal persistent volume controller running in Kubernetes controller-manager does not have any direct interfaces to CSI drivers.
 
-# Build
+## Overview
+The external-provisioner is an external controller that monitors `PersistentVolumeClaim` objects created by user and annotatate by controller-manager and creates/deletes volumes for them. Full design can be found at Kubernetes proposal at [container-storage-interface.md](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/container-storage-interface.md)
 
-```bash
-make csi-provisioner
+## Compatibility
```
msau42: Can you add a feature status section talking about what's alpha/beta/ga, like [here](https://github.com/kubernetes-csi/external-provisioner/pull/238/files#diff-04c6e90faac2675aa89e2176d2eec7d8R11)?

README.md
```diff
@@ -1,50 +1,86 @@
...
 
-# Kubernetes external provisioner that works with CSI volumes
+# CSI provisioner
 
-This is an example external provisioner for Kubernetes which provisions using CSI Volume drivers..  It's under heavy development, so at this time README.md is notes for the developers coding.  Once complete this will change to something user friendly.
+The external-provisioner is a sidecar container that dynamically provisions volumes by calling `ControllerCreateVolume` and `ControlerDeleteVolume` functions of CSI drivers. It is necessary because internal persistent volume controller running in Kubernetes controller-manager does not have any direct interfaces to CSI drivers.
 
-# Build
+## Overview
+The external-provisioner is an external controller that monitors `PersistentVolumeClaim` objects created by user and annotatate by controller-manager and creates/deletes volumes for them. Full design can be found at Kubernetes proposal at [container-storage-interface.md](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/container-storage-interface.md)
```
msau42: do we need to mention the annotations? that's kind of an implementation detail

README.md
jsafrane: Attach/detach is done in clouds and it's "relatively" quick there. The options above are for the provisioners that take hours / days to create volume.

README.md
jsafrane: I tried to track where the parameter comes from and found only this commit:

https://github.com/kubernetes-incubator/external-storage/commit/5277d96aa7d72bf157d329021e9b5fe7511e7d20

I can't find PR where it was discussed. Apparently some storage finds it better to stop provisioning instead of trying forever. @wongma7, do you remember?

README.md
jsafrane: Sure. This is docs for 1.14, so topology is beta, right?

jsafrane: Table added (and Topoplogy moved to Beta)

README.md
jsafrane: Annotations removed.

deploy/kubernetes/deployment.yaml
```diff
@@ -34,8 +36,8 @@ spec:
         - name: csi-provisioner
           image: docker.io/k8scsi/csi-provisioner
```
msau42: How should we keep this image up to date whenever we branch/release?

jsafrane: Fixed (using canary on quay.io) and fixed storage class in examples too (it used csi-flex driver from some reason).

msau42: We'll have to remember to update this when we branch

README.md
```diff
@@ -15,6 +15,16 @@ This information reflects the head of this branch.
...
 
+## Feature status
+
+Various external-provisioner releases come with different alpha / beta features. Alpha features are disabled by default and must be explicitly enabled on command line. Beta features are enabled by default and can be disabled on command line. Check `--help` output for alpha/beta features in each release.
+
+Following table reflects the head of this branch.
+
+| Feature  | Status | Description                                                                                |
+| -------- | ------ | ------------------------------------------------------------------------------------------ |
+| Topology | Beta   | [Topology aware dynamic provisioning](https://kubernetes-csi.github.io/docs/topology.html) |
```
msau42: And restore from snapshot? It's not feature gated, should it be?

README.md
```diff
@@ -15,6 +15,16 @@ This information reflects the head of this branch.
 | ------------------------------------------------------------------------------------------ | ---------------------------------------| --------------- |
 | [CSI Spec v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | quay.io/k8scsi/csi-provisioner:v1.0.1  | 1.13            |
 
+## Feature status
+
```
msau42: Also should we say the provisioner itself is GA?

README.md
wongma7: I am sure somebody requested/added it and I didn't object, I can't find the PR either. csi provisioner doesn't necessarily have to expose it

README.md
```diff
@@ -1,50 +1,94 @@
...
+## Feature status
 
-## Run Storage Provider
+Various external-provisioner releases come with different alpha / beta features. Alpha features are disabled by default and must be explicitly enabled on command line. Beta features are enabled by default and can be disabled on command line. Check `--help` output for alpha/beta features in each release.
 
-### Use HostPath drivers
+Following table reflects the head of this branch.
 
-Go to drivers and run:
+| Feature  | Status | Description                                                                                |
```
msau42: Do we need a min K8s column?

msau42: And a default column

README.md
```diff
@@ -1,50 +1,94 @@
...
+
+#### Other recognized arguments
+* `--provisioning-retry-count <num>`: Sets number of retries of failed provisioning of a single volume. The external-provisioner will give up provisioning after given number of retries. It should be used only with special CSI drivers whose provisioning of volumes is extremely slow (e.g. hours or days) or expensive so the external-provisioner does not waste resources. Set to `0` to retry forever (which is also the default value).
+
+* `--deletion-retry-count <num>`: Sets number of retries of failed deletion of a single volume. The external-provisioner will give up deletion of the volume after given number of retries. It should be used only with special CSI drivers whose deletion of volumes is extremely slow (e.g. hours or days) or expensive so the external-provisioner does not waste resources. Set to `0` to retry forever (which is also the default value).
+
+* `--feature-gates <gates>`: Enables or disables alpha (disabled by default) or beta features (enabled by default). See [list of features](#feature-status) or `--help` output for list of recognized features.
```
msau42: i'm thinking topology will not be enabled by default even though it's beta

msau42: Also, example syntax?

README.md
jsafrane: @xing-yang, do we want to have feature gate for snapshot restore?

IMO, someone installed shapshotter and its CRDs, it could be enough to automatically enable feature in the provisioner.

README.md
jsafrane: Added both, but it's tricky. We introduced CSINodeInfo in 1.13, but 1.13 is compatible only with sidecar 1.0.x, not with current head. For topology as it is implemented in 1.14, 1.14 is the minimal version.

README.md
jsafrane: Fixed

README.md
jsafrane: OK, these parameters were not released, so I am removing them from the doc and I'll send a separate PR to remove them from the code.

README.md
jsafrane: Added a note below the table

README.md
jsafrane: https://github.com/kubernetes-csi/external-provisioner/pull/254

README.md
msau42: We should at least still say it's alpha

README.md
```diff
@@ -1,50 +1,92 @@
 [![Build Status](https://travis-ci.org/kubernetes-csi/external-provisioner.svg?branch=master)](https://travis-ci.org/kubernetes-csi/external-provisioner)
 
-# Kubernetes external provisioner that works with CSI volumes
+# CSI provisioner
 
-This is an example external provisioner for Kubernetes which provisions using CSI Volume drivers..  It's under heavy development, so at this time README.md is notes for the developers coding.  Once complete this will change to something user friendly.
+The external-provisioner is a sidecar container that dynamically provisions volumes by calling `ControllerCreateVolume` and `ControlerDeleteVolume` functions of CSI drivers. It is necessary because internal persistent volume controller running in Kubernetes controller-manager does not have any direct interfaces to CSI drivers.
```
pohly: s/ControlerDeleteVolume/ControllerDeleteVolume/

Is it intentionally that each paragraph is a single line? That'll make future diffs larger than they have to be.

README.md
```diff
@@ -1,50 +1,92 @@
...
+
+The external-provisioner can invoke up to `--worker-threads` (100 by default) `ControllerCreateVolume` **and** up to `--worker-threads` `ControllerDeleteVolume` calls in parallel, i.e. these two calls are counted separately. The external-provisioner assumes that the storage backend can cope with such high number of parallel requests and that the requests are handled in relatively short time (ideally sub-second). Lower value should be used for storage backends that expect slower processing related to newly created / deleted volumes or can handle lower amount of parallel calls.
+
+Details of error handling of individual CSI calls:
+* `ControllerCreateVolume`: The call might have timed out just before the driver provisioned a volume and was sending a response. From that reason, timeouts from `ControllerCreateVolume` is considered as "*volume may be provisioned*" or "*volume is being provisioned in the background*." The external-provisioner will re-try calling `ControllerCreateVolume` after exponential backoff until it gets either successful response or final (non-timeout) error that the volume cannot be created.
```
pohly: s/re-try/retry/

README.md
```diff
@@ -1,50 +1,92 @@
...
+* `ControllerDeleteVolume`: This is similar to `ControllerCreateVolume`, The external-provisioner will re-try calling `ControllerDeleteVolume` with exponential backoff after timeout until it gets either successful response or a final error that the volume cannot be deleted.
+* `Probe`: The external-provisioner re-tries calling Probe until the driver reports it's ready. It re-tries also when it receives timeout from `Probe` call. The external-provisioner has no limit of retries. It is expected that ReadinessProbe on the driver container will catch case when the driver takes too long time to get ready.
```
pohly: s/re-tries/retries/

README.md
```diff
@@ -1,50 +1,92 @@
...
-make csi-provisioner
-```
+## Compatibility
 
-# Test
+This information reflects the head of this branch.
 
-## Start Kubernetes
+| Compatible with CSI Version                                                                | Container Image                        | Min K8s Version |
+| ------------------------------------------------------------------------------------------ | ---------------------------------------| --------------- |
```
msau42: Change this to 1.14 (due to topology being on by default)

README.md
```diff
@@ -1,50 +1,92 @@
...
-## Run Storage Provider
+Various external-provisioner releases come with different alpha / beta features. Alpha features are disabled by default and must be explicitly enabled on command line. Beta features are enabled by default and can be disabled on command line. Check `--help` output for alpha/beta features in each release.
 
-### Use HostPath drivers
+Following table reflects the head of this branch.
 
-Go to drivers and run:
+| Feature  | Status | Default | Min. K8s version | Description                                                                                |
+| -------- | ------ | ------- | ---------------- | ------------------------------------------------------------------------------------------ |
+| Topology | Beta   | Off     | 1.14             | [Topology aware dynamic provisioning](https://kubernetes-csi.github.io/docs/topology.html) |
```
msau42: Ok after our discussion, we're going to have topology on by default and require 1.14 regardless of if you use the feature or not. So we don't need the two columns anymore, sorry :(

msau42: Or maybe we should still keep the default column because alpha features would be off

README.md
```diff
@@ -1,50 +1,92 @@
...
+
+* `--kubeconfig <path>`: Path to Kubernetes client configuration that the external-provisioner uses to connect to Kubernetes API server. When omitted, default token provided by Kubernetes will be used. This option is useful only when the external-provisioner does not run as a Kubernetes pod, e.g. for debugging. Either this or `--master` needs to be set if the external-provisioner is being run out of cluster.
+
+* `--master <url>`: Master URL to build a client config from. When omitted, default token provided by Kubernetes will be used. This option is useful only when the external-provisioner does not run as a Kubernetes pod, e.g. for debugging. Either this or `--kubeconfig` needs to be set if the external-provisioner is being run out of cluster.
+
+* `--volume-name-prefix <prefix>`: Prefix of PersistentVolume names created by the external-provisioner. Default value is "pvc", i.e. created PersistentVolume objects will have name `pvc-<uuid>`.
+
+* `--volume-name-uuid-length`: Length of UUID to be added to `--volume-name-prefix`. Defaults behavior is to NOT truncate the UUID.
```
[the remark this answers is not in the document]
jsafrane: Fixed

README.md
jsafrane: I am not worried about diff size, as long as gihub can highlight the differences.

README.md
jsafrane: fixed

README.md
jsafrane: fixed

README.md
jsafrane: fixed (but the image version is now wrong)

README.md
jsafrane: Removed `Min. K8s version`

README.md
msau42: Let's pre-call it 1.1.0?

msau42: actually how about just canary?

I think official cut versions we can keep in csi-docs

jsafrane: Hmm, I'd like to have **single** place where we have image versions for all supported Kubernetes releases. If it's in csi-docs, then we should link it from here. It will simplify release process a bit.

jsafrane: Removed version from the image.

README.md
jsafrane: Added snapshots as alpha feature, with a note that it cannot be turned off.

README.md
```diff
@@ -1,50 +1,95 @@
...
-## Start Kubernetes
+| Compatible with CSI Version                                                                | Container Image                | Min K8s Version |
+| ------------------------------------------------------------------------------------------ | -------------------------------| --------------- |
+| [CSI Spec v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | quay.io/k8scsi/csi-provisioner | 1.14            |
 
-Run a local kubernetes cluster built from latest master code
+## Feature status
 
-## Run Storage Provider
+Various external-provisioner releases come with different alpha / beta features. Alpha features are disabled by default and must be explicitly enabled on command line. Beta features are enabled by default and can be disabled on command line. Check `--help` output for alpha/beta features in each release.
```
msau42: "alpha features are disabled by default" doesn't make sense with the snapshots table entry.

README.md
```diff
@@ -1,50 +1,95 @@
...
 
-### Use HostPath drivers
+Following table reflects the head of this branch.
 
-Go to drivers and run:
+| Feature    | Status  | Default | Description                                                                                  |
+| ---------- | ------- | ------- | -------------------------------------------------------------------------------------------- |
+| Topology   | Beta    | On      | [Topology aware dynamic provisioning](https://kubernetes-csi.github.io/docs/topology.html)   |
```
msau42: Is it worth mentioning that nodes also have to be on 1.14 for this feature to work?

README.md
```diff
@@ -1,50 +1,95 @@
...
+The external-provisioner may run in the same pod with other external CSI controllers such as the external-attacher, external-snapshotter and/or external-resizer.
 
-```bash
-kubectl delete pvc example-pvc
-```
+Note that the external-provisioner does not scale with more replicas. Only one external-provisioner is elected as leader and running. The others are waiting for the leader to die. They re-elect a new active leader in ~15 seconds after death of the old leader.
+
+### Command line options
+
+#### Important optional arguments that are highly recommended to be used
```
msau42: Simplify to "Recommended optional arguments"

README.md
jsafrane: Note added

README.md
jsafrane: Done

README.md
jsafrane: Removed whole sentences about alpha/beta features.

README.md
```diff
@@ -1,50 +1,95 @@
...
-Go to drivers and run:
+| Feature    | Status  | Default | Description                                                                                   |
+| ---------- | ------- | ------- | --------------------------------------------------------------------------------------------- |
+| Topology   | Beta    | On      | [Topology aware dynamic provisioning](https://kubernetes-csi.github.io/docs/topology.html)  (requires kubelet 1.14 on nodes). |
+| Snapshots* | Alpha   | On      | [Snapshots and Restore](https://kubernetes-csi.github.io/docs/snapshot-restore-feature.html). |
 
-```bash
-bin/hostpathplugin --drivername mydriver  --endpoint unix://tmp/csi.sock --nodeid foobar -v=5
-```
+*) There is no special feature gate for snapshots, it is enabled by running [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter/) and creating its CRDs.
```
msau42: thinking more about this, maybe we should still have a feature gate?  What if someone tries to use snapshots with a beta/GA driver and it causes external-provisioner to crash?

jsafrane: Such change would break existing deployments. There was no alpha feature gate, yet snapshots were enabled before.

BTW, the same is going to happen to migration. It's enabled in kubelet / controller-manager, external-provisioner is just going to provision volumes for in-tree volumes without any gates

jsafrane: @davidz627 @ddebroy, FYI ^

pohly:
Where do we document that `fsType` can be set as parameter in a storage class for a CSI driver? The code is here: https://github.com/kubernetes-csi/external-provisioner/blob/7f7d8281afe84cfaa488fb577a23b3cd81f0cd62/pkg/controller/controller.go#L359-L369

README.md
msau42: Given that snapshots is still alpha I think it is fine to break existing deployments (with a release note).  From a support perspective, we don't want to compromise our beta/GA quality with alpha features.

For migration, I think it is fine to have the feature gate controlled by k8s.  Compared to snapshots, there is nothing we can do to prevent users from using the alpha crds

davidz627: Right, I think for migration if the gates are all on in k8s and k8s is requesting that the external provisioner respond to the migrated volumes - if it has the capability it should just do it. It doesn't seem strictly necessary to me to gate the sidecars seperately

msau42: Thinking about the snapshots case a little more.  Is the snapshots logic in the external-provisioner only triggered by the presence of the PVC.DataSource field?  If so, then that's the same situation as migration, so that's fine.

If you could invoke snapshots logic just from creating the snapshot CRs, then we should feature gate it.

jsafrane:
@pohly, StorageClass parameters should IMO go to https://kubernetes-csi.github.io/docs/external-provisioner.html#storageclass-parameters

jsafrane:
... and `fsType` is already there.

README.md
```diff
@@ -1,50 +1,95 @@
...
-## Run Storage Provider
+Various external-provisioner releases come with different alpha / beta features. Check `--help` output for alpha/beta features in each release.
 
-### Use HostPath drivers
+Following table reflects the head of this branch.
 
-Go to drivers and run:
+| Feature    | Status  | Default | Description                                                                                   |
+| ---------- | ------- | ------- | --------------------------------------------------------------------------------------------- |
+| Topology   | Beta    | On      | [Topology aware dynamic provisioning](https://kubernetes-csi.github.io/docs/topology.html)  (requires kubelet 1.14 on nodes). |
```
msau42: Sorry for flipping back on forth. My latest PR has topology off by default so that the sidecar is backwards compatible with 1.13.

jsafrane: Topology is Off by default now.

jsafrane:
I squashed all the commits, hoping this is the final round of review :-)

README.md
```diff
@@ -1,50 +1,95 @@
...
+
+* `--retry-interval-start <duration>` - Initial retry interval of failed provisioning or deletion. It doubles with each failure, up to `--retry-interval-max` and then it stops increasing. Default value is 1 second. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details.
+
+* `--retry-interval-max <duration>` - Maximum retry interval of failed provisioning or deletion. Default value is 5 minutes. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details.
+
+* `--worker-threads <num>`: Number of simultaneously running `ControllerCreateVolume` and `ControllerDeleteVolume` operations. Default value is `100`.
+
+#### Other recognized arguments
+* `--feature-gates <gates>`: A set of comma separated `<feature-name>=<true|false>` pairs that describe feature gates for alpha/experimental features. See [list of features](#feature-status) or `--help` output for list of recognized features. Example: `--feature-gates Topology=false` to disable Topology feature that's enabled by default.
```
msau42: This has to be updated too

jsafrane: Fixed

PR #251, opened 2019-03-14
Do not stop saving PVs to API server

jsafrane:
CreateProvisionedPVLimiter will cause the controller to never stop saving PVs to API server until it succeeds. Previously, the controller tried couple of times and then deleted the storage volume (i.e. called
ControllerDeleteVolume).

Fixes: #94


msau42:
If you tried to create the PV object, but it already exists, do we know if it is actually for that exact volume, or could it be some other completely different volume?  Ie, could we potentially leak underlying volumes?

jsafrane:
PV name is based on PVC.UID -> another PV will be created when user deletes old PVC and creates a new one. And user can't change PVC.Spec (except for capacity), so the existing PV must be the right one.

msau42:
What if it was another external-provisioner that created the PV?

jsafrane:
If another provisioner created the volume, we must expect that the driver was idempotent and both provisioners got the same volume. If not, yes, one of the volumes is leaked.

PR #253, opened 2019-03-15
Add dynamic provisioning support for CSI migration scenarios

ddebroy:
NOTE 1: The CSI migration provisioning logic here is not directly feature gated. However, the feature flags on the in-tree side sufficiently protects the logic here. This is because the PV controller in-tree will only yield PVCs (for in-tree provisioners) to the external provisioner when the feature flags for CSI migration for a particular plugin is enabled in-tree which in turn will make a plugin advertise `IsMigratedToCSI `: https://github.com/kubernetes/kubernetes/commit/23478f104f6ea26f752f1aae71301b637d2935a5#diff-3f32ad90966ceeb297f57c2a8d726348R1404 

NOTE 2: Unit tests are skipped for now. We need to address https://github.com/kubernetes/kubernetes/issues/74594 to be able to inject unit-tests as desired into the translation framework for unit testing purposes.

Tested e2e with GCE PD CSI driver name and ensured CSI driver got invoked for provisioning but the resulting PV pointed to GCEPersistentDisk

```
I0315 03:43:50.483655       1 controller.go:476] CreateVolumeRequest {Name:pvc-8bd370e9-46d4-11e9-b867-42010a8a001f CapacityRange:required_bytes:6442450944  VolumeCapabilities:[mount:<fs_type:"ext4" > access_mode:<mode:SINGLE_NODE_WRITER > ] Parameters:map[replication-type:none type:pd-standard] Secrets:map[] VolumeContentSource:<nil> AccessibilityRequirements:<nil> XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}
I0315 03:43:50.483838       1 connection.go:180] GRPC call: /csi.v1.Controller/CreateVolume
I0315 03:43:50.483866       1 connection.go:181] GRPC request: {"capacity_range":{"required_bytes":6442450944},"name":"pvc-8bd370e9-46d4-11e9-b867-42010a8a001f","parameters":{"replication-type":"none","type":"pd-standard"},"volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":1}}]}
I0315 03:43:54.881857       1 connection.go:183] GRPC response: {"volume":{"accessible_topology":[{"segments":{"topology.gke.io/zone":"us-west1-a"}}],"capacity_bytes":6442450944,"volume_id":"projects/docker4x/zones/us-west1-a/disks/pvc-8bd370e9-46d4-11e9-b867-42010a8a001f"}}
I0315 03:43:54.883303       1 connection.go:184] GRPC error: <nil>
I0315 03:43:54.883320       1 controller.go:520] create volume rep: {CapacityBytes:6442450944 VolumeId:projects/docker4x/zones/us-west1-a/disks/pvc-8bd370e9-46d4-11e9-b867-42010a8a001f VolumeContext:map[] ContentSource:<nil> AccessibleTopology:[segments:<key:"topology.gke.io/zone" value:"us-west1-a" > ] XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}
I0315 03:43:54.883435       1 controller.go:588] successfully created PV {GCEPersistentDisk:&GCEPersistentDiskVolumeSource{PDName:pvc-8bd370e9-46d4-11e9-b867-42010a8a001f,FSType:ext4,Partition:0,ReadOnly:false,} AWSElasticBlockStore:nil HostPath:nil Glusterfs:nil NFS:nil RBD:nil ISCSI:nil Cinder:nil CephFS:nil FC:nil Flocker:nil FlexVolume:nil AzureFile:nil VsphereVolume:nil Quobyte:nil AzureDisk:nil PhotonPersistentDisk:nil PortworxVolume:nil ScaleIO:nil Local:nil StorageOS:nil CSI:nil}
I0315 03:43:54.883573       1 controller.go:1275] provision "default/podpvc" class "slow": volume "pvc-8bd370e9-46d4-11e9-b867-42010a8a001f" provisioned
I0315 03:43:54.883594       1 controller.go:1292] provision "default/podpvc" class "slow": succeeded
```

```
> kubectl describe pv pvc-8bd370e9-46d4-11e9-b867-42010a8a001f
Name:              pvc-8bd370e9-46d4-11e9-b867-42010a8a001f
Labels:            failure-domain.beta.kubernetes.io/region=us-west1
                   failure-domain.beta.kubernetes.io/zone=us-west1-a
Annotations:       pv.kubernetes.io/provisioned-by: pd.csi.storage.gke.io
Finalizers:        [external-provisioner.volume.kubernetes.io/finalizer kubernetes.io/pv-protection]
StorageClass:      slow
Status:            Bound
Claim:             default/podpvc
Reclaim Policy:    Delete
Access Modes:      RWO
VolumeMode:        Filesystem
Capacity:          6Gi
Node Affinity:     
  Required Terms:  
    Term 0:        failure-domain.beta.kubernetes.io/zone in [us-west1-a]
                   failure-domain.beta.kubernetes.io/region in [us-west1]
Message:           
Source:
    Type:       GCEPersistentDisk (a Persistent Disk resource in Google Compute Engine)
    PDName:     pvc-8bd370e9-46d4-11e9-b867-42010a8a001f
    FSType:     ext4
    Partition:  0
    ReadOnly:   false
Events:         <none>
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -41,6 +41,7 @@ import (
 
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	utilflag "k8s.io/apiserver/pkg/util/flag"
+	csitranslationlib "k8s.io/csi-translation-lib"
 )
 
 var (
@@ -159,19 +160,33 @@ func init() {
 	timeStamp := time.Now().UnixNano() / int64(time.Millisecond)
 	identity := strconv.FormatInt(timeStamp, 10) + "-" + strconv.Itoa(rand.Intn(10000)) + "-" + provisionerName
 
-	// Create the provisioner: it implements the Provisioner interface expected by
-	// the controller
-	csiProvisioner := ctrl.NewCSIProvisioner(clientset, csiAPIClient, *operationTimeout, identity, *volumeNamePrefix, *volumeNameUUIDLength, grpcClient, snapClient, provisionerName, pluginCapabilities, controllerCapabilities)
-	provisionController = controller.NewProvisionController(
-		clientset,
-		provisionerName,
-		csiProvisioner,
-		serverVersion.GitVersion,
+	provisionerOptions := []func(*controller.ProvisionController) error{
 		controller.LeaderElection(*enableLeaderElection),
 		controller.FailedProvisionThreshold(0),
 		controller.FailedDeleteThreshold(0),
 		controller.RateLimiter(workqueue.NewItemExponentialFailureRateLimiter(*retryIntervalStart, *retryIntervalMax)),
 		controller.Threadiness(int(*workerThreads)),
+	}
+
+	supportsMigrationFromInTreePluginName := ""
+	if csitranslationlib.IsMigratedCSIDriverByName(provisionerName) {
+		supportsMigrationFromInTreePluginName, err = csitranslationlib.GetInTreeNameFromCSIName(provisionerName)
+		if err != nil {
+			klog.Fatalf("Failed to get InTree plugin name for migrated CSI plugin %s: %v", provisionerName, err)
+		}
+		klog.V(2).Infof("Supports migration from in-tree plugin: %s", supportsMigrationFromInTreePluginName)
+		provisionerOptions = append(provisionerOptions, controller.AdditionalProvisionerNames([]string{supportsMigrationFromInTreePluginName}))
+	}
+
+	// Create the provisioner: it implements the Provisioner interface expected by
+	// the controller
+	csiProvisioner := ctrl.NewCSIProvisioner(clientset, csiAPIClient, *operationTimeout, identity, *volumeNamePrefix, *volumeNameUUIDLength, grpcClient, snapClient, provisionerName, pluginCapabilities, controllerCapabilities, supportsMigrationFromInTreePluginName)
+	provisionController = controller.NewProvisionController(
+		clientset,
+		provisionerName,
+		csiProvisioner,
+		serverVersion.GitVersion,
+		provisionerOptions...,
 	)
 }
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -42,6 +42,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	csiclientset "k8s.io/csi-api/pkg/client/clientset/versioned"
+	csitranslationlib "k8s.io/csi-translation-lib"
 	"k8s.io/klog"
 
 	"google.golang.org/grpc"
@@ -143,19 +144,20 @@ var (
 
 // CSIProvisioner struct
 type csiProvisioner struct {
-	client                 kubernetes.Interface
-	csiClient              csi.ControllerClient
-	csiAPIClient           csiclientset.Interface
-	grpcClient             *grpc.ClientConn
-	snapshotClient         snapclientset.Interface
-	timeout                time.Duration
-	identity               string
-	volumeNamePrefix       string
-	volumeNameUUIDLength   int
-	config                 *rest.Config
-	driverName             string
-	pluginCapabilities     connection.PluginCapabilitySet
-	controllerCapabilities connection.ControllerCapabilitySet
+	client                                kubernetes.Interface
+	csiClient                             csi.ControllerClient
+	csiAPIClient                          csiclientset.Interface
+	grpcClient                            *grpc.ClientConn
+	snapshotClient                        snapclientset.Interface
+	timeout                               time.Duration
+	identity                              string
+	volumeNamePrefix                      string
+	volumeNameUUIDLength                  int
+	config                                *rest.Config
+	driverName                            string
+	pluginCapabilities                    connection.PluginCapabilitySet
+	controllerCapabilities                connection.ControllerCapabilitySet
+	supportsMigrationFromInTreePluginName string
 }
 
 var _ controller.Provisioner = &csiProvisioner{}
@@ -211,22 +213,24 @@ func NewCSIProvisioner(client kubernetes.Interface,
 	snapshotClient snapclientset.Interface,
 	driverName string,
 	pluginCapabilities connection.PluginCapabilitySet,
-	controllerCapabilities connection.ControllerCapabilitySet) controller.Provisioner {
+	controllerCapabilities connection.ControllerCapabilitySet,
+	supportsMigrationFromInTreePluginName string) controller.Provisioner {
 
 	csiClient := csi.NewControllerClient(grpcClient)
 	provisioner := &csiProvisioner{
-		client:                 client,
-		grpcClient:             grpcClient,
-		csiClient:              csiClient,
-		csiAPIClient:           csiAPIClient,
-		snapshotClient:         snapshotClient,
-		timeout:                connectionTimeout,
-		identity:               identity,
-		volumeNamePrefix:       volumeNamePrefix,
-		volumeNameUUIDLength:   volumeNameUUIDLength,
-		driverName:             driverName,
-		pluginCapabilities:     pluginCapabilities,
-		controllerCapabilities: controllerCapabilities,
+		client:                                client,
+		grpcClient:                            grpcClient,
+		csiClient:                             csiClient,
+		csiAPIClient:                          csiAPIClient,
+		snapshotClient:                        snapshotClient,
+		timeout:                               connectionTimeout,
+		identity:                              identity,
+		volumeNamePrefix:                      volumeNamePrefix,
+		volumeNameUUIDLength:                  volumeNameUUIDLength,
+		driverName:                            driverName,
+		pluginCapabilities:                    pluginCapabilities,
+		controllerCapabilities:                controllerCapabilities,
+		supportsMigrationFromInTreePluginName: supportsMigrationFromInTreePluginName,
 	}
 	return provisioner
 }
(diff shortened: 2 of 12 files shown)


cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -80,6 +82,9 @@ func init() {
 	if *connectionTimeout != 0 {
 		klog.Warningf("Warning: option -connection-timeout is deprecated and has no effect")
 	}
+	if *provisioner != "" {
```
ddebroy: A bit confused why this section with `provisioner` (as well as the declaration above) is showing up as new code given it already exists in master and nothing has changed about it.

pkg/controller/controller.go
```diff
@@ -362,6 +369,24 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
 
+	createVolumeRequestParameters := options.Parameters
+	performInTreeTranslation := false
+	if p.handlesMigrationFromInTreePlugin {
+		storageClassName := options.PVC.Spec.StorageClassName
+		storageClass, err := p.client.StorageV1().StorageClasses().Get(*storageClassName, metav1.GetOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to get storage class named %s: %v", *storageClassName, err)
+		}
+		if storageClass.Provisioner == p.handlesMigrationFromInTreePluginName {
```
davidz627: add the else block that can log that we are not translating the storage class parameters. Just to make this easier to understand (I got confused at first)

pkg/controller/controller.go
```diff
@@ -362,6 +369,24 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
...
+			klog.V(2).Infof("Perform CSI migration for intree plugin %s", storageClass.Provisioner)
```
davidz627: be more specific here. `CSI Migration: translating storage class parameters for in-tree plugin %s`?

pkg/controller/controller.go
```diff
@@ -551,6 +576,13 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 		pv.Spec.PersistentVolumeSource.CSI.FSType = fsType
 	}
 
+	if performInTreeTranslation {
```
davidz627: `performInTreeTranslation` doesn't seem like the right name.

It relates to "this specific volume is using translation logic for migration"
`volumeMigrated`?
`volumeShimmed`?
`volumeTranslated`?

I would go with translated probably but will defer to you

pkg/controller/controller.go
```diff
@@ -238,20 +241,24 @@ func NewCSIProvisioner(client kubernetes.Interface,
...
+		grpcClient:                           grpcClient,
+		csiClient:                            csiClient,
+		csiAPIClient:                         csiAPIClient,
+		snapshotClient:                       snapshotClient,
+		timeout:                              connectionTimeout,
+		identity:                             identity,
+		volumeNamePrefix:                     volumeNamePrefix,
+		volumeNameUUIDLength:                 volumeNameUUIDLength,
+		handlesMigrationFromInTreePlugin:     handlesMigrationFromInTreePlugin,
+		handlesMigrationFromInTreePluginName: handlesMigrationFromInTreePluginName,
```
davidz627: maybe just `supportsInTreePlugin` (name is implied)?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -155,19 +156,35 @@ func init() {
...
+		controller.FailedProvisionThreshold(0),
+		controller.FailedDeleteThreshold(0),
+		controller.RateLimiter(workqueue.NewItemExponentialFailureRateLimiter(*retryIntervalStart, *retryIntervalMax)),
+		controller.Threadiness(int(*workerThreads)),
+	}
+
+	handlesMigrationFromInTreePlugin := false
+	handlesMigrationFromInTreePluginName := ""
+	if csitranslationlib.IsMigratedCSIDriverByName(provisionerName) {
+		handlesMigrationFromInTreePluginName, err = csitranslationlib.GetInTreeNameFromCSIName(provisionerName)
```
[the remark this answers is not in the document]
ddebroy: The suggestion to have options and keep this extensible sounds good. The main concern I have with the options is that some of the logic in https://github.com/kubernetes/csi-translation-lib/blob/master/translate.go is pretty tightly bound to names of specific CSI drivers for certain translations like https://github.com/kubernetes/csi-translation-lib/blob/master/translate.go#L72. Perhaps we can add a note for authors in the migration notes that name changes for a CSI driver needs to be reflected in the csi-translation-lib.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -59,12 +59,32 @@ var (
 	workerThreads        = flag.Uint("worker-threads", 100, "Number of provisioner worker threads, in other words nr. of simultaneous CSI calls.")
 	operationTimeout     = flag.Duration("timeout", 10*time.Second, "Timeout for waiting for creation or deletion of a volume")
 	provisioner          = flag.String("provisioner", "", "This option is deprecated")
+	inTreePluginName     = flag.String("supersedes-in-tree-plugin-name", "", "An in-tree plugin that is superseded by this CSI plugin")
```
jsafrane: I would appreciate "migration" somewhere in the parameter name and description. `--migrated-plugin-name`?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -59,12 +59,32 @@ var (
...
 
+// check that the CSI plugin handles migration for the inTreePluginName according to:
+// https://github.com/kubernetes/csi-translation-lib/blob/master/translate.go#L28
+// A mismatch will lead to failed migration API calls like TranslateCSIPVToInTree. So warn of any mismatch but don't exit.
+func checkCSITranslationLibEntries(csiPluginName, inTreePluginName string) {
+	if !csitranslationlib.IsMigratedCSIDriverByName(csiPluginName) {
+		klog.Warningf("Provisioner name: %s not registered in CSI translation library for performing migrations for %s", csiPluginName, inTreePluginName)
+		return
+	}
+	inTreePluginNameFromLib, err := csitranslationlib.GetInTreeNameFromCSIName(csiPluginName)
```
jsafrane: If we can query in-tree plugin name from CSI driver name, why do we need `--supersedes-in-tree-plugin-name`  options? Should be `--feature-gate` enough? It allows us to enable it by default during beta.

davidz627: oops this is complete opposite of my comment from earlier. I didn't think about enabling it by default during beta. Sorry @ddebroy could you change it back... my bad

ddebroy: Will revert back to querying behavior. That will ensure a single source of truth i.e. the translation library and enablement of migration by default. As far as feature gates is concerned, this will depend on the in-tree feature gates to be enabled for individual plugins (besides migration in general) for the PV controller to pass PVCs with in-tree provisioners to external provisioners.

ddebroy:
Comments above have been addressed and ready for another look.

ddebroy:
Rebased to latest changes.

pkg/controller/controller.go
```diff
@@ -325,6 +329,31 @@ func getVolumeCapability(
 }
 
 func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.PersistentVolume, error) {
+	if options.PVC.Spec.Selector != nil {
```
jsafrane: There is already selector check below.

(this is probably result of rebase)

pkg/controller/controller.go
```diff
@@ -325,6 +329,31 @@ func getVolumeCapability(
...
+		return nil, fmt.Errorf("claim Selector is not supported")
+	}
+
+	createVolumeRequestParameters := options.Parameters
+	migratedVolume := false
+	if p.supportsMigrationFromInTreePluginName != "" {
+		storageClassName := options.PVC.Spec.StorageClassName
+		//TODO(https://github.com/kubernetes-csi/external-provisioner/issues/256): use informers
+		storageClass, err := p.client.StorageV1().StorageClasses().Get(*storageClassName, metav1.GetOptions{})
```
jsafrane: If the only reason for `Get` class is to get the provisioner name, there is one in `PVC.Annotations["volume.beta.kubernetes.io/storage-provisioner"]`. Can this be used instead? Or please add comment why it can't.

pkg/controller/controller.go
```diff
@@ -516,6 +545,13 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 		pv.Spec.PersistentVolumeSource.CSI.FSType = fsType
 	}
 
+	if migratedVolume {
+		pv, err = csitranslationlib.TranslateCSIPVToInTree(pv)
+		if err != nil {
+			return nil, err
```
jsafrane: This leaves the provisioned volume orphaned in the storage backend. We can either delete it (possibly resulting in other errors, oh well...) or return an error that suggests there is an orphan that must be cleaned manually. Both are quite bad, I'd prefer deleting the volume though.

jsafrane: Btw, some log (or even an event?) that this is a critical bug and should be reported / fixed would be helpful.

pkg/controller/controller.go
ddebroy: PVC.Annotations["volume.beta.kubernetes.io/storage-provisioner"] cannot be used for migration scenarios because it's set to the name of the CSI provisioner at https://github.com/kubernetes/kubernetes/blob/596a48dd64bcaa01c1d2515dc79a558a4466d463/pkg/controller/volume/persistentvolume/pv_controller.go#L1383 and https://github.com/kubernetes/kubernetes/blob/596a48dd64bcaa01c1d2515dc79a558a4466d463/pkg/controller/volume/persistentvolume/pv_controller.go#L1396 so that external CSI provisioner can pick it up.

pkg/controller/controller.go
ddebroy: Thanks for catching this. Will remove.

davidz627:
LGTM, will let @jsafrane do the final tagging

jsafrane:
lgtm, please squash the commits (at least no "Address code review comments")

ddebroy:
Commits squashed.

PR #254, opened 2019-03-15
Remove configuration of retry count

jsafrane:
We haven't found use-case where these parameters should be exposed to users. The provisioner will try indefinitelt to provision/delete a volume.

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -45,21 +45,19 @@ import (
 )
 
 var (
-	master                 = flag.String("master", "", "Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.")
-	kubeconfig             = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Either this or master needs to be set if the provisioner is being run out of cluster.")
-	csiEndpoint            = flag.String("csi-address", "/run/csi/socket", "The gRPC endpoint for Target CSI Volume.")
-	connectionTimeout      = flag.Duration("connection-timeout", 0, "This option is deprecated.")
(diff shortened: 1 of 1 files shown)


cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -45,21 +45,19 @@ import (
...
-	kubeconfig             = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Either this or master needs to be set if the provisioner is being run out of cluster.")
-	csiEndpoint            = flag.String("csi-address", "/run/csi/socket", "The gRPC endpoint for Target CSI Volume.")
-	connectionTimeout      = flag.Duration("connection-timeout", 0, "This option is deprecated.")
-	volumeNamePrefix       = flag.String("volume-name-prefix", "pvc", "Prefix to apply to the name of a created volume.")
-	volumeNameUUIDLength   = flag.Int("volume-name-uuid-length", -1, "Truncates generated UUID of a created volume to this length. Defaults behavior is to NOT truncate.")
-	showVersion            = flag.Bool("version", false, "Show version.")
-	enableLeaderElection   = flag.Bool("enable-leader-election", false, "Enables leader election. If leader election is enabled, additional RBAC rules are required. Please refer to the Kubernetes CSI documentation for instructions on setting up these RBAC rules.")
-	provisioningRetryCount = flag.Uint("provisioning-retry-count", 0, "Number of retries of failed provisioning. 0 = retry indefinitely.")
```
davidz627: We are going to have that same error where the provisioner panics if someone is setting this flag. Maybe it is better to keep the flag, and make it do nothing except print a deprecation warning ("we will remove in provisioner version x") and saying it does nothing.

msau42: We have not release a version with this flag so I think it is fine to remove

PR #255, opened 2019-03-18
better deprecated option handling

pohly:
The warning for the deprecated options was only printed
when it was set to a non-default value. When set explicitly to
zero or empty (admittedly unlikely), no warning was printed.

This approach also gets rid of the actual variables and thus catches
code which accidentally still uses it (see
https://github.com/kubernetes-csi/external-provisioner/pull/239).


pohly:
@jsafrane this is an idea that I just had. Not sure whether it's worthwhile.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -44,11 +44,25 @@ import (
 	utilflag "k8s.io/apiserver/pkg/util/flag"
 )
 
+// deprecatedOption can be used to define a deprecated command line
+// flag. Whenever it is set, regardless to which value, a warning is
+// triggered.
+type deprecatedOption struct {
```
davidz627: I think these could exist in a different `util` type package. Maybe a specific file for it

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -44,11 +44,25 @@ import (
...
+	option string
+}
+
+func (d deprecatedOption) String() string { return "" }
+func (d deprecatedOption) Set(value string) error {
```
davidz627: could you be explicit in what interface this is implementing for strict interface checking (mostly for code maintainability and my understanding)
e.g:
`var _ InterfaceName := &deprecatedOption{}`

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -44,11 +44,25 @@ import (
...
+	klog.Warningf("Warning: option %s=%q is deprecated and has no effect", d.option, value)
+	return nil
+}
+func (d deprecatedOption) Type() string { return "" }
+func (d deprecatedOption) Define()      { flag.Var(d, d.option, "This option is deprecated.") }
```
davidz627: could the implementation of `Define` just live inside `Set` instead?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -44,11 +44,25 @@ import (
...
+type deprecatedOption struct {
+	option string
+}
+
+func (d deprecatedOption) String() string { return "" }
+func (d deprecatedOption) Set(value string) error {
+	klog.Warningf("Warning: option %s=%q is deprecated and has no effect", d.option, value)
+	return nil
+}
+func (d deprecatedOption) Type() string { return "" }
```
davidz627: This comment supersedes all my other comments.

in general I'm not sure why we're using a struct here when maybe just a function will do. I think I would prefer that the definition of these deprecated flags more closely mirrored that of a normal flag. Maybe something as simple that would be invoked like this:
`connectionTimeout    = deprecatedflag.Set("connection-timeout")`

pohly: Yes, the API can be made nicer. Let me try something along those lines.

pohly: @davidz627 please have another look. I moved the code into an utility package with a single function as interface. Actually, it's two now, because I also added support for boolean flags.

pkg/deprecatedflags/deprecatedflags.go
```diff
@@ -0,0 +1,57 @@
...
+// return code can be safely ignored and is only provided to support
+// replacing functions like flag.Int in a global variable section.
+func Add(name string) bool {
+	flag.Var(deprecated{name: name}, name, "This option is deprecated.")
+	return true
+}
+
+// AddBool defines a deprecated boolean option. Otherwise it behaves
+// like Add.
+func AddBool(name string) bool {
```
davidz627: is it necessary to distinguish between types (like bool) when the flag is deprecated? Since we are no longer using the value of the flag at all can we just treat them all as strings?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -48,7 +49,8 @@ var (
 	master               = flag.String("master", "", "Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.")
 	kubeconfig           = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Either this or master needs to be set if the provisioner is being run out of cluster.")
 	csiEndpoint          = flag.String("csi-address", "/run/csi/socket", "The gRPC endpoint for Target CSI Volume.")
-	connectionTimeout    = flag.Duration("connection-timeout", 0, "This option is deprecated.")
+	_                    = deprecatedflags.Add("connection-timeout")
+	_                    = deprecatedflags.AddBool("foo")
```
davidz627: remove testing arg (I assume thats what this is)

pkg/deprecatedflags/deprecatedflags.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (d deprecated) String() string { return "" }
+func (d deprecated) Set(value string) error {
+	klog.Warningf("Warning: option %s=%q is deprecated and has no effect", d.name, value)
+	return nil
+}
+func (d deprecated) Type() string     { return "" }
+func (d deprecated) IsBoolFlag() bool { return d.isBool }
+
+var _ flag.Value = deprecated{name: "foo"}
```
davidz627: nit: place above method definitions.

also remove dummy parameter from struct.

pkg/deprecatedflags/deprecatedflags.go
```diff
@@ -0,0 +1,57 @@
...
+	isBool bool
+}
+
+func (d deprecated) String() string { return "" }
+func (d deprecated) Set(value string) error {
+	klog.Warningf("Warning: option %s=%q is deprecated and has no effect", d.name, value)
+	return nil
+}
+func (d deprecated) Type() string     { return "" }
+func (d deprecated) IsBoolFlag() bool { return d.isBool }
```
davidz627: same question, why do we need this?

pkg/deprecatedflags/deprecatedflags.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"k8s.io/klog"
+)
+
+// Add defines a deprecated option which used to take some kind of
+// value. The type of the argument no longer matters, it gets ignored
+// and instead using the option triggers a deprecation warning. The
+// return code can be safely ignored and is only provided to support
+// replacing functions like flag.Int in a global variable section.
+func Add(name string) bool {
```
davidz627: why do these need a return value? can they just return nothing instead?

I'm thinking maybe this was a creative solutiont there is a variable you can set to `_` when you use these functions. If that is the case and it is truly necessary, lets at least document that in a comment here so that ppl in the future aren't confused

pkg/deprecatedflags/deprecatedflags.go
pohly: Yes, it's necessary. Only a boolean flag can be used without a value (`cmd -enable-foo`). In order to keep existing command line invocations valid, we still have to mark up the flag the same way (bool or not bool) as before.

pkg/deprecatedflags/deprecatedflags.go
pohly: Yes, that's what it is for. `var _ = something()` doesn't work unless `something()` returns a value.

I tried to explain that in the last sentence of the comment ("is only provided to support
replacing functions like flag.Int in a global variable section"). Would an example be clearer?

davidz627: oh i just have questionable reading comprehension skills. that makes sense. Thanks

cmd/csi-provisioner/csi-provisioner.go
pohly: Oops, forgot to force-push. I had fixed that already.

pkg/deprecatedflags/deprecatedflags.go
pohly: `IsBoolFlag` is how the `flag` package determines whether a value needs the special bool handling.

pkg/deprecatedflags/deprecatedflags.go
pohly: Done.

jsafrane:
I'd appreciate this in csi-lib-utils, because more sidecars have deprecated `--connection-timeout`.

pohly:
I'll move the utility package to csi-lib-utils.

Gopkg.toml
```diff
@@ -35,21 +35,16 @@
...
-
 # TODO: remove when official 0.4.0 tagged
 [[constraint]]
   name = "github.com/kubernetes-csi/csi-lib-utils"
   version = ">=0.4.0-rc1"
 
+# TODO: remove when released
+[[constraint]]
+  name = "github.com/kubernetes-csi/external-snapshotter"
+  branch = "master"
```
xing-yang: Setting it to master is not safe though.  Maybe we should cut a release of external-snapshotter first.

msau42: I think it is fine. The lock file ensures that we keep on using the correct commit.  And updates are controlled

pohly: I agree. Whoever does the next `dep ensure -update` will have to check the TODOs, and thus will ensure that we do not update beyond the next external-snapshotter release.

pohly:
@msau42 force-pushed a version which passes "make test" and "make build" locally.

PR #262, opened 2019-04-02
vendor: update to k8s.io 1.14, avoid glog

pohly:
The previous vendor update ended up pulling in a release-1.10 branch
of k8s.io/apiserver and with it glog, which broke the resulting
executable (compiled, but failed during init due the double definition
of glog flags).

We have to ensure that this does not happen by explicitly choosing
k8s.io package revisions. In this case, 1.14.0 is used (latest
stable), which has a slight API change (see
https://github.com/kubernetes/kubernetes/pull/71302).


pohly:
The "canary" test job now passes again.

PR #263, opened 2019-04-03
Add Changelog for 1.1 and update README

msau42:


CHANGELOG-1.1.md
```diff
@@ -0,0 +1,32 @@
+# Changelog since v1.0.1
+
+## Breaking Changes
+* Support for the alpha CSINodeInfo CRD has been removed.
+
+## Deprecations
+* Command line flag `--connection-timeout` is deprecated and has no effect.
+* Command line flag `--provisioner` is deprecated and has no effect.
+* Command line flag `--provisioning-retry-count` is deprecated and has no effect.
```
msau42: @jsafrane these two were actually removed and not using the new deprecatedflags lib. Could you add them back?

jsafrane: I can see `--provisioner` in current master:
https://github.com/kubernetes-csi/external-provisioner/blob/8a62ab7dd9915f5c56ae5f4664e94e4930034ee0/cmd/csi-provisioner/csi-provisioner.go#L60

And IMO  `--provisioning-retry-count` was never released, I can't find it in any tag.

msau42: Thanks for checking. I removed the retry count flags

PR #265, opened 2019-04-11
Prow testing

pohly:


msau42:
I'm surprised the tests are passing. Didn't we break backwards compatibility with the new RBAC rules?

pohly:
The periodic tests failed because the unmodified RBAC rules:
https://k8s-testgrid.appspot.com/sig-storage-csi#canary-on-1-13
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-csi-canary-on-kubernetes-1-13/1116643310904020992

The pull jobs intentionally use the modified RBAC rules and therefore tests pass. That is done this way to allow breaking changes which change code and RBAC rules. Otherwise it wouldn't be possible to merge such changes without fiddling with testing. Note that the initial PR changed the code but not the RBAC rules. *That* would have been caught by these pull jobs.

PR #271, opened 2019-04-28
fix provision fail when volume size is unknown

### LINKED ISSUE #127: CreateVolume fails if returned capacity is 0
According to CSI spec `CreateVolume` should be able to return capacity of 0, which means it is unknown. See https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/v0/csi.pb.go#L1266-L1269

But external-provisioner is failing to provision when checking the capacity since it is less than requested capacity https://github.com/kubernetes-csi/external-provisioner/blob/d946213178b9a5107f203fc047526a0c2dd2df63/pkg/controller/controller.go#L381

hoyho:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

According to CSI spec CreateVolume should be able to return capacity = 0, which means it is unknown.
However, provisioner will ignore it.

Fixes #127 

```release-note
CSI driver can return `CreateVolumeResponse` with size 0, which means unknown volume size. 
In this case, Provisioner will use PVC requested size as PV size rather than 0 bytes
```

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -767,6 +767,7 @@ type provisioningTestcase struct {
 	getSecretRefErr   bool
 	getCredentialsErr bool
 	volWithLessCap    bool
+	volWithZeroCap    bool
 	expectedPVSpec    *pvSpec
 	withSecretRefs    bool
 	createVolumeError error
@@ -1341,6 +1342,34 @@ func TestProvision(t *testing.T) {
 			expectErr:   true,
 			expectState: controller.ProvisioningInBackground,
 		},
+		"provision with size 0": {
+			volOpts: controller.ProvisionOptions{
+				StorageClass: &storagev1.StorageClass{
+					Parameters: map[string]string{
+						"fstype": "ext3",
+					},
+					ReclaimPolicy: &deletePolicy,
+				},
+				PVName: "test-name",
+				PVC:    createFakePVC(requestedBytes),
+			},
+			volWithZeroCap: true,
+			expectedPVSpec: &pvSpec{
+				Name:          "test-testi",
+				ReclaimPolicy: v1.PersistentVolumeReclaimDelete,
+				Capacity: v1.ResourceList{
+					v1.ResourceName(v1.ResourceStorage): bytesToGiQuantity(requestedBytes),
+				},
+				CSIPVS: &v1.CSIPersistentVolumeSource{
+					Driver:       "test-driver",
+					VolumeHandle: "test-volume-id",
+					FSType:       "ext3",
+					VolumeAttributes: map[string]string{
+						"storage.kubernetes.io/csiProvisionerIdentity": "test-provisioner",
+					},
+				},
+			},
+		},
 	}
 
 	for k, tc := range testcases {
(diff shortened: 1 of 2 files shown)


pkg/controller/controller_test.go
```diff
@@ -1093,6 +1094,32 @@ func TestProvision(t *testing.T) {
...
+				Parameters: map[string]string{
+					"fstype": "ext3",
+				},
+			},
+			volWithZeroCap: true,
+			expectedPVSpec: &pvSpec{
+				Name:          "test-testi",
+				ReclaimPolicy: v1.PersistentVolumeReclaimDelete,
+				Capacity: v1.ResourceList{
+					v1.ResourceName(v1.ResourceStorage): bytesToGiQuantity(0),
```
msau42: I'm unsure if returning 0 bytes in the PV is going to be helpful.  PVC won't bind to it because the capacity is not >= the requested.

Should we return size == PVC request in this case?  cc @jsafrane

hoyho: Yes, Seems 0 bytes is not a valid size here.  Apiserver will not create PV obj with size 0 . If csi client return CreateVolumeResponse with size 0, Maybe we can reset to PVC requested size

hoyho:
Something weird.. Don't know why CI build failed. 
Actually, I can build it by running `make -k all test` . Here is the output
```bash
 ⚡ root@vm43  ~/diskA/gopath/src/github.com/kubernetes-csi/external-provisioner   bugfix/zeroCap  make -k all test
mkdir -p bin
CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-X main.version=v1.1.0-8-ge561692 -extldflags "-static"' -o ./bin/csi-provisioner ./cmd/csi-provisioner

### test-go:
go test `go list ./... | grep -v -e 'vendor' -e '/test/e2e$' `
?   	github.com/kubernetes-csi/external-provisioner/cmd/csi-provisioner	[no test files]
ok  	github.com/kubernetes-csi/external-provisioner/pkg/controller	(cached)
?   	github.com/kubernetes-csi/external-provisioner/pkg/features	[no test files]
?   	github.com/kubernetes-csi/external-provisioner/release-tools	[no test files]

### test-vet:
go vet `go list ./... | grep -v vendor `

### test-fmt:
files=$(find . -name '*.go' | grep -v './vendor' ); \
if [ $(gofmt -d $files | wc -l) -ne 0 ]; then \
	echo "formatting errors:"; \
	gofmt -d $files; \
	false; \
fi

### test-vendor:
skipping check, dep >= 0.5 required

### test-subtree:
./release-tools/verify-subtree.sh release-tools
release-tools is a clean copy of upstream.

### test-shellcheck:
/usr/bin/docker

/root/diskA/gopath/src/github.com/kubernetes-csi/external-provisioner/release-tools:
Using shellcheck 0.6.0 docker image.
Congratulations! All shell files are passing lint.
```

How to fix it?

pohly:
Try merging your tree into master. Perhaps something changed there which allows a merge to succeed, but then results in some unbuildable code.

If so, you need to rebase and adapt your code.

hoyho:
> Try merging your tree into master. Perhaps something changed there which allows a merge to succeed, but then results in some unbuildable code.
> 
> If so, you need to rebase and adapt your code.

Thank you.  I can reproduce that error now after rebase

pkg/controller/controller_test.go
```diff
@@ -1372,6 +1401,9 @@ func runProvisionTest(t *testing.T, k string, tc provisioningTestcase, requested
 		out.Volume.CapacityBytes = int64(80)
 		controllerServer.EXPECT().CreateVolume(gomock.Any(), gomock.Any()).Return(out, nil).Times(1)
 		controllerServer.EXPECT().DeleteVolume(gomock.Any(), gomock.Any()).Return(&csi.DeleteVolumeResponse{}, nil).Times(1)
+	} else if tc.volWithZeroCap {
```
msau42: I think it would be good to refactor this test to have all of this test case specific behavior passed in as a func() with the rest of the test inputs.  But that can be done as a separate PR.

msau42:
This lgtm, also would like @jsafrane's input on if this should be the proper behavior.
/approve
/assign @jsafrane

msau42:
Can you also add a release note for this?

hoyho:
> Can you also add a release note for this?

done

pkg/controller/controller_test.go
hoyho: Ok，I think we can add an issue to discuss more detail about refactoring test and resolve it in another PR

hoyho:
PTAL @jsafrane

jsafrane:
Looks ok, can you please rebase it?

hoyho:
> Looks ok, can you please rebase it?

done @jsafrane

hoyho:
The job pull-kubernetes-csi-external-provisioner-1-14-on-kubernetes-1-14 seems to be stuck?

PR #272, opened 2019-05-07
Refactor/gomod

hoyho:
**What type of PR is this?**

 /kind feature

remove tons of dependency from `vendor` 
use go module to replace dep as following reasons:

- According to [The Go Blog](https://blog.golang.org/using-go-modules), Go 1.13 will enable go modules by default and deprecate GOPATH mode.
- Save lots of disk space because the vendor directory may include 10~100X code than our source code. (It just like Node.js module system, The whole Github live in each project)
- Kubernetes have finished switching to go modules, This PR should be safety and achievable.
- Maybe increase speed and reliability of building and verifying

Fixes #

somewhere is still using dep? 

```release-note
Migrating dep to go modules
```


pohly:
We are already experimenting with Go modules in https://github.com/kubernetes-csi/csi-test/pull/188. Instead of starting yet another discussion around that here, let's finish that PR first.

hoyho:
fine, hope we can merge that PR ASAP

PR #274, opened 2019-05-07
Add secret support for Provision and Delete from pvc name and namespace

### LINKED ISSUE #170: Need pvc namespace passed to CSI driver
The in-tree Portworx Kubernetes driver currently relies on determining the namespace of the PVC during creation.

There are two issues with the current implementations in which none pass this information during creation of CSI volume:

1: [On mounts](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/csi-api/pkg/apis/csi/v1alpha1/types.go#L78-L90) the side car containers can pass the pod.Namespace (or more likely, the pvc name and namespace) to the CSI driver. We would like to have this on Creation.

and

2: [Secrets](https://github.com/kubernetes-csi/external-provisioner/blob/b336e3dbd66c822f32cdf5c6e839f32be2ed7dfc/pkg/controller/controller.go#L537) Only passes secrets of the PVC namespace for every call _except_ during creation.

We need `1:` to work and also would really to have `2:` to be more efficient in obtaining secrets.

### LINKED ISSUE #233: Allow PVC name/namespace as template for provisioner-secret
**Background:**

Issue https://github.com/kubernetes-csi/external-provisioner/issues/170

You can specify a `provisioner-secret-name` and `provisioner-secret-namespace` as a parameter in the `StorageClass`. Every volume provisioned using that `StorageClass` uses that secret. There was a desire to be able to specify a different secret for every volume generated.

To enable this @liggitt introduced templating so that secret references could be different per volume (see pull #69 and see [documentation](https://kubernetes-csi.github.io/docs/secrets-and-credentials.html#createdelete-volume-secret)).

The problem is that for `provisioner-secret-name` and `provisioner-secret-namespace` the only supported template variable is `${pv.name}`. They do not support PVC.

**Problem:**
This is problematic because `pv.name` is generated *during* volume provisioning, it is hard to predict what the generated PV name will be, so it is hard to provision a secret named `pv.name` before the volume is provisioned.

Effectively it means that  `provisioner-secret-name` and `provisioner-secret-namespace` don't really support templating and have to be hard coded so provisioning secrets effectively can not be different per volume.

**Proposal:**
Support the template variable `${pvc.namespace}`, `${pvc.name}`, and `${pvc.annotations['<ANNOTATION_KEY>']}` for both the `provisioner-secret-name` and `provisioner-secret-namespace` secret values. This would enable a different secret per provisioned volume.

The counter argument to this proposal per @liggitt:

> The reason the secrets in the pvc namespace are not made available for paired create/delete operations is that the PVC and its namespace may not exist at deletion time

**My suggestion:**
Unless there is a security issue, relax the limitation and adopt the proposal so that storage systems (like Portworx) where secrets are not needed for volume delete can do this. And add documentation to discourage storage systems that require secret on delete from doing this. 

CC @lpabon @harsh-px @liggitt @msau42

ggriffiths:
Signed-off-by: Grant Griffiths <ggp493@gmail.com>

**What type of PR is this?**
/kind feature

* This allows Provision to pass the correct PVC object so that the k8s secrets can be pulled and used during provisioning. 

Fixes #170 
Fixes #233

* I'm still testing this change on my k8s cluster, which is why I've added the WIP tag.

```release-note
Users can now provide a secret name and namespace during provision by passing the correct storage class parameters: "provisioner-secret-name" and "provisioner-secret-namespace"
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -444,8 +444,12 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
 	rep := &csi.CreateVolumeResponse{}
 
 	// Resolve provision secret credentials.
-	// No PVC is provided when resolving provision/delete secret names, since the PVC may or may not exist at delete time.
-	provisionerSecretRef, err := getSecretReference(provisionerSecretParams, options.StorageClass.Parameters, pvName, nil)
+	provisionerSecretRef, err := getSecretReference(provisionerSecretParams, options.StorageClass.Parameters, pvName, &v1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      options.PVC.Name,
+			Namespace: options.PVC.Namespace,
+		},
+	})
 	if err != nil {
 		return nil, err
 	}
@@ -684,18 +688,23 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
 	if len(storageClassName) != 0 {
 		if storageClass, err := p.client.StorageV1().StorageClasses().Get(storageClassName, metav1.GetOptions{}); err == nil {
 			// Resolve provision secret credentials.
-			// No PVC is provided when resolving provision/delete secret names, since the PVC may or may not exist at delete time.
-			provisionerSecretRef, err := getSecretReference(provisionerSecretParams, storageClass.Parameters, volume.Name, nil)
+			provisionerSecretRef, err := getSecretReference(provisionerSecretParams, storageClass.Parameters, volume.Name, &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      volume.Spec.ClaimRef.Name,
+					Namespace: volume.Spec.ClaimRef.Namespace,
+				},
+			})
 			if err != nil {
 				return err
 			}
+
 			credentials, err := getCredentials(p.client, provisionerSecretRef)
 			if err != nil {
-				return err
+				// Continue with deletion, as the secret may have already been deleted.
+				klog.Errorf("Failed to get credentials for volume %s: %s", volume.Name, err.Error())
 			}
 			req.Secrets = credentials
 		}
-
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
 	defer cancel()
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -558,6 +558,20 @@ func TestGetSecretReference(t *testing.T) {
 			pvc:          nil,
 			expectRef:    &v1.SecretReference{Name: "name", Namespace: "ns"},
 		},
+		"simple - valid, pvc name and namespace": {
+			secretParams: provisionerSecretParams,
+			params: map[string]string{
+				provisionerSecretNameKey:      "param-name",
+				provisionerSecretNamespaceKey: "param-ns",
+			},
+			pvc: &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "name",
+					Namespace: "ns",
+				},
+			},
+			expectRef: &v1.SecretReference{Name: "param-name", Namespace: "param-ns"},
+		},
 		"simple - invalid name": {
 			secretParams: nodePublishSecretParams,
 			params:       map[string]string{nodePublishSecretNameKey: "bad name", nodePublishSecretNamespaceKey: "ns"},
@@ -572,7 +586,20 @@ func TestGetSecretReference(t *testing.T) {
 			expectRef:    nil,
 			expectErr:    true,
 		},
-		"template - valid": {
+		"template - PVC name annotations not supported for Provision and Delete": {
+			secretParams: provisionerSecretParams,
+			params: map[string]string{
+				prefixedProvisionerSecretNameKey: "static-${pv.name}-${pvc.namespace}-${pvc.name}-${pvc.annotations['akey']}",
+			},
+			pvc: &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "name",
+					Namespace: "ns",
+				},
+			},
+			expectErr: true,
+		},
+		"template - valid nodepublish secret ref": {
 			secretParams: nodePublishSecretParams,
 			params: map[string]string{
 				nodePublishSecretNameKey:      "static-${pv.name}-${pvc.namespace}-${pvc.name}-${pvc.annotations['akey']}",
@@ -587,6 +614,63 @@ func TestGetSecretReference(t *testing.T) {
 				},
 			},
 			expectRef: &v1.SecretReference{Name: "static-pvname-pvcnamespace-pvcname-avalue", Namespace: "static-pvname-pvcnamespace"},
+		},
+		"template - valid provisioner secret ref": {
+			secretParams: provisionerSecretParams,
+			params: map[string]string{
+				provisionerSecretNameKey:      "static-provisioner-${pv.name}-${pvc.namespace}-${pvc.name}",
+				provisionerSecretNamespaceKey: "static-provisioner-${pv.name}-${pvc.namespace}",
+			},
+			pvName: "pvname",
+			pvc: &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pvcname",
+					Namespace: "pvcnamespace",
+				},
+			},
+			expectRef: &v1.SecretReference{Name: "static-provisioner-pvname-pvcnamespace-pvcname", Namespace: "static-provisioner-pvname-pvcnamespace"},
+		},
+		"template - valid, with pvc.name": {
+			secretParams: provisionerSecretParams,
+			params: map[string]string{
+				provisionerSecretNameKey:      "${pvc.name}",
+				provisionerSecretNamespaceKey: "ns",
+			},
+			pvc: &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pvcname",
+					Namespace: "pvcns",
+				},
+			},
+			expectRef: &v1.SecretReference{Name: "pvcname", Namespace: "ns"},
+		},
+		"template - valid, provisioner with pvc name and namepsace": {
+			secretParams: provisionerSecretParams,
+			params: map[string]string{
+				provisionerSecretNameKey:      "${pvc.name}",
+				provisionerSecretNamespaceKey: "${pvc.namespace}",
+			},
+			pvc: &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pvcname",
+					Namespace: "pvcns",
+				},
+			},
+			expectRef: &v1.SecretReference{Name: "pvcname", Namespace: "pvcns"},
+		},
+		"template - valid, static pvc name and templated namespace": {
+			secretParams: provisionerSecretParams,
+			params: map[string]string{
+				provisionerSecretNameKey:      "static-name-1",
+				provisionerSecretNamespaceKey: "${pvc.namespace}",
+			},
+			pvc: &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "name",
+					Namespace: "ns",
+				},
+			},
+			expectRef: &v1.SecretReference{Name: "static-name-1", Namespace: "ns"},
 		},
 		"template - invalid namespace tokens": {
 			secretParams: nodePublishSecretParams,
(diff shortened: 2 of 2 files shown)


ggriffiths:
Verified that this change works on my k8s cluster with our CSI driver. Ready for review.

pkg/controller/controller.go
```diff
@@ -445,8 +445,7 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 	rep := &csi.CreateVolumeResponse{}
 
 	// Resolve provision secret credentials.
-	// No PVC is provided when resolving provision/delete secret names, since the PVC may or may not exist at delete time.
-	provisionerSecretRef, err := getSecretReference(provisionerSecretParams, createVolumeRequestParameters, pvName, nil)
+	provisionerSecretRef, err := getSecretReference(provisionerSecretParams, createVolumeRequestParameters, pvName, options.PVC)
```
msau42: Are you planning to add Delete in a separate PR?

ggriffiths: I'll add it into this PR actually @msau42

pkg/controller/controller.go
```diff
@@ -664,15 +663,23 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
 	storageClassName := volume.Spec.StorageClassName
 	if len(storageClassName) != 0 {
 		if storageClass, err := p.client.StorageV1().StorageClasses().Get(storageClassName, metav1.GetOptions{}); err == nil {
+
+			// Get PVC for secret reference if the volume is bound.
+			var pvc *v1.PersistentVolumeClaim
```
msau42: Can you add a unit test for this deletion case?

pkg/controller/controller.go
```diff
@@ -664,15 +663,23 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
...
+			if volume.Spec.ClaimRef != nil {
+				pvc, _ = p.client.CoreV1().PersistentVolumeClaims(volume.Spec.ClaimRef.Namespace).Get(volume.Spec.ClaimRef.Name, metav1.GetOptions{})
```
msau42: No need to check for this.  This will always fail because volume deletion is triggered by PVC being gone.

pkg/controller/controller_test.go
```diff
@@ -585,6 +596,48 @@ func TestGetSecretReference(t *testing.T) {
...
+			},
+			pvc: &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pvcname",
+					Namespace: "pvcns",
+				},
+			},
+			expectRef: &v1.SecretReference{Name: "pvcname", Namespace: "pvcns"},
+		},
+		"template - valid, static pvc name and templated namespace": {
```
msau42: Because volume deletion is triggered off of PVC deletion, we don't support PVC.annotations as a template for create/delete.  Can you add logic in getSecretReference for that, and also unit tests?

ggriffiths: I added a new `getProvisionSecretReference` that checks for this, since `getSecretReference` is used for other secret references.

pkg/controller/controller.go
ggriffiths: Added simple Delete unit tests.

pkg/controller/controller.go
```diff
@@ -663,19 +663,17 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
 	storageClassName := volume.Spec.StorageClassName
 	if len(storageClassName) != 0 {
 		if storageClass, err := p.client.StorageV1().StorageClasses().Get(storageClassName, metav1.GetOptions{}); err == nil {
```
msau42: log a warning if it's an error?

pkg/controller/controller.go
```diff
@@ -445,7 +445,7 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 	rep := &csi.CreateVolumeResponse{}
 
 	// Resolve provision secret credentials.
-	provisionerSecretRef, err := getSecretReference(provisionerSecretParams, createVolumeRequestParameters, pvName, options.PVC)
+	provisionerSecretRef, err := getProvisionSecretReference(provisionerSecretParams, createVolumeRequestParameters, pvName, options.PVC)
```
msau42: Would it be simpler to just pass in a new PVC object with only name and namespace set? Like in delete?

pkg/controller/controller.go
```diff
@@ -756,6 +754,20 @@ func verifyAndGetSecretNameAndNamespaceTemplate(secret secretParamsMap, storageC
 	}
 }
 
+// getProvisionSecretReference returns a reference to the provisioning secret,
+// with an additional check for unsupported PVC annotations.
+func getProvisionSecretReference(secretParams secretParamsMap, storageClassParams map[string]string, pvName string, pvc *v1.PersistentVolumeClaim) (*v1.SecretReference, error) {
+	// validate that no annotations are being passed in as secrets
+	for k, v := range pvc.Annotations {
```
msau42: This should be storageclass parameters, not pvc.Annotations

pkg/controller/controller.go
```diff
@@ -756,6 +754,20 @@ func verifyAndGetSecretNameAndNamespaceTemplate(secret secretParamsMap, storageC
...
+		if (strings.Contains(k, prefixedProvisionerSecretNameKey) && strings.Contains(v, "pvc.annotations")) ||
```
msau42: This may look nicer with a case statement

pkg/controller/controller.go
ggriffiths: True, the new function was due to my mix up of PVC.Annotations and SC.Parameters. Changed to use a new PVC object with name and namespace set.

pkg/controller/controller.go
ggriffiths: Changed to be Errorf

pkg/controller/controller.go
```diff
@@ -787,6 +791,11 @@ func getProvisionSecretReference(secretParams secretParamsMap, storageClassParam
 // - the resolved name is not a valid secret name
 // - the resolved namespace is not a valid namespace name
 func getSecretReference(secretParams secretParamsMap, storageClassParams map[string]string, pvName string, pvc *v1.PersistentVolumeClaim) (*v1.SecretReference, error) {
+	// Ensure that no unsupported SC parameters are passed in.
```
msau42: This will cause secret templates for other operations to fail. It's probably simplest to just let resolveTemplate() fail.

ggriffiths: Makes sense. Removed this validation.

pkg/controller/controller_test.go
```diff
@@ -569,21 +580,70 @@ func TestGetSecretReference(t *testing.T) {
 			expectRef:    nil,
 			expectErr:    true,
 		},
+		"simple - PVC name annotations not supported for Provision and Delete": {
+			secretParams: provisionerSecretParams,
+			params: map[string]string{
+				prefixedProvisionerSecretNameKey: "static-${pv.name}-${pvc.namespace}-${pvc.name}-${pvc.annotations['akey']}",
+			},
+			pvc:       &v1.PersistentVolumeClaim{},
```
msau42: Can you fill in name and namespace for this pvc too, so we know it's not failing because of name or namespace missing?  Also does pv name need to be passed in?

pkg/controller/controller_test.go
```diff
@@ -569,21 +580,70 @@ func TestGetSecretReference(t *testing.T) {
...
 				},
 			},
-			expectRef: &v1.SecretReference{Name: "static-pvname-pvcnamespace-pvcname-avalue", Namespace: "static-pvname-pvcnamespace"},
+			expectRef: &v1.SecretReference{Name: "static-pvname-pvcnamespace-pvcname", Namespace: "static-pvname-pvcnamespace"},
+		},
+		"template - valid, with pvc name": {
+			secretParams: provisionerSecretParams,
+			params: map[string]string{
+				provisionerSecretNameKey:      "${pvc.name}",
+				provisionerSecretNamespaceKey: "ns",
```
msau42: make this ns something different than the PVC ns so we know we didn't take it from the pvc.

pkg/controller/controller_test.go
```diff
@@ -1787,3 +1847,88 @@ func TestProvisionWithMountOptions(t *testing.T) {
...
+						prefixedProvisionerSecretNameKey: "static-${pv.name}-${pvc.namespace}-${pvc.name}-${pvc.annotations['akey']}",
+					},
+				},
+				Spec: v1.PersistentVolumeSpec{
+					PersistentVolumeSource: v1.PersistentVolumeSource{},
+				},
+			},
+			expectErr: true,
+		},
+		"simple - valid": deleteTestcase{
```
msau42: This should be failing since annotations is not supported?

pkg/controller/controller_test.go
```diff
@@ -1787,3 +1847,88 @@ func TestProvisionWithMountOptions(t *testing.T) {
...
+				},
+				Spec: v1.PersistentVolumeSpec{
+					PersistentVolumeSource: v1.PersistentVolumeSource{
+						CSI: &v1.CSIPersistentVolumeSource{
+							VolumeHandle: "vol-id-1",
+						},
+					},
+				},
+			},
+			expectErr: false,
```
msau42: Is there a way to validate what secret reference we generated?  And expect if it got set or ignored?

ggriffiths: I don't think easily from testing Delete at a high level. I think getSecretReference tests should cover that.

pkg/controller/controller_test.go
ggriffiths: This wasn't failing because I was using PV Annotations instead of SC Parameters in the test.

ggriffiths:
Just re-tested this on my local k8s cluster and Provision/delete are working. 

Also `pvc.annotations` as SC params are correctly causing errors in provision/delete:

 `error resolving value "static-${pv.name}-${pvc.namespace}-${pvc.name}-${pvc.annotations['akey']}": invalid tokens: ["pvc.annotations['akey']" "pvc.name"]`

pkg/controller/controller_test.go
```diff
@@ -558,6 +558,17 @@ func TestGetSecretReference(t *testing.T) {
 			pvc:          nil,
 			expectRef:    &v1.SecretReference{Name: "name", Namespace: "ns"},
 		},
+		"simple - valid, pvc name and namespace": {
+			secretParams: provisionerSecretParams,
+			params:       map[string]string{provisionerSecretNameKey: "name", provisionerSecretNamespaceKey: "ns"},
```
msau42: Can you make the keys used in the params different from the values in the PVC? That way we make sure we're not accidentally using pvc values.

pkg/controller/controller_test.go
```diff
@@ -572,21 +583,75 @@ func TestGetSecretReference(t *testing.T) {
...
+					Namespace: "ns",
+				},
+			},
+			expectErr: true,
+		},
 		"template - valid": {
 			secretParams: nodePublishSecretParams,
 			params: map[string]string{
-				nodePublishSecretNameKey:      "static-${pv.name}-${pvc.namespace}-${pvc.name}-${pvc.annotations['akey']}",
+				nodePublishSecretNameKey:      "static-${pv.name}-${pvc.namespace}-${pvc.name}",
```
msau42: annotations are allowed for nodepublish right?

pkg/controller/controller_test.go
```diff
@@ -1843,3 +1908,130 @@ func TestProvisionWithMountOptions(t *testing.T) {
...
+					Namespace: "ns",
+				},
+				Parameters: map[string]string{
+					prefixedProvisionerSecretNameKey:      "static-${pv.name}-${pvc.namespace}-${pvc.name}-${pvc.annotations['akey']}",
+					prefixedProvisionerSecretNamespaceKey: "static-${pv.name}-${pvc.namespace}-${pvc.name}-${pvc.annotations['akey']}",
+				},
+			},
+			expectErr: true,
+		},
+		"simple - valid case": deleteTestcase{
```
msau42: Also one more test case where claimRef is set

pkg/controller/controller_test.go
ggriffiths: Yeah, they are. Just made this into two unit tests - one for nodePublish secrets and one for provisioner secrets

pkg/controller/controller_test.go
ggriffiths: Added this test case.

pkg/controller/controller_test.go
```diff
@@ -588,6 +615,78 @@ func TestGetSecretReference(t *testing.T) {
...
+			pvName: "pvname",
+			pvc: &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pvcname",
+					Namespace: "pvcnamespace",
+				},
+			},
+			expectRef: &v1.SecretReference{Name: "static-provisioner-pvname-pvcnamespace-pvcname", Namespace: "static-provisioner-pvname-pvcnamespace"},
+		},
+		"template - invalid provisioner secret, annotations not supported": {
```
msau42: this is already at L589?

msau42:
Also please squash your commits

pkg/controller/controller_test.go
ggriffiths: Good catch, removed this duplicate test.

ggriffiths:
Squashed into one commit. Thanks for the review!

msau42:
/lgtm

Thanks!

PR #277, opened 2019-05-15
Bump provisioner lib and migration library dependencies. Update migration translation to translate the entire storage class.

davidz627:
This PR updates the controller to translate the entire storage class when the migration path is hit. This allows us to modify things such as topology, parameters, mount options etc.

/kind feature
/sig storage

Fixes: https://github.com/kubernetes/kubernetes/issues/77235

```release-note
Fixes migration scenarios for Topology, fstype, and accessmodes for the kubernetes.io/gce-pd in-tree plugin
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"os"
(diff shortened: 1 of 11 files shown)


pkg/controller/controller.go
```diff
@@ -558,6 +558,14 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 	return pv, nil
 }
 
+func getReclaimPolicy(sc *storagev1.StorageClass) v1.PersistentVolumeReclaimPolicy {
+	reclaim := v1.PersistentVolumeReclaimDelete
```
msau42: I don't think we should default reclaim policy if nothing is set in the StorageClass.  If it's not set, then we should not fill it in, and let api server default it.

davidz627: done

PR #281, opened 2019-05-19
[release-1.0] Add secret support for Provision and Delete from pvc name and namespace

### LINKED ISSUE #170: Need pvc namespace passed to CSI driver

### LINKED ISSUE #233: Allow PVC name/namespace as template for provisioner-secret

oleksiys:
Backport #274 to Kubernetes 1.13

**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**

- This allows Provision to pass the correct PVC object so that the k8s secrets can be pulled and used during provisioning.
- We need this fix in Kubernetes 1.13

**Which issue(s) this PR fixes:**
- Fixes #170
- Fixes #233

Does this PR introduce a user-facing change?:

```release-note
Users can now provide a secret name and namespace during provision by passing the correct storage class parameters: "provisioner-secret-name" and "provisioner-secret-namespace"
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	_ "k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/klog"
 
 	"github.com/golang/glog"
 
@@ -546,8 +547,12 @@ func (p *csiProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 	rep := &csi.CreateVolumeResponse{}
 
 	// Resolve provision secret credentials.
-	// No PVC is provided when resolving provision/delete secret names, since the PVC may or may not exist at delete time.
-	provisionerSecretRef, err := getSecretReference(provisionerSecretParams, options.Parameters, pvName, nil)
+	provisionerSecretRef, err := getSecretReference(provisionerSecretParams, options.Parameters, pvName, &v1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      options.PVC.Name,
+			Namespace: options.PVC.Namespace,
+		},
+	})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


shay-berman:
@oleksiys is there any documentation that need to be updated as part of this PR?  Like what is the user impact here.

oleksiys:
@shay-berman well in general external provisioner document [1] must be updated. But I'd say it's more related to the original fix #274 than this specific portback.

[1] https://kubernetes-csi.github.io/docs/secrets-and-credentials.html#createdelete-volume-secret

saad-ali:
/lgtm
/approve

Should this also be cherry picked to `release-1.1` in addition to `release-1.0` branch?

oleksiys:
@saad-ali AFAICT it's not related to the PR itself:
```
Activated service account credentials for: [pr-kubekins@kubernetes-jenkins-pull.iam.gserviceaccount.com]
+ ./.prow.sh
/usr/local/bin/runner.sh: line 117: ./.prow.sh: No such file or directory
+ EXIT_VALUE=127
```
Is it an infrastructure problem or am I missing something?

msau42:
cc @pohly 
I believe we didn't add the prow tests until 1.1.  Do we need to exclude these jobs from older branches?

pohly:
> I believe we didn't add the prow tests until 1.1. Do we need to exclude these jobs from older branches?

Yes. This was already done for most of the jobs, but due to an oversight was missing from the "unit" test job.

Fix is in https://github.com/kubernetes/test-infra/pull/12866

oleksiys:
Thank you!

/retest

PR #282, opened 2019-05-21
Introduce new flag - strict-topology

### LINKED ISSUE #221: Wrong AccessibilityRequirement passed in CreateVolumeRequest

avalluri:
With the current implementation, In delayed binding case, CSI driver is offered
with all nodes topology that are matched with 'selected node' topology keys in
`CreateVolumeRequest.AccessibilityRequirements`. So this allows the driver to
select any node from the passed `preferred/requisite` list to create volume. But this
results in scheduling failure when the volume created on a node other than
Kubernetes selected node.

To address this, introduced new flag "--strict-topology', when set, in case of
delayed binding, the driver is offered with only selected node topology, so that
the driver has to create the volume on this node.

This new flag can be used by drivers that support strict topology for volumes with delayed binding.

**What type of PR is this?**
/kind bug
/kind design

In case of delayed binding, creating volume on the different topology that is not accessed by the Kubernetes selected node for Pod scheduling, leads to unresolvable scheduling failures. So we should not allow the driver to create such volumes. We can avoid this by passing right/strict accessibility topologies instead of 'aggregated topology' to CreateVolume request.

Fixes #221 

Yes

```release-note
support strict topology for volumes with delayed binding
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 ### Command line options
 
-#### Recommended optional arguments"
+#### Recommended optional arguments
 * `--csi-address <path to CSI socket>`: This is the path to the CSI driver socket inside the pod that the external-provisioner container will use to issue CSI operations (`/run/csi/socket` is used by default).
 
 * `--enable-leader-election`: Enables leader election. This is mandatory when there are multiple replicas of the same external-provisioner running for one CSI driver. Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.
@@ -63,6 +63,8 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 #### Other recognized arguments
 * `--feature-gates <gates>`: A set of comma separated `<feature-name>=<true|false>` pairs that describe feature gates for alpha/experimental features. See [list of features](#feature-status) or `--help` output for list of recognized features. Example: `--feature-gates Topology=true` to enable Topology feature that's disabled by default.
+
+* `--strict-topology`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of delayed binding. See [the table below](#topology-support) for an explanation how this option changes the result. This option has no effect if either `Topology` feature is disabled or `Immediate` volume binding mode is used.
 
 * `--kubeconfig <path>`: Path to Kubernetes client configuration that the external-provisioner uses to connect to Kubernetes API server. When omitted, default token provided by Kubernetes will be used. This option is useful only when the external-provisioner does not run as a Kubernetes pod, e.g. for debugging. Either this or `--master` needs to be set if the external-provisioner is being run out of cluster.
 
@@ -82,6 +84,17 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 * `--provisioner`: This option was used to set a provisioner name to look for in the StorageClass. Currently, the option does not have any effect and the external-provisioner uses the CSI driver name.
 
 * `--leader-election-type`: This option was used to choose which leader election resource type to use. Currently, the option defaults to `endpoints`, but will be removed in the future to only support `Lease` based leader election.
+
+### Topology support
+When `Topology` feature is enabled and the driver specifies `VOLUME_ACCESSIBILITY_CONSTRAINTS` in its plugin capabilities, external-provisioner prepares `CreateVolumeRequest.AccessibilityRequirements` while calling `Controller.CreateVolume`. The driver has to consider these topology constraints while creating the volume. Below table shows how these `AccessibilityRequirements` are prepared:
+
+[Delayed binding](https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode) | Strict topology | [Allowed topologies](https://kubernetes.io/docs/concepts/storage/storage-classes/#allowed-topologies) | [Resulting accessability requirements](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume)
+:---: |:---:|:---:|:---|
+Yes | Yes | Irrelevant | `Requisite` = `Preferred` = Selected node topology
+Yes | No  | No | `Requisite` = Aggregated cluster topology<br>`Preferred` = `Requisite` with selected node topology as first element
+Yes | No | Yes | `Requisite` = Allowed topologies<br>`Preferred` = `Requisite` with selected node topology as first element
+No | Irrelevant | No | `Requisite` = Aggregated cluster topology<br>`Preferred` = `Requisite` with randomly selected node topology as first element
+No | Irrelevant | Yes | `Requisite` = Allowed topologies<br>`Preferred` = `Requisite` with randomly selected node topology as first element
 
 ### CSI error and timeout handling
 The external-provisioner invokes all gRPC calls to CSI driver with timeout provided by `--timeout` command line argument (15 seconds by default).
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -62,6 +62,7 @@ var (
 
 	enableLeaderElection = flag.Bool("enable-leader-election", false, "Enables leader election. If leader election is enabled, additional RBAC rules are required. Please refer to the Kubernetes CSI documentation for instructions on setting up these RBAC rules.")
 	leaderElectionType   = flag.String("leader-election-type", "endpoints", "the type of leader election, options are 'endpoints' (default) or 'leases' (strongly recommended). The 'endpoints' option is deprecated in favor of 'leases'.")
+	strictTopology       = flag.Bool("strict-topology", false, "Passes only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.")
 
 	featureGates        map[string]bool
 	provisionController *controller.ProvisionController
@@ -178,7 +179,7 @@ func main() {
 
 	// Create the provisioner: it implements the Provisioner interface expected by
 	// the controller
-	csiProvisioner := ctrl.NewCSIProvisioner(clientset, *operationTimeout, identity, *volumeNamePrefix, *volumeNameUUIDLength, grpcClient, snapClient, provisionerName, pluginCapabilities, controllerCapabilities, supportsMigrationFromInTreePluginName)
+	csiProvisioner := ctrl.NewCSIProvisioner(clientset, *operationTimeout, identity, *volumeNamePrefix, *volumeNameUUIDLength, grpcClient, snapClient, provisionerName, pluginCapabilities, controllerCapabilities, supportsMigrationFromInTreePluginName, *strictTopology)
 	provisionController = controller.NewProvisionController(
 		clientset,
 		provisionerName,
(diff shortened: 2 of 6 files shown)


pkg/controller/topology_test.go
```diff
@@ -876,6 +880,23 @@ func TestTopologyAggregation(t *testing.T) {
 				{Segments: map[string]string{"com.example.csi/zone": "zone2"}},
 			},
 		},
+		"selected node; different values across cluster with strict topology": {
```
pohly: Can you add the case with no selected node and strict topology? Actually, all previous testcases should be duplicated and compared with strict topology, to ensure that they work as expected.

pkg/controller/topology.go
```diff
@@ -200,6 +211,8 @@ func GenerateAccessibilityRequirements(
 		hash, index := getPVCNameHashAndIndexOffset(pvcName)
 		i := (hash + index) % uint32(len(requisiteTerms))
 		preferredTerms = sortAndShift(requisiteTerms, nil, i)
+	} else if strictTopology {
+		preferredTerms = requisiteTerms
```
pohly: This entire "else" close is for the "Delayed binding" case, right? Better move the comment up or add a separate comment, because now it's at first glance not obvious which case is handled here.

pkg/controller/topology_test.go
avalluri: I made changes to tests, such that now every topology test runs with and without strict topology.

pkg/controller/topology.go
avalluri: yes, I agree. I made the suggested change.

avalluri:
> Can you add a README.md section which explains the topology support and add something for the new mode there?

@pohly I tried adding a section to ReadMe that explains how `AccessibilityRequirements` are prepared. Can you please have a look if it is good enough.

README.md
```diff
@@ -64,6 +64,8 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 #### Other recognized arguments
 * `--feature-gates <gates>`: A set of comma separated `<feature-name>=<true|false>` pairs that describe feature gates for alpha/experimental features. See [list of features](#feature-status) or `--help` output for list of recognized features. Example: `--feature-gates Topology=true` to enable Topology feature that's disabled by default.
 
+* `--strict-topology`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of delayed binding. Use this option to pass only selected node topology to `AccessibilityRequirements` instead of aggregated cluster topology. This option has no effect either `Topology` feature is disabled or `Immediate` volume binding mode is used.
```
pohly: Instead of the "Use this option..." sentence I would just say:
```
See [the table below](#topology-support) for an explanation how this option changes the result.
```

That's because this shorthand description doesn't fully explain the effect and trying to make it complete could become lengthy. The table is much nicer.

Missing "if" in "no effect *if* either".

README.md
```diff
@@ -83,6 +85,17 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--leader-election-type`: This option was used to choose which leader election resource type to use. Currently, the option defaults to `endpoints`, but will be removed in the future to only support `Lease` based leader election.
 
+### Topology support
+When `Topology` feature is enabled and the driver specifies `VOLUME_ACCESSIBILITY_CONSTRAINTS` in its plugin capabilities, external-provisioner prepares `CreateVolumeRequest.AccessibilityRequirements` while calling `Controller.CreateVolume`. The driver has to consider these topology constraints while creating the volume. Below table shows how these `AccessibilityRequirements` are prepared:
+
+Delayed binding | Strict Topology | AllowedTopologies | AccessabilityRequirements
+:---: |:---:|:---:|:---|
+Yes | Yes | Yes/No | `Requisite` = `Preferred` = Selected node topology
```
pohly: What is "AllowedTopologies"? To be on the safe side and ensure that all readers will understand the table, better add a short explanation or link for each column.

The spelling of the column headers is inconsistent (First word upcase, title case, camel case). Not sure which one is best, but better pick just one.

"Yes/No" confused me at first. Perhaps replace it with "ignored", "irrelevant" or "*" (wildcard)? When using "*", a short comment under the table would be useful.

Other than that the table is very useful, thanks.

README.md
avalluri: Made suggested changes.

README.md
avalluri: > What is "AllowedTopologies"? To be on the safe side and ensure that all readers will understand the table, better add a short explanation or link for each column.
> 
Added links to column headers.
> The spelling of the column headers is inconsistent (First word upcase, title case, camel case). Not sure which one is best, but better pick just one.
> 
Fixed to 'First word upper case'.

> "Yes/No" confused me at first. Perhaps replace it with "ignored", "irrelevant" or "_" (wildcard)? When using "_", a short comment under the table would be useful.
> 
Used "Irrelevant" in place of "Yes/No".

README.md
```diff
@@ -83,6 +85,17 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--leader-election-type`: This option was used to choose which leader election resource type to use. Currently, the option defaults to `endpoints`, but will be removed in the future to only support `Lease` based leader election.
 
+### Topology support
+When `Topology` feature is enabled and the driver specifies `VOLUME_ACCESSIBILITY_CONSTRAINTS` in its plugin capabilities, external-provisioner prepares `CreateVolumeRequest.AccessibilityRequirements` while calling `Controller.CreateVolume`. The driver has to consider these topology constraints while creating the volume. Below table shows how these `AccessibilityRequirements` are prepared:
+
+[Delayed binding](https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode) | Strict topology | [Allowed topologies](https://kubernetes.io/docs/concepts/storage/storage-classes/#allowed-topologies) | [Resulted accessability requirements](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume)
```
pohly: nit: s/Resulted accessability requirements/Resulting accessability requirements/

avalluri: Done, Thanks.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -62,6 +62,7 @@ var (
 
 	enableLeaderElection = flag.Bool("enable-leader-election", false, "Enables leader election. If leader election is enabled, additional RBAC rules are required. Please refer to the Kubernetes CSI documentation for instructions on setting up these RBAC rules.")
 	leaderElectionType   = flag.String("leader-election-type", "endpoints", "the type of leader election, options are 'endpoints' (default) or 'leases' (strongly recommended). The 'endpoints' option is deprecated in favor of 'leases'.")
+	strictTopology       = flag.Bool("strict-topology", false, "Passes only selected node topology to CreateVolume Request, unlike default behavior of passing all nodes that match with topology keys of the selected node.")
```
msau42: To match the wording in the README:

"passing all nodes" => "passing aggregated cluster topologies"

pkg/controller/topology.go
```diff
@@ -158,20 +160,58 @@ func GenerateAccessibilityRequirements(
...
 	// 2. Generate CSI Requisite Terms
 	var requisiteTerms []topologyTerm
-	if len(allowedTopologies) == 0 {
-		// Aggregate existing topologies in nodes across the entire cluster.
-		var err error
-		requisiteTerms, err = aggregateTopologies(kubeClient, driverName, selectedCSINode)
-		if err != nil {
-			return nil, err
+	if selectedCSINode != nil && strictTopology {
+		// Make sure that selected node topology is in allowed topologies list
```
msau42: We could probably be more efficient and just assume Kubernetes does the right thing.

pkg/controller/topology.go
```diff
@@ -158,20 +160,58 @@ func GenerateAccessibilityRequirements(
...
 
 	// 2. Generate CSI Requisite Terms
 	var requisiteTerms []topologyTerm
-	if len(allowedTopologies) == 0 {
-		// Aggregate existing topologies in nodes across the entire cluster.
-		var err error
-		requisiteTerms, err = aggregateTopologies(kubeClient, driverName, selectedCSINode)
-		if err != nil {
-			return nil, err
+	if selectedCSINode != nil && strictTopology {
```
msau42: Can this be changed to a switch statement with the other 2 conditions, since all 3 are mutually exclusive from each other?

pkg/controller/topology.go
avalluri: I agree, but there is a test "`topology from selected node is not in allowedTopologies`" for this, so I added this check to satisfy the test.

pkg/controller/topology.go
avalluri: Those are not mutually exclusive. It's possible that both `allowedTopologies` and `selectedNode` set, and resulted topology depends on `strictTopology` value.

I could move this block to inside above `if selectedNode != nil {..}`

PR #283, opened 2019-05-23
Need to check VolumeContentSource if creating volume from snapshot

### LINKED ISSUE #280: Need to check VolumeContentSource if creating volume from snapshot
In Provision, we should check if VolumeContentSource is populated if creating volume from snapshot is successful and log event to indicate its status.

Note: This needs to be fixed in the CSI hostpath driver first.  Otherwise, CI won't pass.
https://github.com/kubernetes-csi/csi-driver-host-path/issues/50

zhucan:
**What type of PR is this?**
/kind bug

Need to check VolumeContentSource if creating volume from snapshot

Fixes #280 

```release-note
Checks if volume content source is populated if creating a volume from a snapshot source.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -114,6 +114,8 @@ const (
 	tokenPVNameKey       = "pv.name"
 	tokenPVCNameKey      = "pvc.name"
 	tokenPVCNameSpaceKey = "pvc.namespace"
+
+	deleteVolumeRetryCount = 5
 )
 
 var (
@@ -550,15 +552,27 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 		delReq := &csi.DeleteVolumeRequest{
 			VolumeId: rep.GetVolume().GetVolumeId(),
 		}
-		delReq.Secrets = provisionerCredentials
-		ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
-		defer cancel()
-		_, err := p.csiClient.DeleteVolume(ctx, delReq)
+		err = cleanupVolume(p, delReq, provisionerCredentials)
 		if err != nil {
 			capErr = fmt.Errorf("%v. Cleanup of volume %s failed, volume is orphaned: %v", capErr, pvName, err)
 		}
 		// use InBackground to retry the call, hoping the volume is deleted correctly next time.
 		return nil, controller.ProvisioningInBackground, capErr
+	}
+
+	if options.PVC.Spec.DataSource != nil {
+		contentSource := rep.GetVolume().ContentSource
+		if contentSource == nil {
+			sourceErr := fmt.Errorf("volume content source missing")
+			delReq := &csi.DeleteVolumeRequest{
+				VolumeId: rep.GetVolume().GetVolumeId(),
+			}
+			err = cleanupVolume(p, delReq, provisionerCredentials)
+			if err != nil {
+				sourceErr = fmt.Errorf("%v. cleanup of volume %s failed, volume is orphaned: %v", sourceErr, pvName, err)
+			}
+			return nil, controller.ProvisioningInBackground, sourceErr
+		}
 	}
 
 	pv := &v1.PersistentVolume{
@@ -1064,4 +1078,18 @@ func isFinalError(err error) bool {
 	// All other errors mean that provisioning either did not
 	// even start or failed. It is for sure not in progress.
 	return true
+}
+
+func cleanupVolume(p *csiProvisioner, delReq *csi.DeleteVolumeRequest, provisionerCredentials map[string]string) error {
+	var err error = nil
+	delReq.Secrets = provisionerCredentials
+	ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+	defer cancel()
+	for i := 0; i < deleteVolumeRetryCount; i++ {
+		_, err = p.csiClient.DeleteVolume(ctx, delReq)
+		if err == nil {
+			break
+		}
+	}
+	return err
 }
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -1603,6 +1603,7 @@ func TestProvisionFromSnapshot(t *testing.T) {
 		snapshotStatusReady  bool
 		expectedPVSpec       *pvSpec
 		expectErr            bool
+		notPopulated         bool
 	}{
 		"provision with volume snapshot data source": {
 			volOpts: controller.ProvisionOptions{
@@ -1795,6 +1796,36 @@ func TestProvisionFromSnapshot(t *testing.T) {
 			snapshotStatusReady: false,
 			expectErr:           true,
 		},
+		"fail not populated volume content source": {
+			volOpts: controller.ProvisionOptions{
+				StorageClass: &storagev1.StorageClass{
+					Parameters: map[string]string{},
+				},
+				PVName: "test-name",
+				PVC: &v1.PersistentVolumeClaim{
+					ObjectMeta: metav1.ObjectMeta{
+						UID: "testid",
+					},
+					Spec: v1.PersistentVolumeClaimSpec{
+						Selector: nil,
+						Resources: v1.ResourceRequirements{
+							Requests: v1.ResourceList{
+								v1.ResourceName(v1.ResourceStorage): resource.MustParse(strconv.FormatInt(requestedBytes, 10)),
+							},
+						},
+						AccessModes: []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce},
+						DataSource: &v1.TypedLocalObjectReference{
+							Name:     snapName,
+							Kind:     "VolumeSnapshot",
+							APIGroup: &apiGrp,
+						},
+					},
+				},
+			},
+			snapshotStatusReady: true,
+			expectErr:           true,
+			notPopulated:        true,
+		},
 	}
 
 	tmpdir := tempDir(t)
@@ -1838,7 +1869,23 @@ func TestProvisionFromSnapshot(t *testing.T) {
 		// When the following if condition is met, it is a valid create volume from snapshot
 		// operation and CreateVolume is expected to be called.
 		if tc.restoredVolSizeSmall == false && tc.wrongDataSource == false && tc.snapshotStatusReady {
-			controllerServer.EXPECT().CreateVolume(gomock.Any(), gomock.Any()).Return(out, nil).Times(1)
+			if tc.notPopulated {
+				out.Volume.ContentSource = nil
+				controllerServer.EXPECT().CreateVolume(gomock.Any(), gomock.Any()).Return(out, nil).Times(1)
+				controllerServer.EXPECT().DeleteVolume(gomock.Any(), &csi.DeleteVolumeRequest{
+					VolumeId: "test-volume-id",
+				}).Return(&csi.DeleteVolumeResponse{}, nil).Times(1)
+			} else {
+				snapshotSource := csi.VolumeContentSource_Snapshot{
+					Snapshot: &csi.VolumeContentSource_SnapshotSource{
+						SnapshotId: "sid",
+					},
+				}
+				out.Volume.ContentSource = &csi.VolumeContentSource{
+					Type: &snapshotSource,
+				}
+				controllerServer.EXPECT().CreateVolume(gomock.Any(), gomock.Any()).Return(out, nil).Times(1)
+			}
 		}
 
 		pv, err := csiProvisioner.Provision(tc.volOpts)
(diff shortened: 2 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -509,6 +509,26 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
 		return nil, capErr
 	}
 
+	if needSnapshotSupport {
+		contentSource := rep.GetVolume().ContentSource
+		if contentSource == nil {
+			sourceErr := fmt.Errorf("created volume contentsource missing")
```
xing-yang: volume content source missing

pkg/controller/controller.go
```diff
@@ -509,6 +509,26 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
...
+			ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+			defer cancel()
+			_, err := p.csiClient.DeleteVolume(ctx, delReq)
+			if err != nil {
+				sourceErr = fmt.Errorf("%v. Cleanup of volume %s failed, volume is orphaned: %v", sourceErr, pvName, err)
+			}
+			return nil, sourceErr
+		}
+
+		klog.Infof("created volume %s from snapshot %s is successful", pvName, contentSource.GetSnapshot().SnapshotId)
```
xing-yang: s/is successful/successfully

xing-yang:
Please add a release note:

```release-note
Checks if volume content source is populated if creating a volume from a snapshot source.
```

pkg/controller/controller.go
```diff
@@ -509,6 +509,26 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
...
+			sourceErr := fmt.Errorf("created volume contentsource missing")
+			delReq := &csi.DeleteVolumeRequest{
+				VolumeId: rep.GetVolume().GetVolumeId(),
+			}
+			delReq.Secrets = provisionerCredentials
+			ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+			defer cancel()
+			_, err := p.csiClient.DeleteVolume(ctx, delReq)
+			if err != nil {
+				sourceErr = fmt.Errorf("%v. Cleanup of volume %s failed, volume is orphaned: %v", sourceErr, pvName, err)
```
j-griffith: Adding in the original sourceErr from above, I was worried that might get masked out but you covered it.

humblec: may be use small letter for "Cleanup of volume.."

xing-yang:
@zhucan There is an error in the unit test.  Please take a look.  Here are the errors in the unit test log:

--- FAIL: TestProvisionFromSnapshot (5.00s)
    driver.mock.go:192: Unexpected call to *driver.MockControllerServer.DeleteVolume([context.Background.WithCancel.WithDeadline(2019-06-12 17:17:11.587723947 +0000 UTC m=+5.077308410 [4.999929329s]).WithValue(peer.peerKey{}, &peer.Peer{Addr:(*net.UnixAddr)(0xc000375260), AuthInfo:credentials.AuthInfo(nil)}).WithValue(metadata.mdIncomingKey{}, metadata.MD{":authority":[]string{"unix:///tmp/external-attacher-test-196419863/csi.sock"}, "content-type":[]string{"application/grpc"}, "user-agent":[]string{"grpc-go/1.19.1"}}).WithValue(grpc.streamKey{}, <stream: 0xc000506600, /csi.v1.Controller/DeleteVolume>) volume_id:"test-volume-id" ]) at /home/prow/go/src/github.com/kubernetes-csi/external-provisioner/vendor/github.com/kubernetes-csi/csi-test/driver/driver.mock.go:192 because: 
        Expected call at /home/prow/go/src/github.com/kubernetes-csi/external-provisioner/pkg/controller/controller_test.go:1703 has already been called the max number of times.
    controller_test.go:1715: test "provision with volume snapshot data source": got error: volume content source missing. cleanup of volume test-testi failed, volume is orphaned: rpc error: code = DeadlineExceeded desc = context deadline exceeded
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x12322e8]

zhucan:
> @zhucan There is an error in the unit test "TestProvisionFromSnapshot". Please take a look. Here are the errors in the unit test log:
> 
> ```
> --- FAIL: TestProvisionFromSnapshot (5.00s)
>     driver.mock.go:192: Unexpected call to *driver.MockControllerServer.DeleteVolume([context.Background.WithCancel.WithDeadline(2019-06-12 17:17:11.587723947 +0000 UTC m=+5.077308410 [4.999929329s]).WithValue(peer.peerKey{}, &peer.Peer{Addr:(*net.UnixAddr)(0xc000375260), AuthInfo:credentials.AuthInfo(nil)}).WithValue(metadata.mdIncomingKey{}, metadata.MD{":authority":[]string{"unix:///tmp/external-attacher-test-196419863/csi.sock"}, "content-type":[]string{"application/grpc"}, "user-agent":[]string{"grpc-go/1.19.1"}}).WithValue(grpc.streamKey{}, <stream: 0xc000506600, /csi.v1.Controller/DeleteVolume>) volume_id:"test-volume-id" ]) at /home/prow/go/src/github.com/kubernetes-csi/external-provisioner/vendor/github.com/kubernetes-csi/csi-test/driver/driver.mock.go:192 because: 
>         Expected call at /home/prow/go/src/github.com/kubernetes-csi/external-provisioner/pkg/controller/controller_test.go:1703 has already been called the max number of times.
>     controller_test.go:1715: test "provision with volume snapshot data source": got error: volume content source missing. cleanup of volume test-testi failed, volume is orphaned: rpc error: code = DeadlineExceeded desc = context deadline exceeded
> panic: runtime error: invalid memory address or nil pointer dereference [recovered]
> 	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
> 	panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x12322e8]
> ```

because the normal unit test

xing-yang:
@zhucan Please check the unit test failure in pull-kubernetes-csi-external-provisioner-unit.  Thanks.

zhucan:
> @zhucan Please check the unit test failure in pull-kubernetes-csi-external-provisioner-unit. Thanks.

ok

pkg/controller/controller.go
```diff
@@ -532,6 +532,26 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
...
+			ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+			defer cancel()
+			_, err := p.csiClient.DeleteVolume(ctx, delReq)
+			if err != nil {
+				sourceErr = fmt.Errorf("%v. cleanup of volume %s failed, volume is orphaned: %v", sourceErr, pvName, err)
+			}
+			return nil, sourceErr
+		}
+
+		klog.Infof("created volume %s from snapshot %s successfully", pvName, contentSource.GetSnapshot().SnapshotId)
```
msau42: This can be clone too, not just snapshot right?

pkg/controller/controller.go
```diff
@@ -532,6 +532,26 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
 		return nil, capErr
 	}
 
+	if options.PVC.Spec.DataSource != nil && (rc.clone || rc.snapshot) {
+		contentSource := rep.GetVolume().ContentSource
```
msau42: Do we store contentSource anywhere in the PV object?

pkg/controller/controller_test.go
```diff
@@ -1503,6 +1504,8 @@ func TestProvisionFromSnapshot(t *testing.T) {
 				},
 			},
 			snapshotStatusReady: true,
+			notPopulated:        true,
+			expectErr:           true,
```
msau42: Do we have a success test case?

pkg/controller/controller.go
zhucan: Only DataSource is not nil, and the pvc create from snapshot or clone with other pvc

pkg/controller/controller_test.go
zhucan: because 

> Do we have a success test case?

In default, the volume content source not be populated, why?

pkg/controller/controller.go
xing-yang: Yes, it can be clone too.

pkg/controller/controller_test.go
xing-yang: By default the volume content source will not be populated because it will be an empty volume.

xing-yang: This should be a success test case.  Volume content source should be populated.  This seems to be wrong.

zhucan: In the success test case, Actually  the volume content source not be populated, why?

msau42: The mock server used for unit testing needs to be updated. See https://github.com/kubernetes-csi/external-provisioner/blob/a297985bed50f97901b869558ceb099b286ea544/pkg/controller/controller_test.go#L416

pkg/controller/controller_test.go
```diff
@@ -1819,17 +1866,19 @@ func TestProvisionFromSnapshot(t *testing.T) {
 		}
 
 		if tc.expectedPVSpec != nil {
-			if pv.Name != tc.expectedPVSpec.Name {
-				t.Errorf("test %q: expected PV name: %q, got: %q", k, tc.expectedPVSpec.Name, pv.Name)
-			}
+			if pv != nil {
+				if pv.Name != tc.expectedPVSpec.Name {
```
msau42: Ideally, we should be comparing the whole PV object, not just the PVSource, and we should just do a deep equal on the entire object, not just individual fields.  That way, we won't miss validating new fields.

zhucan: ReclaimPolicy and AccessModes must to be compared？

msau42: Actually is volume content supposed to be populated in PV object? That's what I had in mind.

If not then nevermind, this is more of a test cleanup issue

zhucan: I agree it that volume content should be populated in pv object, actually the volume content not to be populated, so I hadn't compare it

msau42: @xing-yang @jingxu97 can you clarify if volumecontent is supposed to be populated in PV object, or only PVC object?

xing-yang: @msau42 VolumeContentSource is not populated in PV object.  Only PVC has a DataSource.  We should probably take another look of this during API review.

msau42: Thanks for clarifying. I think this PR is ready to go then

msau42:
/approve

Can you squash your commits?

zhucan:
@msau42 I have squashed it

pkg/controller/controller.go
```diff
@@ -561,6 +561,24 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
...
+			sourceErr := fmt.Errorf("volume content source missing")
+			delReq := &csi.DeleteVolumeRequest{
+				VolumeId: rep.GetVolume().GetVolumeId(),
+			}
+			delReq.Secrets = provisionerCredentials
+			ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+			defer cancel()
+			_, err := p.csiClient.DeleteVolume(ctx, delReq)
+			if err != nil {
+				sourceErr = fmt.Errorf("%v. cleanup of volume %s failed, volume is orphaned: %v", sourceErr, pvName, err)
```
jingxu97: do we need to try to delete volume a few times if it fails? Another option is still create PV object, but indicate the source is missing. This will require API change and need more discussion.

msau42: What does provisioner currently do?

xing-yang: Currently the provisioner silently creates an empty volume if the content source is missing because there is no check.  That's why this PR tries to add a check there.

I think it is not useful to create an empty volume when user wants to create a volume from snapshot.  Typically user wants to do this because the original volume is corrupted so they want to restore the volume.  An empty volume won't solve their problems.  The reason that the content source is missing is most likely because the driver didn't implement "create volume from snapshot" so it just goes through "create a new volume" code path.

jingxu97: The concern here is how to deal with the orphaned volumes

xing-yang: Similar delete volume logic is on line 556 in this routine.  We should handle orphaned volumes consistently.  Retries may help.

msau42: Let's retry Delete a few times. That's what pv controller does

xing-yang: @zhucan can you add retries for volume clean up here?  If delete volume fails, retry a few times.

zhucan: copy that, I will do it @xing-yang  @msau42

pkg/controller/controller.go
```diff
@@ -561,6 +563,31 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
...
+			}
+			delReq.Secrets = provisionerCredentials
+			ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+			defer cancel()
+			for i := 0; i < deleteVolumeRetryCount; i++ {
+				_, err = p.csiClient.DeleteVolume(ctx, delReq)
+				if err == nil {
+					break
+				}
+				klog.V(3).Infof("%v. cleanup of volume %s failed", sourceErr, pvName)
```
msau42: Errorf instead?

pkg/controller/controller.go
```diff
@@ -561,6 +563,31 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
...
+			}
+			if err != nil {
+				sourceErr = fmt.Errorf("%v. cleanup of volume %s failed, volume is orphaned: %v", sourceErr, pvName, err)
+			}
+			klog.Error(sourceErr)
```
msau42: I don't think we need to log here, it should be logged by the caller

zhucan: @msau42 Thanks for your comments

pkg/controller/controller.go
zhucan: copy that

pkg/controller/controller.go
```diff
@@ -561,6 +563,30 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
...
+		contentSource := rep.GetVolume().ContentSource
+		if contentSource == nil {
+			sourceErr := fmt.Errorf("volume content source missing")
+			delReq := &csi.DeleteVolumeRequest{
+				VolumeId: rep.GetVolume().GetVolumeId(),
+			}
+			delReq.Secrets = provisionerCredentials
+			ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+			defer cancel()
+			for i := 0; i < deleteVolumeRetryCount; i++ {
```
msau42: @xing-yang pointed out that the above Deletion method should also retry too. Can you put this into a common method and then have it called in both places?

zhucan: @msau42 I will do it

pkg/controller/controller.go
```diff
@@ -550,17 +552,29 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
...
-		_, err := p.csiClient.DeleteVolume(ctx, delReq)
+		err = cleanupVolume(p, delReq, provisionerCredentials)
 		if err != nil {
 			capErr = fmt.Errorf("%v. Cleanup of volume %s failed, volume is orphaned: %v", capErr, pvName, err)
 		}
 		// use InBackground to retry the call, hoping the volume is deleted correctly next time.
 		return nil, controller.ProvisioningInBackground, capErr
 	}
 
+	if options.PVC.Spec.DataSource != nil && (rc.clone || rc.snapshot) {
```
msau42: Do we need to check `rc`? It should have already errored earlier right?

zhucan: @msau42 we needn't  check rc in here,  we have checked it before.

msau42: can you remove the check?

zhucan: @msau42  copy that, thanks for your comment.

msau42:
/lgtm
Thanks!

PR #288, opened 2019-06-06
[release-1.2]  Introduce new flag - strict-topology

### LINKED ISSUE #221: Wrong AccessibilityRequirement passed in CreateVolumeRequest

avalluri:
Backport #282 to release-1.2

**What type of PR is this?**
/kind bug
/kind design

Allows the driver to support delayed binding properly.

Fixes #221 

```release-note
support strict topology for volumes with delayed binding
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 ### Command line options
 
-#### Recommended optional arguments"
+#### Recommended optional arguments
 * `--csi-address <path to CSI socket>`: This is the path to the CSI driver socket inside the pod that the external-provisioner container will use to issue CSI operations (`/run/csi/socket` is used by default).
 
 * `--enable-leader-election`: Enables leader election. This is mandatory when there are multiple replicas of the same external-provisioner running for one CSI driver. Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.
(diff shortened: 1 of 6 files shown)


msau42:
/lgtm
/approve
/hold

As a sanity check, given that hostpath doesn't implement topology, I want to first run the pd e2es on a multi-zone cluster to make sure this hasn't caused any regressions

msau42:
Tested manually with repd

/hold cancel

PR #301, opened 2019-06-18
Add controller expand secret ref to PV

### LINKED ISSUE #278: Update external-provisioner to set resize secrets on resizing when necessary
Since we added https://github.com/kubernetes/kubernetes/pull/77516, we need to make sure that volume resizing secrets are set in `CSIPersistentVolumeSource` whenever they are created.

cc @leakingtapan

ggriffiths:
**What type of PR is this?**
/kind feature

This PR adds in the ControllerExpandSecretRef to the PV.

Fixes #278

* I renamed `prefixedResizerSecretNameKey` to `prefixedControllerExpandSecretNameKey` . This naming matches the other secret key names (nodePublishSecretNameKey etc).
* I also changed these keys to be `controller-expand-secret-name-key` and `controller-expand-secret-namespace-key` (to match the other key naming). 
* I can change these keys back if needed (although we haven't documented them yet)
* This change also updates k8s.io/api to latest in release-1.15 branch. I did this with an `[[override]]` instead of updating all dependencies to also have the release-1.15 dep constraint.

```release-note
The resizer secret name and namespace keys have been changed from the following values:
csi.storage.k8s.io/resizer-secret-name
csi.storage.k8s.io/resizer-secret-namespace

to be the following values:
csi.storage.k8s.io/controller-expand-secret-name
csi.storage.k8s.io/controller-expand-secret-namespace

This is a breaking change and is being introduced so that these keys match the naming convention for other secret name/namespace keys.
```


msau42:
Renaming the secret is fine since resizing is still an alpha feature. Can you add a release note for this and note that it's a breaking change?

Gopkg.toml
```diff
@@ -47,6 +47,10 @@
 # because of the lack of semantic versioning in Kubernetes. Therefore we
 # have to pick a certain release, otherwise we (more or less randomly) end
 # up with something older or master.
+[[override]]
+  name = "k8s.io/api"
```
msau42: i'm surprised client-go didn't need to be updated?

ggriffiths: Yeah, I found this override block makes it possible. It's more of a temporary fix until all other dependencies update to k8s.io/api 1.15.
https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#override

msau42:
/approve

Will let @gnufied lgtm

Can you also specify in the release note what the old key names were?

PR #302, opened 2019-06-18
update readme with --leader-election-namespace flag

msau42:
**What type of PR is this?**
/kind documentation

Fixes #

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--leader-election-type`: The resource type to use for leader election, options are 'endpoints' (default) or 'leases' (recommended)
 
+* `--leader-election-namespace`: Namespace where leader election object will be created. It is recommended that this parameter is populated from Kubernetes DownwardAPI with the namespace where the external-provisioner runs in.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,8 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--leader-election-type`: The resource type to use for leader election, options are 'endpoints' (default) or 'leases' (recommended)
 
+* `--leader-election-namespace`: Namespace where the external-provisioner runs and where leader election object will be created. It is recommended that this parameter is populated from Kubernetes DownwardAPI.
```
verult: Technically external-provisioner doesn't *have* to run in the same namespace as the lease object, for example if the provisioner uses an User account that has cluster-wide scope. Definitely recommended though

msau42: Yeah I would just keep the documentation simple and not try to encourage confusing practices.

verult: What about this:

`Namespace where leader election object will be created. It is recommended that this parameter is populated from Kubernetes DownwardAPI from the namespace field of the pod in which the external-provisioner runs.`

The original wording could potentially be interpreted as: if I want the external provisioner to run in some namespace, I have to set this field in order for the provisioner to work correctly in that namespace (even if leader election isn't enabled)

msau42: Updated. I worded it slightly differently from your suggestion. Hope it captures the same idea.

PR #305, opened 2019-06-20
Add 1.3 changelog

msau42:
**What type of PR is this?**
/kind documentation

Fixes #

```release-note
NONE
```

### DIFF
--- a/CHANGELOG-1.3.md
+++ b/CHANGELOG-1.3.md
@@ -1,0 +1,27 @@
+# Changelog since v1.2.0
+
+## Breaking Changes
+
+- The alpha resizer secret name and namespace keys have been changed from the following values:
+    - `csi.storage.k8s.io/resizer-secret-name`
+    - `csi.storage.k8s.io/resizer-secret-namespace`
+
+  to be the following values:
+    - `csi.storage.k8s.io/controller-expand-secret-name`
+    - `csi.storage.k8s.io/controller-expand-secret-namespace`
+
+  This is a breaking change and is being introduced so that these keys match the naming convention for other secret name/namespace keys. ([#301](https://github.com/kubernetes-csi/external-provisioner/pull/301), [@ggriffiths](https://github.com/ggriffiths))
+
+## New Features
+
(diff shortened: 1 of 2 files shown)


CHANGELOG-1.3.md
```diff
@@ -0,0 +1,27 @@
+# Changelog since v1.2.0
```
msau42: Do we want to compare against 1.2.0 (which was prematurely released), or 1.2.1?

jsafrane: IMO we should follow what Kubernetes does and its 1.14 changelog starts with ` Changelog since v1.13.0`

README.md
```diff
@@ -11,24 +11,22 @@ The external-provisioner is an external controller that monitors `PersistentVolu
...
 
-| Feature        | Status  | Default | Description                                                                                   |
-| -------------- | ------- | ------- | --------------------------------------------------------------------------------------------- |
-| Topology       | Beta    | Off     | [Topology aware dynamic provisioning](https://kubernetes-csi.github.io/docs/topology.html)  (requires kubelet 1.14 on nodes). |
-| Snapshots*     | Alpha   | On      | [Snapshots and Restore](https://kubernetes-csi.github.io/docs/snapshot-restore-feature.html). |
-| CSIMigration** | Alpha   | On      | [Migrating in-tree volume plugins to CSI](https://kubernetes.io/docs/concepts/storage/volumes/#csi-migration). |
-
-*) There is no special feature gate for snapshots. It is enabled by running [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter/) and creating its CRDs.
-**) There is no special feature gate for CSI migration. It is enabled by turning on CSI migration in Kubernetes.
+| Feature        | Status  | Default | Description                                                                                   | Provisioner Feature Gate Required |
```
jsafrane: Do we have a feature for resize? Or why are resize keys mentioned in this changelog?

msau42: The note about resize is referring to taking the storageclass resize secrets parameter and setting it in the PV object when provisioning.  I don't think that's worth calling out as a feature as we do it for all other operation secrets

jsafrane: ack

PR #309, opened 2019-07-01
fix pvc and source PVC storageclass name comparison for pvc datasource

### LINKED ISSUE #308: cannot create  PVC with datasource=PersistentVolumeClaim when clone PVC and source PVC storageclass name is same
logs:

`
/rbd-pvc-restore"
I0701 12:39:00.779025       1 controller.go:979] Final error received, removing PVC 1e98da4d-49d2-41b2-9c16-b9a7784caa38 from claims in progress
W0701 12:39:00.779060       1 controller.go:886] Retrying syncing claim "1e98da4d-49d2-41b2-9c16-b9a7784caa38", failure 3
E0701 12:39:00.779093       1 controller.go:908] error syncing claim "1e98da4d-49d2-41b2-9c16-b9a7784caa38": failed to provision volume with StorageClass "csi-rbd-sc": error getting handle for DataSource Type PersistentVolumeClaim by Name rbd-pvc: the source PVC and destination PVCs must be in the same storage class for cloning.  Source is in 0xc00049a470, but new PVC is in 0xc000516bb0
`

CC @j-griffith

Madhu-1:
**What type of PR is this?**

This fixes  the  issue in storageclass name comparison for PVC  creation with datasource=PersistanceVolumeClaim
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #308

```release-note
None
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -661,9 +661,18 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
 	if sourcePVC.ObjectMeta.DeletionTimestamp != nil {
 		return nil, fmt.Errorf("the PVC DataSource %s is currently being deleted", options.PVC.Spec.DataSource.Name)
 	}
-	if sourcePVC.Spec.StorageClassName != options.PVC.Spec.StorageClassName {
+
+	if sourcePVC.Spec.StorageClassName == nil {
+		return nil, fmt.Errorf("the source PVC (%s) storageclass cannot be empty", sourcePVC.Name)
+	}
+
+	if options.PVC.Spec.StorageClassName == nil {
+		return nil, fmt.Errorf("the requested PVC (%s) storageclass cannot be empty", options.PVC.Name)
+	}
+
+	if *sourcePVC.Spec.StorageClassName != *options.PVC.Spec.StorageClassName {
 		return nil, fmt.Errorf("the source PVC and destination PVCs must be in the same storage class for cloning.  Source is in %v, but new PVC is in %v",
-			sourcePVC.Spec.StorageClassName, options.PVC.Spec.StorageClassName)
+			*sourcePVC.Spec.StorageClassName, *options.PVC.Spec.StorageClassName)
 	}
 	capacity := options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
 	requestedSize := capacity.Value()
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -2144,8 +2145,10 @@ func runDeleteTest(t *testing.T, k string, tc deleteTestcase) {
 // TestProvisionFromPVC tests create volume clone
 func TestProvisionFromPVC(t *testing.T) {
 	var requestedBytes int64 = 1000
-	invalidSCName := "invalid-sc"
+	fakeSc1 := "fake-sc-1"
+	fakeSc2 := "fake-sc-2"
 	srcName := "fake-pvc"
+	invalidPVC := "invalid-pv"
 	pvName := "test-testi"
 	deletePolicy := v1.PersistentVolumeReclaimDelete
 
@@ -2160,6 +2163,7 @@ func TestProvisionFromPVC(t *testing.T) {
 	testcases := map[string]struct {
 		volOpts              controller.ProvisionOptions
 		restoredVolSizeSmall bool
+		restoredVolSizeBig   bool
 		wrongDataSource      bool
 		pvcStatusReady       bool
 		expectedPVSpec       *pvSpec
@@ -2178,7 +2182,8 @@ func TestProvisionFromPVC(t *testing.T) {
 						UID: "testid",
 					},
 					Spec: v1.PersistentVolumeClaimSpec{
-						Selector: nil,
+						StorageClassName: &fakeSc1,
+						Selector:         nil,
 						Resources: v1.ResourceRequirements{
 							Requests: v1.ResourceList{
 								v1.ResourceName(v1.ResourceStorage): resource.MustParse(strconv.FormatInt(requestedBytes, 10)),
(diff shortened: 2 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -661,9 +661,9 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
 	if sourcePVC.ObjectMeta.DeletionTimestamp != nil {
 		return nil, fmt.Errorf("the PVC DataSource %s is currently being deleted", options.PVC.Spec.DataSource.Name)
 	}
-	if sourcePVC.Spec.StorageClassName != options.PVC.Spec.StorageClassName {
+	if *sourcePVC.Spec.StorageClassName != *options.PVC.Spec.StorageClassName {
```
msau42: Could this be nil?

Also can we add a unit test for this?

Madhu-1: @msau42 yes this can be nil also. updated code and added unit testing.

msau42:
@j-griffith actually does storageclass need to be the same for cloning?  From a user's perspective, you at least want the content to be cloned, but do you require that all the underlying storage properties are exactly the same (ie hdd vs sdd?)

ripulpatel:
@msau42 ,

Re

> @j-griffith actually does storageclass need to be the same for cloning? From a user's perspective, you at least want the content to be cloned, but do you require that all the underlying storage properties are exactly the same (ie hdd vs sdd?)

We should not require properties exactly the same as dev-test clone could very well use cheaper media like hdd over it's original source's media which could be on faster media ssd.

pkg/controller/controller.go
j-griffith: DataSource and rquested provisioned PVC are *required* to be in the same storage class.  Also, I'm not sure I follow, how can the sourcePVC have a "nil" storage class?  Once it's provisioned it will have a storage class, even if one wasn't specified at creation, the default class would be used and the object would have a storage class assigned to it no?

j-griffith:
> @msau42 ,
> 
> Re
> 
> > @j-griffith actually does storageclass need to be the same for cloning? From a user's perspective, you at least want the content to be cloned, but do you require that all the underlying storage properties are exactly the same (ie hdd vs sdd?)
> 
> We should not require properties exactly the same as dev-test clone could very well use cheaper media like hdd over it's original source's media which could be on faster media ssd.

@ripulpatel As per the proposal we do in fact require the same storage class.  We did this intentionally for alpha in particular.  There are a number of things that need to be solved to do cross SC cloning to deal with identifying the use of the same plugin (certainly possible but we intentionally did not do that for the introduction of the feature).

pkg/controller/controller.go
Madhu-1: @j-griffith this is a kind of safety check, will it cause any issue if we do `nil` check?

msau42:
> There are a number of things that need to be solved to do cross SC cloning to deal with identifying the use of the same plugin

Is that the only issue? I think it can be solved by comparing datasource's pv's driver name with storageclass provisioner name (which must be equal)

j-griffith:
> > There are a number of things that need to be solved to do cross SC cloning to deal with identifying the use of the same plugin
> 
> Is that the only issue? I think it can be solved by comparing datasource's pv's driver name with storageclass provisioner name (which must be equal)

@msau42 I also have some concerns because we're using the SC for things like defining the fstype.  So for example `SC-1.fstype=xfs` and `SC-2.fstype=ext4`, even if theyr'e the same backend you can't "clone" across them.  We don't have a way to express that capability.  This also applies to plugins that use things like SCs to define block vs filesystem mode. Basically there are a lot of little "corner" cases like this that IMO make cross SC cloning a bit risky.

ripulpatel:
> > > There are a number of things that need to be solved to do cross SC cloning to deal with identifying the use of the same plugin
> > 
> > 
> > Is that the only issue? I think it can be solved by comparing datasource's pv's driver name with storageclass provisioner name (which must be equal)
> 
> @msau42 I also have some concerns because we're using the SC for things like defining the fstype. So for example `SC-1.fstype=xfs` and `SC-2.fstype=ext4`, even if theyr'e the same backend you can't "clone" across them. We don't have a way to express that capability. This also applies to plugins that use things like SCs to define block vs filesystem mode. Basically there are a lot of little "corner" cases like this that IMO make cross SC cloning a bit risky.

@j-griffith ,
I agree these are 'corner' cases but aren't these checks be done by plugins (storage providers) instead? Plugins can better know capabilities of its clone so that it can judge on what properties are derived from parent and what properties are modifiable in the clone? For provisioner all that information can be opaque. IMO provisioner should only have check if plugins for Parent/Source and clone is same or not.

pkg/controller/controller.go
j-griffith: It won't hurt at all, and might as well err on the side of caution.

j-griffith:
@Madhu-1 Ok, now I'm catching up :)  If you could rebase I'm happy to approve/lgtm, thanks for fixing this up.

humblec:
@Madhu-1 please fix the PR title and also rebase.

Madhu-1:
@j-griffith I have updated some of the test cases PTAL.

pkg/controller/controller_test.go
```diff
@@ -2160,6 +2163,7 @@ func TestProvisionFromPVC(t *testing.T) {
 	testcases := map[string]struct {
 		volOpts              controller.ProvisionOptions
 		restoredVolSizeSmall bool
+		restoredVolSizeBig   bool
```
j-griffith: We should probably add a description column to the test table so we don't have to infer it ourselves or guess what was intended.  I'll leave it up to you if you want to add that or I can submit a follow up PR to include it.  Just let me know.

Madhu-1:
@j-griffith I will leave it to you. As you have more information about it

Madhu-1:
@j-griffith addressed review comments, PTAL

Madhu-1:
@j-griffith anything needed on this one?

j-griffith:
@Madhu-1 Sorry, I dropped this one; looks good to me now, thanks for making those changes.

j-griffith:
Wooohooo!

PR #310, opened 2019-07-01
use volume ID for creating PVC from another PVC

Madhu-1:
**What type of PR is this?**

>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

current PVC request is sending the PV name while creating a volume from a volume, as per the CSI spec we need to send the volume_id for creation of PVC from another PVC.

Fixes #

```release-note
This PR send the volume_id instead of  PVC Name to the csi-driver in volumeCreate when datasource  is PVC
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -671,9 +671,22 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
 		return nil, fmt.Errorf("error, new PVC request must be greater than or equal in size to the specified PVC data source, requested %v but source is %v", requestedSize, sourcePVC.Spec.Size())
 	}
 
+	if sourcePVC.Spec.VolumeName == "" {
+		return nil, fmt.Errorf("volume name is empty in source PVC %s", sourcePVC.Name)
+	}
+
+	sourcePV, err := p.client.CoreV1().PersistentVolumes().Get(sourcePVC.Spec.VolumeName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("error getting PV %s from api server: %v", sourcePVC.Spec.VolumeName, err)
+	}
+
+	if sourcePV.Spec.CSI == nil {
+		return nil, fmt.Errorf("error getting volume source from persistantVolumeClaim:persistanceVolume %s:%s", sourcePVC.Name, sourcePVC.Spec.VolumeName)
+	}
+
 	volumeSource := csi.VolumeContentSource_Volume{
 		Volume: &csi.VolumeContentSource_VolumeSource{
-			VolumeId: sourcePVC.Spec.VolumeName,
+			VolumeId: sourcePV.Spec.CSI.VolumeHandle,
 		},
 	}
 	klog.V(5).Infof("VolumeContentSource_Volume %+v", volumeSource)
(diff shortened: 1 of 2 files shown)


msau42:
/ok-to-test

Please also add a release note

pkg/controller/controller.go
```diff
@@ -671,9 +671,18 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
...
+	}
+
+	if sourcePV.Spec.CSI == nil {
+		return nil, fmt.Errorf("error getting volume source from persistantVolumeClaim:persistanceVolume %s:%s", sourcePVC.Name, sourcePVC.Spec.VolumeName)
+	}
+
 	volumeSource := csi.VolumeContentSource_Volume{
 		Volume: &csi.VolumeContentSource_VolumeSource{
-			VolumeId: sourcePVC.Spec.VolumeName,
+			VolumeId: sourcePV.Spec.CSI.VolumeHandle,
```
msau42: Do we have a test for this?

Madhu-1: updated test case

Madhu-1:
> /ok-to-test
> 
> Please also add a release note

@msau42 sorry where I have to add a release note

msau42:
Please add the release note in the original PR comment, where the blank box currently is

pkg/controller/controller.go
msau42: Thanks, can the unit test also validate the pv spec has the correct volumeid?

Madhu-1: @msau42 this is already done here https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller_test.go#L2195

Madhu-1:
@msau42 added a release note PTAL

msau42:
/approve
Will let @j-griffith lgtm

pkg/controller/controller.go
```diff
@@ -671,9 +671,18 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
 		return nil, fmt.Errorf("error, new PVC request must be greater than or equal in size to the specified PVC data source, requested %v but source is %v", requestedSize, sourcePVC.Spec.Size())
 	}
 
+	sourcePV, err := p.client.CoreV1().PersistentVolumes().Get(sourcePVC.Spec.VolumeName, metav1.GetOptions{})
```
j-griffith: The `PVC.Spec.VolumeName` field is "optional", would it be more reliable to just use the `Volume` field of the PVC directly?  Then you can do something like `volumeHandleToId` to set the `VolumeId` field below.

Madhu-1: do we have `volume` field in PVC spec? 

The approach I followed here is to get the PV name from PVC and then fetch volumeID from PV spec. do we have any other way of getting the volume name from the PVC?

j-griffith: Actually you're correct here, Spec.VolumeName is the only handle; and the CO should set this when the claim is bound so this should be fine.  You might want to add a check that the Spec.VolumeName value is not nil before issuing the Get call for the PV though.

Madhu-1:
@j-griffith done PTAL

PR #312, opened 2019-07-08
Retry provisioning of volumes after transient error

### LINKED ISSUE #311: PVC deletion results in leaked volume on the storage provider, if CreateVolume never completed in time
I am facing an issue with the kubernetes CSI provisioner sidecar that leads to a volume leak on Ceph, the nature of things are as follows,

1. Create a PVC (pointing to a CSI provisioner, in my case ceph-csi)
2. Ensure the request CreateVolume made by the provisioner sidecar will always timeout
  - IOW, ensure it does not succeed till the test is complete (I did this with a simple sleep that was more than the --timeout configured on the provisioner sidecar)
3. Post 1 or 2 retries by the provisioner sidecar to create the volume, delete the PVC

My expectation was that the provisioner would attempt creates till it got an error or a success and then later delete the created volume as the PVC (and PV) are no longer around.

What happened, post step (3), was the provisioner stopped requesting the create, and as the original calls to ceph had succeeded, there was a volume leak on the Ceph side. (neither did the provisioner call delete but that is expected as it does not have the volume ID to call delete with).

Provisioner sidecar versions attempted: v1.2.0 and v1.3.0

logs from the test (provisioner logs and ceph-csi logs): https://paste.fedoraproject.org/paste/J1mddHUmrtMmEbmla8gF9Q

The issue is opened to understand if the behavior is as expected, or there is a bug somewhere.

jsafrane:
The provisioner should retry CreateVolume call after a transient error (such as timeout), because the CSI driver may be creating a volume in the background.

Therefore `ProvisionerExt` interface need to be implemented. `ProvisionExt()` returns:

- `Finished`, if it can be 100% sure that the driver is not creating a volume
- `NoChange`, if something (temporarily?) failed before reaching the CSI driver, for example when Kubernetes API server is not reachable.
- `InBackground`, if error returned by the driver (or gRPC) is transient.

/kind bug

Fixes #311

```release-note
Fixed potential leak of volumes after CSI driver timeouts.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -25,6 +25,9 @@ import (
 	"strings"
 	"time"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-csi/csi-lib-utils/connection"
 	"github.com/kubernetes-csi/external-provisioner/pkg/features"
@@ -179,6 +182,7 @@ type csiProvisioner struct {
 
 var _ controller.Provisioner = &csiProvisioner{}
 var _ controller.BlockProvisioner = &csiProvisioner{}
+var _ controller.ProvisionerExt = &csiProvisioner{}
 
 var (
 	// Each provisioner have a identify string to distinguish with others. This
@@ -353,8 +357,14 @@ func getVolumeCapability(
 }
 
 func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.PersistentVolume, error) {
+	// The controller should call ProvisionExt() instead, but just in case...
+	pv, _, err := p.ProvisionExt(options)
+	return pv, err
+}
+
+func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.PersistentVolume, controller.ProvisioningState, error) {
 	if options.StorageClass == nil {
-		return nil, errors.New("storage class was nil")
+		return nil, controller.ProvisioningFinished, errors.New("storage class was nil")
 	}
 
 	migratedVolume := false
@@ -367,7 +377,7 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
 			klog.V(2).Infof("translating storage class for in-tree plugin %s to CSI", options.StorageClass.Provisioner)
 			storageClass, err := csitranslationlib.TranslateInTreeStorageClassToCSI(p.supportsMigrationFromInTreePluginName, options.StorageClass)
 			if err != nil {
-				return nil, fmt.Errorf("failed to translate storage class: %v", err)
+				return nil, controller.ProvisioningFinished, fmt.Errorf("failed to translate storage class: %v", err)
 			}
 			options.StorageClass = storageClass
 			migratedVolume = true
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -449,7 +459,7 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
 	if options.PVC.Spec.DataSource != nil && (rc.clone || rc.snapshot) {
 		volumeContentSource, err := p.getVolumeContentSource(options)
 		if err != nil {
-			return nil, fmt.Errorf("error getting handle for DataSource Type %s by Name %s: %v", options.PVC.Spec.DataSource.Kind, options.PVC.Spec.DataSource.Name, err)
+			return nil, controller.ProvisioningNoChange, fmt.Errorf("error getting handle for DataSource Type %s by Name %s: %v", options.PVC.Spec.DataSource.Kind, options.PVC.Spec.DataSource.Name, err)
```
ShyamsundarR: This is possibly more a clarification than a request for change, why are the following series of errors `ProvisioningNoChange` rather than `ProvisioningFinished`?

We have not issued the `CreateVolume` and there is an error, do we want the provisioner to retry, because these errors are recoverable?

jsafrane: Because this may be second iteration of provisioning. The first one failed in CSI driver with `ProvisioningInBackground`. The provisioner tries for second time and now  due to a network hiccup it cannot e.g. get a Secret. But the provisioning is still in progress from the last attempt, hence `NoChange` and the provisioner must try again until it succeeds.

Which may be tricky when someone e.g. has deleted the Secret. So there it is still possible to leak a volume, but only when admin actively screws something up.

ShyamsundarR: Ack! thank you for the explanation.

humblec:
LGTM. Thanks @jsafrane 
@wongma7 @msau42 Can you review this please?  the volume leaking is a serious issue , so looking for a quick review here :)

pkg/controller/controller.go
```diff
@@ -589,13 +603,13 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
 		if err != nil {
 			klog.Warningf("failed to translate CSI PV to in-tree due to: %v. Deleting provisioned PV", err)
 			p.Delete(pv)
-			return nil, err
+			return nil, controller.ProvisioningFinished, err
```
msau42: We don't check return value of `p.Delete()`. What if that deletion failed partially?

pkg/controller/controller.go
```diff
@@ -1004,3 +1018,26 @@ func deprecationWarning(deprecatedParam, newParam, removalVersion string) string
...
+	switch st.Code() {
+	case codes.Canceled, // gRPC: Client Application cancelled the request
+		codes.DeadlineExceeded,  // gRPC: Timeout
+		codes.Unavailable,       // gRPC: Server shutting down, TCP connection broken - previous CreateVolume() may be still in progress.
+		codes.ResourceExhausted: // gRPC: Server temporarily out of resources - previous Provision() may be still in progress.
+		return false
+	}
+	// All other errors mean that provisioning either did not
+	// even start or failed. It is for sure not in progress.
+	return true
```
msau42: Would it be safer to do the reverse?  Whitelist known final error codes (that we define in the csi spec), and assume all other error codes need to be retried?

pkg/controller/controller.go
jsafrane: Fixed

jsafrane:
+ fixed vendoring

pkg/controller/controller.go
jsafrane: I am not sure it would bring any value. It has been working well similarly in external-attacher.

msau42: It means that if some new error code was added or we misunderstood the meaning of some other error code, then we will do the safer operation to retry

msau42: Said another way, by default, we will retry

pkg/controller/controller.go
```diff
@@ -588,14 +602,19 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
 		pv, err = csitranslationlib.TranslateCSIPVToInTree(pv)
 		if err != nil {
 			klog.Warningf("failed to translate CSI PV to in-tree due to: %v. Deleting provisioned PV", err)
-			p.Delete(pv)
-			return nil, err
+			deleteErr := p.Delete(pv)
+			if err != nil {
```
msau42: Should be `deleteErr`

pkg/controller/controller.go
msau42: CSI spec also defines ABORTED error code as operation in progress

jsafrane: Not for `CreateVolumeRequest`. IMO it's for multiple calls for a single volume and there is no volume yet.

Or is there a bug in CSI spec?

pkg/controller/controller.go
jsafrane: fixed

pkg/controller/controller.go
jsafrane: I'm re-reading CSI spec again, it seems that ABORTED can be returned to any call... I added it to the list.

pkg/controller/controller.go
```diff
@@ -539,7 +552,8 @@ func (p *csiProvisioner) Provision(options controller.ProvisionOptions) (*v1.Per
 		if err != nil {
 			capErr = fmt.Errorf("%v. Cleanup of volume %s failed, volume is orphaned: %v", capErr, pvName, err)
```
msau42: Is this message still relevant if we retry?

PR #314, opened 2019-07-16
Remove redundant mount options test and fold into normal provisioner test

davidz627:
/assign @bswartz @jsafrane

/kind cleanup

```release-note
NONE
```


davidz627:
/assign @msau42 
for approval

PR #320, opened 2019-07-17
[1.1] Cherry pick: Retry provisioning of volumes after transient error

jsafrane:
This is cherry-pick of #312 into 1.1

---

The provisioner should retry CreateVolume call after a transient error
(such as timeout), because the CSI driver may be creating a volume in the
background.

Therefore ProvisionerExt interface need to be implemented. ProvisionExt()
returns:

- Finished, if it can be 100% sure that the driver is not creating a volume
- NoChange, if something (temporarily?) failed before reaching the CSI
  driver, for example when Kubernetes API server is not reachable.
- InBackground, if error returned by the driver (or gRPC) is transient.

```release-note
Fixed potential leak of volumes after CSI driver timeouts.
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -836,6 +836,8 @@
     "github.com/kubernetes-csi/external-snapshotter/pkg/client/clientset/versioned/fake",
     "github.com/spf13/pflag",
     "google.golang.org/grpc",
+    "google.golang.org/grpc/codes",
+    "google.golang.org/grpc/status",
     "k8s.io/api/core/v1",
     "k8s.io/api/storage/v1",
     "k8s.io/api/storage/v1beta1",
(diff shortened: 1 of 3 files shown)


msau42:
Looks like we need to update release tools for 1.1 branch like #315

pohly:
> Looks like we need to update release tools for 1.1 branch like #315

Isn't the release-1.1 branch unmaintained now because we don't need that version of the external-provisioner for anything? At least that's what I thought and therefore I hadn't created a csi-release-tools PR for that branch.

According to https://kubernetes-csi.github.io/docs/external-provisioner.html we currently need release-1.0 (for Kubernetes 1.13), release-1.2 (for 1.14) and release-1.3 (for 1.15).

jsafrane:
Sorry for the noise
/close

jsafrane:
Just for the reference, there is no `ProvisionerExt` interface in 1.0.1, so this PR does not need to be backported there.

PR #323, opened 2019-08-09
Update go-client to get fix for selflink deprecation

msau42:
**What type of PR is this?**
/kind bug

SelfLink is being deprecated in https://github.com/kubernetes/kubernetes/pull/80978 and we need to update the client-go dependency to pick up a fix in `GetReference` which is being called by sig-storage-lib-external-provisioner [here](https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/c525773885fccef89a1edd22bb3f813d50548ed1/controller/controller.go#L1214)

Fixes #

```release-note
Fixes issue when SelfLink removal is turned on in Kubernetes.
```


msau42:
cc @wojtek-t

I still need to test this on Kubernetes alpha clusters on master.

pull-kubernetes-csi-csi-driver-host-path-alpha-1-15-on-kubernetes-master can do it but currently master jobs are broken, which I am also trying to fix: https://github.com/kubernetes-csi/csi-driver-host-path/pull/77

msau42:
/assign @jsafrane @childsb 
all the external-provisioners also need to be updated and rebuilt

msau42:
The alpha master job pull job passed.  I'm also manually running all the k/k e2es with a private hostpath csi driver on an alpha master cluster and those are passing as well.

wojtek-t:
/lgtm

/hold
[If you don't think it's required to someone else to also take a look (in fact it's pretty much only dependency update, feel free to cancel hold]

wojtek-t:
/hold cancel
[didn't mean to add it given my lgtm wasn't effective anyway]

PR #327, opened 2019-08-12
Use csiNode informer to prevent hitting the API server for all time

### LINKED ISSUE #144: Consider using Node & CSINodeInfo informers

mucahitkurt:
**What type of PR is this?**
/kind cleanup

Use csiNode shared index informer instead of hitting the API server for all csiNode resource request.

Fixes #144

```release-note
NONE
```

cc @msau42


cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -179,9 +181,27 @@ func main() {
 		provisionerOptions = append(provisionerOptions, controller.AdditionalProvisionerNames([]string{supportsMigrationFromInTreePluginName}))
 	}
 
+	// Create informer to prevent hit the API server for all resource request
+	csiNodeInformer := v1beta1.NewCSINodeInformer(clientset, 1*time.Hour, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	go csiNodeInformer.Run(context.TODO().Done())
```
verult: IMO context.Background() is more appropriate

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -179,9 +181,27 @@ func main() {
 		provisionerOptions = append(provisionerOptions, controller.AdditionalProvisionerNames([]string{supportsMigrationFromInTreePluginName}))
 	}
 
+	// Create informer to prevent hit the API server for all resource request
+	csiNodeInformer := v1beta1.NewCSINodeInformer(clientset, 1*time.Hour, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
```
verult: Change `1*time.Hour` into a named const. Ditto for all the other time constants

mucahitkurt:
> Thanks for the PR!
> 
> Was the Gopkg.lock change intentional? There isn't any Gopkg.toml changes.
> 
> Did you get a chance to test the provisioner end to end manually?

Thanks for the comments, I'll try to fix them.

Yes, Gopkg.lock change is intentional, Gopkg.toml is not changed because I did not add a new library dependency, I just added some package dependency under client-go.

No,  I didn't make e2e manual test, TBH I'm not very familiar with the all environment of external provisioner but I'll try to run end test the external provisioner.

pkg/controller/topology_test.go
```diff
@@ -1528,3 +1550,20 @@ func requisiteEqual(t1, t2 []*csi.Topology) bool {
 
 	return unchecked.Len() == 0
 }
+
+func runCsiNodeInformer(kubeClient *fakeclientset.Clientset, t *testing.T, csiNodeNum int) (cache.SharedIndexInformer, chan struct{}) {
```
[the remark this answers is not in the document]
mucahitkurt: I'll delete it after HasSynced method usage

pkg/controller/topology_test.go
```diff
@@ -25,13 +25,19 @@ import (
...
 	"k8s.io/kubernetes/pkg/apis/core/helper"
+	"time"
 )
 
 const (
-	testDriverName     = "com.example.csi/test-driver"
-	preBetaNodeVersion = "1.13.0"
+	testDriverName      = "com.example.csi/test-driver"
+	preBetaNodeVersion  = "1.13.0"
+	interval            = 100 * time.Millisecond
```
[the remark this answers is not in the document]
mucahitkurt: I'll delete this after HasSynced method usage

pkg/controller/topology.go
```diff
@@ -263,12 +265,12 @@ func GenerateAccessibilityRequirements(
...
 // getSelectedCSINode returns the CSINode object for the given selectedNode.
 func getSelectedCSINode(
-	kubeClient kubernetes.Interface,
+	informer cache.SharedIndexInformer,
 	selectedNode *v1.Node) (*storage.CSINode, error) {
 
-	// TODO (#144): use informers
-	selectedNodeInfo, err := kubeClient.StorageV1beta1().CSINodes().Get(selectedNode.Name, metav1.GetOptions{})
-	if err != nil {
+	key := fmt.Sprintf("%s/%s", selectedNode.Namespace, selectedNode.Name)
```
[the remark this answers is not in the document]
mucahitkurt: No it's not right, it should be only name

msau42:
It might be worth implementing topology first in the hostpath csi driver: https://github.com/kubernetes-csi/csi-driver-host-path/issues/54.  Then, we have automated pull/ci jobs that can test the topology path. (We'll also need to update those pull jobs to bring up a multi-node cluster)

I am also working on adding additional topology tests as part of https://github.com/kubernetes/kubernetes/issues/71289

mucahitkurt:
@verult Do you have any idea why `pull-kubernetes-csi-external-provisioner-1-13-on-kubernetes-1-13` tests are failed? Thanks!

msau42:
```
E0816 19:40:18.164363       1 reflector.go:125] github.com/kubernetes-csi/external-provisioner/cmd/csi-provisioner/csi-provisioner.go:186: Failed to list *v1beta1.CSINode: the server could not find the requested resource
```

Topology was an alpha feature in 1.13, so the beta object doesn't exist.  To avoid this, we should only create the informer if the topology feature gate in the external-provisioner is enabled.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -179,9 +182,33 @@ func main() {
 		provisionerOptions = append(provisionerOptions, controller.AdditionalProvisionerNames([]string{supportsMigrationFromInTreePluginName}))
 	}
 
+	var csiNodeInformer cache.SharedIndexInformer
+	if utilfeature.DefaultFeatureGate.Enabled(features.Topology) {
+		// Create informer to prevent hit the API server for all resource request
+		csiNodeInformer := v1beta1.NewCSINodeInformer(clientset, ctrl.ResyncPeriodOfCsiNodeInformer, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
```
msau42: I'm not sure indexing by namespace makes sense because CSINode is a non-namespaced object.

pkg/controller/controller.go
```diff
@@ -235,7 +239,8 @@ func NewCSIProvisioner(client kubernetes.Interface,
 	pluginCapabilities connection.PluginCapabilitySet,
 	controllerCapabilities connection.ControllerCapabilitySet,
 	supportsMigrationFromInTreePluginName string,
-	strictTopology bool) controller.Provisioner {
+	strictTopology bool,
+	informer cache.SharedIndexInformer) controller.Provisioner {
```
msau42: instead of passing in the entire informer to the provisioner, how about only passing in the Lister(), since all we need is Get() and List()

cmd/csi-provisioner/csi-provisioner.go
mucahitkurt: I'll change to `cache.Indexers{}`

pkg/controller/controller.go
mucahitkurt: I'll change to `informer.GetStore()`

verult:
I'm not an expert with informers, but I'm not sure if this is the right way to instantiate the informer. The external-attacher example has something like this:
https://github.com/kubernetes-csi/external-attacher/blob/c745c0f1032a6b30f376577ce5ce9262cd44f74d/cmd/csi-attacher/main.go#L101

And the lister (which @msau42 referred to) is probably what we want instead: https://github.com/kubernetes-csi/external-attacher/blob/c745c0f1032a6b30f376577ce5ce9262cd44f74d/cmd/csi-attacher/main.go#L142

mucahitkurt:
> I'm not an expert with informers, but I'm not sure if this is the right way to instantiate the informer. The external-attacher example has something like this:
> https://github.com/kubernetes-csi/external-attacher/blob/c745c0f1032a6b30f376577ce5ce9262cd44f74d/cmd/csi-attacher/main.go#L101
> 
> And the lister (which @msau42 referred to) is probably what we want instead: https://github.com/kubernetes-csi/external-attacher/blob/c745c0f1032a6b30f376577ce5ce9262cd44f74d/cmd/csi-attacher/main.go#L142

I saw the usage of `SharedInformerFactory` factory(and the warning about not using directly informer) but I confused  about `SharedInformer` and `SharedIndexInformer`. I thought that `SharedInformerFactory` produces `SharedInformer` and I need `SharedIndexInformer`. But `SharedInformerFactory` already produces `SharedIndexInformer`.

So, I'll change my usage like the one external-attacher uses, thanks for the heads up!

mucahitkurt:
@verult would you PTAL? Thanks!

msau42:
/retest
Now that hostpath driver deployment has been updated to use topology and we create 2-node clusters, topology codepath can be better exercised now

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -179,9 +182,31 @@ func main() {
 		provisionerOptions = append(provisionerOptions, controller.AdditionalProvisionerNames([]string{supportsMigrationFromInTreePluginName}))
 	}
 
+	var csiNodeLister storagelisters.CSINodeLister
+	var factory informers.SharedInformerFactory
+	if utilfeature.DefaultFeatureGate.Enabled(features.Topology) {
+		// Create informer to prevent hit the API server for all resource request
+		factory = informers.NewSharedInformerFactory(clientset, ctrl.ResyncPeriodOfCsiNodeInformer)
```
verult: if the lister is only necessary when topology feature is used, should it use `supportsTopology()` instead to capture all conditions?

https://github.com/kubernetes-csi/external-provisioner/blob/b9613492f7c958d9ba615f71dcbe64d392a9888b/pkg/controller/controller.go#L657

pkg/controller/topology.go
```diff
@@ -289,37 +291,33 @@ func getSelectedCSINode(
...
-		// TODO (#144): use informers
-		nodeInfos, err := kubeClient.StorageV1beta1().CSINodes().List(metav1.ListOptions{})
+		csiNodes, err := csiNodeLister.List(labels.Everything())
 		if err != nil {
-			// Require CSINode beta feature on K8s apiserver to be enabled.
-			// We don't want to fallback and provision in the wrong topology if there's some temporary
-			// error with the API server.
-			return nil, fmt.Errorf("error listing CSINodes: %v", err)
+			klog.Warningf("No csi nodes found")
+			return nil, nil
```
verult: How come it's OK to not return an error now?

cmd/csi-provisioner/csi-provisioner.go
mucahitkurt: Do you mean to change the condition `utilfeature.DefaultFeatureGate.Enabled(features.Topology)` with `pluginCapabilities[csi.PluginCapability_Service_VOLUME_ACCESSIBILITY_CONSTRAINTS] &&
		utilfeature.DefaultFeatureGate.Enabled(features.Topology)` or do you mean to use `csiNodeLister` as initiated inside the `csiProvisioner` with the `supportsTopology()` check instead of a external provided property of the struct?

pkg/controller/topology.go
mucahitkurt: I couldn't remember any valid case for this change, I think I changed it in a wrong way, I'll change it to return error.

cmd/csi-provisioner/csi-provisioner.go
mucahitkurt: I added additional check for plugin capabilities.

verult: Yup this is what I meant, though in its current form, I'm worried that someone later on will only change one place but not the other. It would be good to have a shared utility function instead in a different file (maybe package/util)

mucahitkurt: Makes sense, thanks. Putting the function inside `topology.go` instead of util package makes sense to me.

mucahitkurt:
@verult PTAL

verult:
/lgtm

Thanks @mucahitkurt !

vendor/modules.txt
```diff
@@ -168,10 +168,6 @@ k8s.io/api/apps/v1
 k8s.io/api/apps/v1beta1
```
msau42: is updating this file necessary? it looks like just some lines got moved around?

mucahitkurt: Yes content is the same, just some order changes, but I'm getting error from CI like below when I don't change this file with the below command;

`ERROR: vendor directory *not* up-to-date, it did get modified by 'GO111MODULE=on go mod vendor':`

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-provisioner/327/pull-kubernetes-csi-external-provisioner-unit/1187927085243437058

https://travis-ci.org/kubernetes-csi/external-provisioner/builds/603113049?utm_source=github_status&utm_medium=notification

Issue #329, opened 2019-08-14
"PV created" log message too verbose


saad-ali:
**Problem**
Today when a PV is created the external-provisioner logs the entire PV object volume source:

https://github.com/kubernetes-csi/external-provisioner/blob/c098da646c86fab4680cb12e8e3eee5fb62976ab/pkg/controller/controller.go#L620

This is pretty verbose and can fill up logs fast if you are creating a lot of disks (the person who complained about this is creating 1000s of disks per day).

However we shouldn't just blindly reduce the verbosity of this log line because it is important for normal debugging to know when a PV is successfully created.

**Suggested Fix**

Let's break this log line in to two:

1. A standard verbosity `Infof` that says PV successfully created with minimal info (e.g. PVC name, PV name, CSI driver name, and CSI volume name).
2. A L4 or L5 verbosity log message with the full PV object volume source.

PR #331, opened 2019-08-19
Improve PV creation logging

Madhu-1:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

log PV object with v(5) verbosity.

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

Fixes https://github.com/kubernetes-csi/external-provisioner/issues/329

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -617,8 +617,8 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 		}
 	}
 
-	klog.Infof("successfully created PV %+v", pv.Spec.PersistentVolumeSource)
-
+	klog.V(2).Infof("successfully created PV %v for PVC %v and csi volume name %v", pv.Name, options.PVC.Name, pv.Spec.CSI.VolumeHandle)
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -617,8 +617,8 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 		}
 	}
 
-	klog.Infof("successfully created PV %+v", pv.Spec.PersistentVolumeSource)
-
+	klog.Infof("successfully created PV %v for PVC %v driver name %v csi volume name %v", pv.Name, options.PVC.Name, pv.Spec.CSI.Driver, pv.Spec.CSI.VolumeHandle)
```
msau42: Explicitly set a log level here.  How about 2?

msau42: I don't think we need to log driver name here? The external-provisioner only handles one driver.

Madhu-1: yes you  are right, removed driver name logging and set the log level to 2

PR #334, opened 2019-08-27
Readable volume names

### LINKED ISSUE #67: PVC.UID for volume name unique but hard for humans to differentiate
Issue raised by @cduchesne 

With PR https://github.com/kubernetes-csi/external-provisioner/pull/66 the names of the newly provisioned volumes will be based on the UID of the Kubernetes PVC object. Although the UIDs are unique, the UID does not differ much, making it harder for humans to read and differentiate between them.

Example from @cduchesne:
```
[root@prometheus csi-scaleio]# for i in {01..10}; do curl -s http://127.0.0.1:8080/api/v1/namespaces/scaleio/persistentvolumeclaims/vol$i | jq '.metadata.uid'; done
"f151bfd8-3751-11e8-ba9f-0cc47ac6c23c"
"f16e31ee-3751-11e8-ba9f-0cc47ac6c23c"
"f18a35c0-3751-11e8-ba9f-0cc47ac6c23c"
"f1a737cc-3751-11e8-ba9f-0cc47ac6c23c"
"f1c6e464-3751-11e8-ba9f-0cc47ac6c23c"
"f1e6f657-3751-11e8-ba9f-0cc47ac6c23c"
"f2034a06-3751-11e8-ba9f-0cc47ac6c23c"
"f223b1e4-3751-11e8-ba9f-0cc47ac6c23c"
"f2422240-3751-11e8-ba9f-0cc47ac6c23c"
"f25d4bec-3751-11e8-ba9f-0cc47ac6c23c"
```

Since the primary issue here is readability and Kubernetes does the same thing (https://github.com/kubernetes-csi/external-provisioner/pull/66#discussion_r178855141), I would consider this issue low pri. It would be nice if we can come up with a better naming scheme that is deterministicly generated from PVC without collisions.

mcronce:
**What type of PR is this?**
/kind feature

Adds an option for the dynamic provisioner to produce PVs with human-readable names.  This enables the storage administrator to go straight to the storage layer, usually[1] without having to detour to the Kubernetes API to ask about PV names, to manually inspect or hand-edit an application's config/database/etc.

[1] Even with this enabled, in cases where you have lots of PVCs with similar names in similarly-named namespaces, you might still need to ask the Kubernetes API to find the right PV.

Fixes #67 

I've had these changes made for nine months and have been merging master in periodically, using a container image built from my fork in a staging environment.  It's definitely passed the test of time.  :)

```release-note
Adds an option (`--volume-names-readable`) to produce PVs with human-readable names; with cooperation from the CSI driver, this can make it easier for storage administrators to find state for a given application.
```


pkg/controller/controller.go
```diff
@@ -286,22 +290,36 @@ func (p *csiProvisioner) checkDriverCapabilities(rc *requiredCapabilities) error
...
+		if len(pvcNamespace) == 0 {
+			return "", fmt.Errorf("corrupted PVC object, it is missing namespace")
+		}
+		if len(pvcName) == 0 {
+			return "", fmt.Errorf("corrupted PVC object, it is missing name")
+		}
+		pvcHash := md5.Sum([]byte(pvcNamespace + "/" + pvcName))
+		pvcNamespace = strings.Replace(pvcNamespace, "-", "", -1)
+		pvcName = strings.Replace(pvcName, "-", "", -1)
+		fullName = fmt.Sprintf("%s-%s-%x", pvcNamespace, pvcName, pvcHash)
```
jsafrane: Both namespace name and PV names can be up 63 characters long. PV name is also limited to 63 characters. With a long namespace names, there may be no space for PVC name nor the hash and the resulting name is not unique.

jsafrane:
Please do not use merge commits, branch always from master branch and rebase periodically.

jsafrane:
And if you find a better way how to compute unique, yet human friendly names, please add unit tests.

pkg/controller/controller.go
mcronce: I didn't realize PV names were limited to 63 characters; however, even with that in mind, it's going to be desirable for some administrators (me, for example) to trade having to be careful about namespace/PVC naming for being able to find storage more easily.

I can certainly add more clarity to the help text on the flag.

msau42:
The PVC name + namespace is available in the PV object.  Is it sufficient when listing PV objects to have it output the PVC name + namespace?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -53,6 +53,7 @@ var (
 	_                    = deprecatedflags.Add("connection-timeout")
 	volumeNamePrefix     = flag.String("volume-name-prefix", "pvc", "Prefix to apply to the name of a created volume.")
 	volumeNameUUIDLength = flag.Int("volume-name-uuid-length", -1, "Truncates generated UUID of a created volume to this length. Defaults behavior is to NOT truncate.")
+	volumeNamesReadable  = flag.Bool("volume-names-readable", false, "If enabled, includes the PVC namespace and name in VolumeRequests' suggested names.  Note that, combined with --volume-name-uuid-length, this can cause naming collisions.")
```
jsafrane: > this can cause naming collisions

Name collisions will lead to a PVC that will be Pending forever, because name of the PV it wants collided with another PVC. Users should not face these collisions at all, they're unpredictable and hard to fix - it requires users to change deployments of their applications.

pkg/controller/controller.go
jsafrane: > Both namespace name and PV names can be up 63 characters long. PV name is also limited to 63 characters

I checked today and I was wrong, namespace name is has limit 63 characters, but PVC and PV names have 253 characters.

Still, with PVC a long PVC name, the code above produces PV name that will be rejected by API server. And it does not allow any space to add namespace name to a very long PVC name.

cmd/csi-provisioner/csi-provisioner.go
jsafrane: Name collision can lead to stealing data. If attacker crafts namespace + pvc name that leads to collision with existing volume, CSI driver, obeying idempotency, must return the same volume as someone else is already using. Kubernetes is multi-tenant system and must not allow stealing data between namespaces.

pkg/controller/controller.go
jsafrane: > Both namespace name and PV names can be up 63 characters long. PV name is also limited to 63 characters

I checked today and I was wrong, namespace name is has limit 63 characters, but PVC name can have 253 characters. CSI specifies that any string, incl. volume name, can have up to 128 characters. So you don't have even place for full PVC name, not mentioning the namespace.

cmd/csi-provisioner/csi-provisioner.go
mcronce: Yes; that is a trade off that must be considered when using this option.  Some of us are not administrating multi-tenant Kubernetes systems.  We're administering single-tenant Kubernetes systems that, for example, require an easy to way to find a given application's data in the storage provider.

mcronce:
> The PVC name + namespace is available in the PV object. Is it sufficient when listing PV objects to have it output the PVC name + namespace?

Unfortunately, it does not.  When I'm on a storage node I don't have access to the Kubernetes API.

PR #335, opened 2019-08-30
Update snapshot CRD to v1beta

xing-yang:
**What type of PR is this?**
/kind feature

This PR updates dependencies to VolumeSnapshot CRD v1beta1.

Fixes #

Note: Dependency files from external-snapshotter are manually copied at this point.
Snapshot beta APIs are not finalized yet.

```release-note
Updates VolumeSnapshot CRD to v1beta1
```


vendor/github.com/kubernetes-csi/external-snapshotter/pkg/apis/volumesnapshot/v1beta1/types.go
```diff
@@ -0,0 +1,369 @@
...
+	// driver is the name of the CSI driver used to create the physical snapshot on
+	// the underlying storage system.
+	// This MUST be the same as the name returned by the CSI GetPluginName() call for
+	// that driver.
+	// Required.
+	Driver string `json:"driver" protobuf:"bytes,3,opt,name=driver"`
+
+	// name of the SnapshotClass to which this snapshot belongs.
+	// +optional
+	SnapshotClassName *string `json:"snapshotClassName,omitempty" protobuf:"bytes,4,opt,name=snapshotClassName"`
```
xing-yang: We should call it VolumeSnapshotClassName to be consistent with the same thing in VolumeSnapshotSpec.  VolumeSnapshotClassName was also the original name in the content spec.

xing-yang:
Hi @msau42, this PR uses the latest external-snapshotter master code.  It is ready for review.

pkg/controller/controller.go
```diff
@@ -808,36 +808,31 @@ func (p *csiProvisioner) getSnapshotSource(options controller.ProvisionOptions)
 	}
 	klog.V(5).Infof("VolumeSnapshot %+v", snapshotObj)
 
-	snapContentObj, err := p.snapshotClient.VolumesnapshotV1alpha1().VolumeSnapshotContents().Get(snapshotObj.Spec.SnapshotContentName, metav1.GetOptions{})
-	if err != nil {
-		klog.Warningf("error getting snapshotcontent %s for snapshot %s/%s from api server: %s", snapshotObj.Spec.SnapshotContentName, snapshotObj.Namespace, snapshotObj.Name, err)
-		return nil, fmt.Errorf("snapshot in dataSource not bound or invalid")
+	if snapshotObj.Status.BoundVolumeSnapshotContentName == nil {
+		return nil, fmt.Errorf("snapshot %s does not have a content", options.PVC.Spec.DataSource.Name)
```
msau42: how about "snapshot %s not bound"?

pkg/controller/controller.go
```diff
@@ -795,11 +795,11 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
 // getSnapshotSource verifies DataSource.Kind of type VolumeSnapshot, making sure that the requested Snapshot is available/ready
 // returns the VolumeContentSource for the requested snapshot
 func (p *csiProvisioner) getSnapshotSource(options controller.ProvisionOptions) (*csi.VolumeContentSource, error) {
-	snapshotObj, err := p.snapshotClient.VolumesnapshotV1alpha1().VolumeSnapshots(options.PVC.Namespace).Get(options.PVC.Spec.DataSource.Name, metav1.GetOptions{})
+	snapshotObj, err := p.snapshotClient.SnapshotV1beta1().VolumeSnapshots(options.PVC.Namespace).Get(options.PVC.Spec.DataSource.Name, metav1.GetOptions{})
 	if err != nil {
 		return nil, fmt.Errorf("error getting snapshot %s from api server: %v", options.PVC.Spec.DataSource.Name, err)
 	}
-	if snapshotObj.Status.ReadyToUse == false {
+	if snapshotObj.Status.ReadyToUse == nil || *snapshotObj.Status.ReadyToUse == false {
```
msau42: Can we move this check below the status.Bound check?  I think it makes sense in terms of ordering to:
1. Check for deletion
2. Check for bound
3. Check if ready

pkg/controller/controller.go
```diff
@@ -808,36 +808,31 @@ func (p *csiProvisioner) getSnapshotSource(options controller.ProvisionOptions)
...
-		klog.Warningf("snapshotcontent %s for snapshot %s/%s is not bound", snapshotObj.Spec.SnapshotContentName, snapshotObj.Namespace, snapshotObj.Name)
-		return nil, fmt.Errorf("snapshot in dataSource not bound or invalid")
-	}
+	snapContentObj, err := p.snapshotClient.SnapshotV1beta1().VolumeSnapshotContents().Get(*snapshotObj.Status.BoundVolumeSnapshotContentName, metav1.GetOptions{})
 
-	if snapContentObj.Spec.VolumeSnapshotRef.UID != snapshotObj.UID || snapContentObj.Spec.VolumeSnapshotRef.Namespace != snapshotObj.Namespace || snapContentObj.Spec.VolumeSnapshotRef.Name != snapshotObj.Name {
-		klog.Warningf("snapshotcontent %s for snapshot %s/%s is bound to a different snapshot", snapshotObj.Spec.SnapshotContentName, snapshotObj.Namespace, snapshotObj.Name)
+	if err != nil {
+		klog.Warningf("error getting snapshotcontent %s for snapshot %s/%s from api server: %s", *snapshotObj.Status.BoundVolumeSnapshotContentName, snapshotObj.Namespace, snapshotObj.Name, err)
 		return nil, fmt.Errorf("snapshot in dataSource not bound or invalid")
```
msau42: maybe change all these user-visible error messages to the same as above "snapshot %s not bound" (and make it a constant)

pkg/controller/controller.go
```diff
@@ -808,36 +808,31 @@ func (p *csiProvisioner) getSnapshotSource(options controller.ProvisionOptions)
...
+	if snapContentObj.Spec.Driver != options.StorageClass.Provisioner {
+		klog.Warningf("snapshotcontent %s for snapshot %s/%s is handled by a different CSI driver than requested by StorageClass %s", *snapshotObj.Status.BoundVolumeSnapshotContentName, snapshotObj.Namespace, snapshotObj.Name, options.StorageClass.Name)
 		return nil, fmt.Errorf("snapshot in dataSource not bound or invalid")
 	}
 
 	klog.V(5).Infof("VolumeSnapshotContent %+v", snapContentObj)
 	snapshotSource := csi.VolumeContentSource_Snapshot{
 		Snapshot: &csi.VolumeContentSource_SnapshotSource{
-			SnapshotId: snapContentObj.Spec.VolumeSnapshotSource.CSI.SnapshotHandle,
+			SnapshotId: *snapContentObj.Status.SnapshotHandle,
```
msau42: Could this be nil?

xing-yang: Yes.  Added a check.

pkg/controller/controller.go
xing-yang: Sure

pkg/controller/controller.go
xing-yang: Sure

pkg/controller/controller.go
xing-yang: Sure

pkg/controller/controller.go
msau42: Can we also add unit tests for these various new nil checks?  It may be simpler to restructure the test to pass in the exact content object to use, or use a builder pattern

xing-yang: sure.

xing-yang:
@msau42 unit tests added. PTAL.  Thanks.

PR #337, opened 2019-09-06
Use Node informer to prevent hitting the API server for all time

### LINKED ISSUE #144: Consider using Node & CSINodeInfo informers

mucahitkurt:
**What type of PR is this?**
/kind cleanup

Use Node informer instead of hitting the API server for all Node resource request.

Fixes #144

I used the commit from the PR: https://github.com/kubernetes-csi/external-provisioner/pull/327, I can also change that PR to add node informer.

```release-note
NONE
```

cc @msau42 @verult

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -43,6 +43,7 @@ import (
 
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
+	"k8s.io/client-go/listers/core/v1"
 	storagelisters "k8s.io/client-go/listers/storage/v1beta1"
 	utilflag "k8s.io/component-base/cli/flag"
 	csitrans "k8s.io/csi-translation-lib"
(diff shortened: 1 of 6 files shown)


mucahitkurt:
@verult PTAL

PR #339, opened 2019-09-10
WIP: Test hostpath driver with topology

msau42:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -183,7 +183,7 @@ configvar CSI_PROW_WORK "$(mkdir -p "$GOPATH/pkg" && mktemp -d "$GOPATH/pkg/csip
 #
 # When no deploy script is found (nothing in `deploy` directory,
 # CSI_PROW_HOSTPATH_REPO=none), nothing gets deployed.
-configvar CSI_PROW_HOSTPATH_VERSION "v1.2.0-rc2" "hostpath driver"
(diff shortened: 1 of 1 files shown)


msau42:
@pohly how is the version of the hostpath driver deployment chosen? I tried changing CSI_PROW_HOSTPATH_VERSION=v1.2.0-rc6, but it still seems to be deploying v1.1.0 of the hostpath driver

pohly:
> I tried changing CSI_PROW_HOSTPATH_VERSION=v1.2.0-rc6, but it still seems to be deploying v1.1.0 of the hostpath driver

`CSI_PROW_HOSTPATH_VERSION` just controls which revision of the csi-driver-host-path repo is checked out. What gets deployed is then determined by the deployments in that revision, which may differ depending on the Kubernetes version.

In v1.2.0-rc6, all three deployments still use v1.1.0.

msau42:
Thanks I think I was confusing myself with different changes going on.

I found out that we actually need 2 worker nodes to test a multi-node cluster (not a control-plane + worker). I'm trying that out here. Looks like one of the tests is assuming pv supports multiple nodes. We'll need to fix that test case to turn it off for hostpath

PR #343, opened 2019-09-12
Unit tests for migration

davidz627:
/kind cleanup

This PR adds unit tests for CSI Migration. The code should work and is ready for review

/assign @msau42 @jsafrane @ddebroy 
/cc @ddebroy @leakingtapan @shanesiebken

```release-note
NONE
```


pkg/controller/util.go
```diff
@@ -0,0 +1,95 @@
...
+	IsPVMigratable(pv *v1.PersistentVolume) bool
+	TranslateInTreePVToCSI(pv *v1.PersistentVolume) (*v1.PersistentVolume, error)
+
+	IsMigratedCSIDriverByName(csiPluginName string) bool
+	GetInTreeNameFromCSIName(pluginName string) (string, error)
+}
+
+var _ CSITranslationLibraryInterface = &fakeTranslationLibrary{}
+
+type fakeTranslationLibrary struct{}
```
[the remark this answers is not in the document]
misterikkit: Is this fake specific to `pkg/controller/` in any way?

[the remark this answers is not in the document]
davidz627: I think this is closer to scenario 1. I don't really know what to name the file though... `fake_translator_test.go` is what i've come up with so far. It's not specific to `pkg/controller` but I forsee it will only be used there in this repo

pkg/controller/util.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (f *fakeTranslationLibrary) TranslateInTreeStorageClassToCSI(inTreePluginName string, sc *storagev1.StorageClass) (*storagev1.StorageClass, error) {
```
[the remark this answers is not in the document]
davidz627: Alright I've done some stuff and now its a GoMock. PTAL.

davidz627:
/assign @ddebroy @leakingtapan 
ping @misterikkit 
this PR is ready for review

pkg/controller/controller_test.go
```diff
@@ -1390,48 +1391,6 @@ func TestProvision(t *testing.T) {
...
 }
 
-func TestProvisionWithMigration(t *testing.T) {
-	inTreePluginName := "kubernetes.io/gce-pd"
-	migrationDriverName := "pd.csi.storage.gke.io"
-	var requestBytes int64 = 100000
-
-	deletePolicy := v1.PersistentVolumeReclaimDelete
-	testcases := map[string]provisioningTestcase{
-		"should ignore in-tree with migration": {
```
davidz627: this test has been moved to the new `TestProvisionWithMigration` function and greatly simplified

pkg/controller/controller_test.go
```diff
@@ -2781,3 +2747,192 @@ func TestProvisionFromPVC(t *testing.T) {
...
+					Spec: v1.PersistentVolumeSpec{
+						PersistentVolumeSource: v1.PersistentVolumeSource{
+							CSI: &v1.CSIPersistentVolumeSource{
+								VolumeHandle: "mock-id",
+							},
+						},
+					},
+				}, nil).Times(1)
+		} else {
+			mockTranslator.EXPECT().IsPVMigratable(gomock.Any()).Return(false).AnyTimes()
```
ddebroy: When `tc.translateToCSI` is false, we expect one call to `IsPVMigratable`. Given that, will it make sense to have `.Times(1)` here (instead of `.AnyTimes()`)?

vendor/k8s.io/csi-translation-lib/plugins/openstack_cinder.go
```diff
@@ -54,6 +55,10 @@ func (t *osCinderCSITranslator) TranslateInTreeInlineVolumeToCSI(volume *v1.Volu
 
 	cinderSource := volume.Cinder
 	pv := &v1.PersistentVolume{
+		ObjectMeta: metav1.ObjectMeta{
+			// A.K.A InnerVolumeSpecName required to match for Unmount
+			Name: volume.Name,
+		},
```
misterikkit: Why is this in this PR?

pkg/controller/controller_test.go
```diff
@@ -2781,3 +2747,190 @@ func TestProvisionFromPVC(t *testing.T) {
...
+		t.Logf("running test: %s", tc.name)
+		// Set up test
+		tmpdir := tempDir(t)
+		defer os.RemoveAll(tmpdir)
+		mockController, driver, _, controllerServer, csiConn, err := createMockServer(t, tmpdir)
+		if err != nil {
+			t.Fatal(err)
+		}
+		mockTranslator := NewMockProvisionerCSITranslator(mockController)
+		defer mockController.Finish()
```
misterikkit: `defer` executes when the enclosing function returns, and not at the end of each loop iteration. Is that intended and desirable? (Mainly asking because I don't know how this mock API works.)

pkg/controller/controller_test.go
```diff
@@ -2781,3 +2747,190 @@ func TestProvisionFromPVC(t *testing.T) {
...
+		{
+			name:        "ignore in-tree PVC when provisioned by in-tree",
+			provisioner: inTreePluginName,
+			annotation:  map[string]string{annStorageProvisioner: inTreePluginName},
+			expectErr:   true,
+		},
+	}
+
+	for k, tc := range testcases {
+		t.Logf("running test: %s", tc.name)
```
misterikkit: It's preferred to use `t.Run` for table-driven tests. Then you can completely ignore the `k` value that you log in certain cases below. Any test log or test error will have the context of the subtest name.

pkg/controller/translator.mock.go
```diff
@@ -0,0 +1,95 @@
+// Automatically generated by MockGen. DO NOT EDIT!
+// Source: github.com/kubernetes-csi/external-provisioner/pkg/controller (interfaces: ProvisionerCSITranslator)
```
misterikkit: There are a couple things bugging me about this. Primarily,
- I don't know how to update this mock if I change the interface.
- There's no import statement referencing MockGen, so I don't know where it came from (beyond a google search)
- I don't know how someone is supposed to use this mock. Its own godoc isn't helping. I guess I should read [this](https://godoc.org/github.com/golang/mock/gomock)?

Update: I read it and it answered most of my questions.

pkg/controller/controller_test.go
```diff
@@ -2781,3 +2747,190 @@ func TestProvisionFromPVC(t *testing.T) {
...
+				&v1.PersistentVolume{
+					Spec: v1.PersistentVolumeSpec{
+						PersistentVolumeSource: v1.PersistentVolumeSource{
+							CSI: &v1.CSIPersistentVolumeSource{
+								VolumeHandle: "mock-id",
+							},
+						},
+					},
+				}, nil).Times(1)
+		}
```
misterikkit: It's possible with mock/expect libraries like this to overfit on the current implementation. Is the behavior described by these expectations really invariant? It is invalid to call `IsPVMigratable` twice from `Delete`? And `TranslateInTreePVToCSI` should only be called if `IsPVMigratable` returns true?

pkg/controller/controller_test.go
```diff
@@ -2781,3 +2747,190 @@ func TestProvisionFromPVC(t *testing.T) {
...
+		if tc.translateToCSI {
+			mockTranslator.EXPECT().TranslateInTreePVToCSI(gomock.Any()).Return(
+				&v1.PersistentVolume{
+					Spec: v1.PersistentVolumeSpec{
+						PersistentVolumeSource: v1.PersistentVolumeSource{
+							CSI: &v1.CSIPersistentVolumeSource{
+								VolumeHandle: "mock-id",
+							},
+						},
+					},
```
misterikkit: I would suggest moving this object definition into a variable or helper to improve readability. e.g. instead of this deeply nested struct literal, you could write `Return(pvWithCSISource)` or `Return(pvWithCSISource())`

This advice comes from https://abseil.io/tips/122

vendor/k8s.io/csi-translation-lib/translate.go
```diff
@@ -125,16 +136,20 @@ func GetInTreePluginNameFromSpec(pv *v1.PersistentVolume, vol *v1.Volume) (strin
 		}
 		return "", fmt.Errorf("could not find in-tree plugin name from persistent volume %v", pv)
 	} else if vol != nil {
-		// TODO(dyzz): Implement inline volume migration support
-		return "", errors.New("inline volume migration not yet supported")
+		for _, curPlugin := range inTreePlugins {
```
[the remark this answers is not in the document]
davidz627: these are all vendored in changes (bringing external-provisioner up to HEAD of the translation lib). Arguably this could be a seperate PR since it does actually introduce functional changes

vendor/k8s.io/csi-translation-lib/translate.go
```diff
@@ -35,9 +35,20 @@ var (
...
+// New creates a new CSITranslator which does real translation
+// for "in-tree plugins <-> csi drivers"
+func New() CSITranslator {
+	return CSITranslator{}
+}
+
 // TranslateInTreeStorageClassToCSI takes in-tree Storage Class
 // and translates it to a set of parameters consumable by CSI plugin
-func TranslateInTreeStorageClassToCSI(inTreePluginName string, sc *storage.StorageClass) (*storage.StorageClass, error) {
+func (ctl CSITranslator) TranslateInTreeStorageClassToCSI(inTreePluginName string, sc *storage.StorageClass) (*storage.StorageClass, error) {
```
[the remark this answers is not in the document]
davidz627: https://github.com/kubernetes/kubernetes/pull/83626

vendor/k8s.io/csi-translation-lib/plugins/openstack_cinder.go
davidz627: moved vendor update changes to seperate PR for cleanliness. Also to check for functional changes

pkg/controller/translator.mock.go
davidz627: ack. followed the style of other tests in this repo but agree it's not super clear. But also I learned it to write these tests and it didn't take long so I'm not too concerned about difficulty

pkg/controller/controller_test.go
davidz627: this is what I was worried about with this library in the first place. I definitely feel like I'm overfitting a little - IIUC that's kind of just how these libraries work... The intention is to not be perscriptive in the number of calls but rather to specify a specific return value so the test will continue to completion after doing translation.

> TranslateInTreePVToCSI should only be called if IsPVMigratable

This is invariant

> It is invalid to call IsPVMigratable twice from Delete

This is not.

Would it be more accurate to set this to `MinTimes`? Or do you have different guidance all-together. I'm not really too familiar with this type of testing so I don't really know how to avoid the overfitting problem since by nature we're asserting implementation details and not just relying on the input/output interfaces

pkg/controller/controller_test.go
davidz627: no I think it should be in the end of the loop iteration - nested it inside `t.Run` so that should work now

pkg/controller/util.go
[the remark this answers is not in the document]
misterikkit: If the fake is only consumed in `controller_test.go` I would not mind its implementation being in that file. If you'd prefer to keep it in a separate file, then `fake_translator_test.go` seems fine.

pkg/controller/controller_test.go
davidz627: Updated so that its not overfitting as much on the implementation by using `AnyTimes` on implementation details (still similar semantics but now the main purpose of those mock calls is to set return value expectations).

We are doing extra strict checks on the "interfaces" of the function - inputs -> outputs (in this case both the function returns AND the inputs to the CSI calls)

pkg/controller/controller_test.go
```diff
@@ -1471,6 +1431,7 @@ func runProvisionTest(t *testing.T, k string, tc provisioningTestcase, requested
 	if err != nil {
 		t.Fatal(err)
 	}
+	mockTranslator := NewMockProvisionerCSITranslator(mockController)
```
misterikkit: Does this test do anything with `mockTranslator`? If not, can we pass in a real translator? (assuming the test passes fine with it.)

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+						},
+					}, nil).Times(1)
+			}
+
+			// Make a Provision call
+			volOpts := controller.ProvisionOptions{
+				StorageClass: &storagev1.StorageClass{
+					Provisioner:   tc.provisioner,
+					Parameters:    map[string]string{"fstype": "ext3"},
+					ReclaimPolicy: &deletePolicy,
```
misterikkit: Is this one of those APIs where you can't take the address of a literal, so you need to either declare the value beforehand or wrap it in a function that returns a pointer? :cry:

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Set up test
+			tmpdir := tempDir(t)
+			defer os.RemoveAll(tmpdir)
+			mockController, driver, _, controllerServer, csiConn, err := createMockServer(t, tmpdir)
+			if err != nil {
```
misterikkit: Note for future refactors. A helper like `createMockServer` that takes a `*testing.T` may as well call `t.Fatal` for us rather than returning an error.

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+				t.Fatal(err)
+			}
+			mockTranslator := NewMockProvisionerCSITranslator(mockController)
+			defer mockController.Finish()
+			defer driver.Stop()
```
misterikkit: Does ordering matter between `Finish()` and `Stop()`?

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+				inTreePluginName, false, mockTranslator)
+
+			// Set up return values (AnyTimes to avoid overfitting on implementation)
+			if tc.expectTranslation {
+				// Have fake translation provide same as input SC but with
+				// Parameter indicating it has been translated
+				mockTranslator.EXPECT().TranslateInTreeStorageClassToCSI(gomock.Any(), gomock.Any()).DoAndReturn(
+					func(inTreePluginName string, sc *storagev1.StorageClass) (*storagev1.StorageClass, error) {
+						newSC := sc.DeepCopy()
+						newSC.Parameters[translatedAnnKey] = "foo"
```
misterikkit: Neat to use Parameters instead of Annotations. I'd suggest renaming `translatedAnnKey` to just `translatedKey`

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+			defer mockController.Finish()
+			defer driver.Stop()
+			clientSet := fakeclientset.NewSimpleClientset()
+			pluginCaps, controllerCaps := provisionCapabilities()
+			csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner",
+				"test", 5, csiConn.conn, nil, driverName, pluginCaps, controllerCaps,
+				inTreePluginName, false, mockTranslator)
+
+			// Set up return values (AnyTimes to avoid overfitting on implementation)
+			if tc.expectTranslation {
```
misterikkit: If we don't expect translation but it accidentally gets called, we still want "foo" set on the objects so that we can assert whether it happened.

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+							newPV.Annotations = map[string]string{}
+						}
+						newPV.Annotations[translatedAnnKey] = "foo"
+						return newPV, nil
+					},
+				).AnyTimes()
+			}
+
+			// Set an expectation that the Create should be called
+			if tc.expectCreate {
```
misterikkit: Does `mockController.Finish()` assert functions are called zero times when you don't set any expectation? In other words, what would happen if we didn't expect create but it did get called?

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+				}
+				return
+			}
+
+			if controller.ProvisioningFinished != state {
+				t.Errorf("expected ProvisioningState %s, got %s", controller.ProvisioningFinished, state)
+			}
+
+			if tc.expectTranslation {
+				if _, ok := pv.Annotations[translatedAnnKey]; !ok {
```
misterikkit: how about just `if "foo" == pv.Annotations[translatedAnnKey]` ?

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+					t.Errorf("got error: %v", err)
+				}
+				return
+			}
+
+			if controller.ProvisioningFinished != state {
+				t.Errorf("expected ProvisioningState %s, got %s", controller.ProvisioningFinished, state)
+			}
+
+			if tc.expectTranslation {
```
misterikkit: Should we also assert the reverse?

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+				if _, ok := pv.Annotations[translatedAnnKey]; !ok {
+					t.Errorf("got no translated annotation %s on the pv, expected one", translatedAnnKey)
```
misterikkit: nit about the error message: What would be most useful in a test failure log? "expected this annotation", or "expected PV to be translated to in-tree"?

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+				}
+			}
+		})
+
+	}
+}
+
+func createFakePVCWithAnnotation(ann map[string]string, requestBytes int64) *v1.PersistentVolumeClaim {
```
misterikkit: nit: this isn't a fake. maybe call it `createPVCWithAnnotation`?

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+				Requests: v1.ResourceList{
+					v1.ResourceName(v1.ResourceStorage): resource.MustParse(strconv.FormatInt(requestBytes, 10)),
+				},
+			},
+		},
+	}
+}
+
+func TestDeleteMigration(t *testing.T) {
+	var (
```
misterikkit: nit: These could be const

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+		translatedHandle = "translated-handle"
+		normalHandle     = "no-translation-handle"
+	)
+
+	testCases := []struct {
+		name           string
+		pv             *v1.PersistentVolume
+		translateToCSI bool
```
misterikkit: nit: would it make sense to rename this to `expectTranslation` as per the previous test?

pkg/controller/controller_test.go
```diff
@@ -3156,3 +3129,257 @@ func TestProvisionFromPVC(t *testing.T) {
...
+					VolumeId: volID,
+				}).Return(&csi.DeleteVolumeResponse{}, nil).Times(1)
+
+			// Run Delete
+			err = csiProvisioner.Delete(tc.pv)
+			if tc.expectErr && err == nil {
+				t.Error("Expected error, got none")
+			}
+			if !tc.expectErr && err != nil {
+				t.Errorf(" got error: %v", err)
```
misterikkit: nit: leading whitespace

pkg/controller/controller_test.go
davidz627: nope, changing all other tests to use real translator

pkg/controller/controller_test.go
davidz627: yeah

pkg/controller/controller_test.go
davidz627: ack

pkg/controller/controller_test.go
davidz627: I'm not sure - this is how every other test does it.

pkg/controller/controller_test.go
davidz627: yes, if no expectation is set for create the test will fail if the function is called

pkg/controller/controller_test.go
davidz627: don't want to make another const just for "foo" and also don't want to have magic strings. The important thing is the existence of the `translatedKey` at all so I think I'll keep it as such. It's a little less pretty but I think less "distracting"

PR #341, opened 2019-09-12
Ignore provisioning for in-tree volumes that have not been migrated

davidz627:
/kind bug

The provisioner just picks up any in-tree PVC that is related to the CSI Driver and treats it as a "migrated" PVC and continues to provision it. In reality it should use the annotation as a signal on whether the in-tree pv-controller has stood down or not (migrated or not).

/assign @jsafrane @msau42 @ddebroy 
/cc @leakingtapan

Fixes #340

```release-note
Fixes issue where provisioner provisions volumes for in-tree PVC's which have not been migrated
```

PR #345, opened 2019-09-13
Ignore non-migrated in-tree PVC when provisioning

davidz627:
Cherry-pick of: https://github.com/kubernetes-csi/external-provisioner/pull/341

/kind bug
/assign @jsafrane @msau42 
/cc @ddebroy @leakingtapan 

```release-note
Fixes issue where provisioner provisions volumes for in-tree PVC's which have not been migrated
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -208,7 +208,7 @@ func main() {
 		} else if *leaderElectionType == "leases" {
 			le = leaderelection.NewLeaderElection(clientset, lockName, run)
 		} else {
-			klog.Error("--leader-election-type must be either 'endpoints' or 'lease'")
+			klog.Error("--leader-election-type must be either 'endpoints' or 'leases'")
 			os.Exit(1)
 		}
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -113,6 +113,8 @@ const (
 	tokenPVNameKey       = "pv.name"
 	tokenPVCNameKey      = "pvc.name"
 	tokenPVCNameSpaceKey = "pvc.namespace"
+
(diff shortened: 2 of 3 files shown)


msau42:
@pohly do you know why test-subtree is failing in travis but not prow?

I'm inclined to just disable travis PR on all our repos. prow unit job should have all the same functionality right?

msau42:
I disabled travis PR testing, but we may have to manually merge this one

pohly:
> @pohly do you know why test-subtree is failing in travis but not prow?

That was fun to investigate. I couldn't reproduce it initially, even after compiling the same git version that TravisCI uses. But then the `grafted` comment in the `git log` output made me suspicious: that commit is a normal merge, not a graft commit. After cloning with `--depth=50` I could reproduce the problem.

Shallow clones are problematic. For example, `git describe --tags` also can fail because the relevant tag is too far in the past and wasn't pulled.

We can:
- disable PR testing with TravisCI (but then building the branch and thus pushing images may still fail?)
- do full git clones in TravisCI (https://docs.travis-ci.com/user/customizing-the-build/#git-clone-depth)
- replace `--no-merges` with `--max-parents=1 --min-parents=1`, which will then ignore both merge commits and this incomplete commit
- add one more fake commit to the branch that gets merged when we hit the problem: I think it only triggers if a merge commit is exactly at depth=50 from the root.

The last option would be the fastest and doesn't depend on updating csi-release-tools.

msau42:
Hmm having image tag build potentially fail is not great either. I'm going to re-enable travis and let's try the method to add one more commit to this pr?

msau42:
Once we are able to migrate image releasing to prow, then we can disable travis altogether

davidz627:
@msau42 @jsafrane fixed travis with a dummy commit

msau42:
/lgtm
I can't believe that worked

PR #349, opened 2019-09-28
Fix typo for leader-election-type

leakingtapan:
**What type of PR is this?**
/kind bug

Fix typo for leader-election-type

Fixes #

```release-note

```


davidz627:
I don't think this is a typo really but the fix does make it more consistent.
/lgtm

PR #350, opened 2019-09-30
Don't process PVCs for in-tree volumes

jsafrane:
If I understand [code in `ProvisionExt`](https://github.com/kubernetes-csi/external-provisioner/blob/3184db61ab93bbe588fff4a223bf23c49a5d5937/pkg/controller/controller.go#L374), PVCs that require migrated in-tree provisioner in their StorageClass have their `volume.beta.kubernetes.io/storage-provisioner` annotation set to **CSI driver name**.

Therefore we do not need the provisioner library to listen for PVCs with in-tree provisioner at all and save some (little) CPU in unrelated PVC processing. I keep the check in `ProvisionExt`, just in case.

@ddebroy, PTAL carefuly.

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -187,6 +187,7 @@ type csiProvisioner struct {
 var _ controller.Provisioner = &csiProvisioner{}
 var _ controller.BlockProvisioner = &csiProvisioner{}
 var _ controller.ProvisionerExt = &csiProvisioner{}
+var _ controller.Qualifier = &csiProvisioner{}
 
 var (
 	// Each provisioner have a identify string to distinguish with others. This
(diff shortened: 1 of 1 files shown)


msau42:
/assign @ddebroy @davidz627 
Will this work or does lib still need the option to check `storageclass.provisioner`?

ddebroy:
The additionalProvisionerNames stuff is necessary to keep this check at https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/master/controller/controller.go#L1234 happy for migration scenarios.

Without populating additionalProvisionerNames, we end up with:
```
controller.go:1235] provision "default/podpvc" class "standard": unknown provisioner 
"kubernetes.io/gce-pd" requested in claim's StorageClass
```

We discussed this a bit at https://github.com/kubernetes-csi/external-provisioner/pull/341#issuecomment-531025966

jsafrane:
I see. I reworked the PR to still pass additional provisioners to the lib, but rejecting PVCs in `ShouldProvision`. PTAL.

jsafrane:
> var _ controller.Qualifier = &csiProvisioner{}


Good idea, added.

PR #356, opened 2019-10-08
1.1:  Fix snapshotRef checks

jsafrane:
**What type of PR is this?**
/kind bug

Fixed checking of binding of PVC and Snapshots when creating volumes.

```release-note
Added extra verification of source Snapshot before provisioning.
```


jsafrane:
/close
we don't support 1.1

PR #360, opened 2019-10-08
Update 1.3 changelog for 1.3.1

msau42:
**What type of PR is this?**
/kind documentation

Fixes #

```release-note
NONE
```

### DIFF
--- a/CHANGELOG-1.3.md
+++ b/CHANGELOG-1.3.md
@@ -1,3 +1,13 @@
+# Changelog since v1.3.0
+
+## Bug Fixes
+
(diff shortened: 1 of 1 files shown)


msau42:
@pohly I think I'm hitting the same issue as https://github.com/kubernetes-csi/external-provisioner/pull/345#issuecomment-533437383 again.  It was worked around in the other PR by adding a no-op commit, but that doesn't seem to have completely solved the problem. Do you think we need to configure travis to do full git clone to actually fix the issue?

msau42:
Rebased to pick up prow fixes

PR #368, opened 2019-10-09
Fix potential nil pointer

cwdsuzhou:
**What type of PR is this?**

/kind bug

Fix potential nil pointer

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -424,7 +424,8 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 
 		switch options.PVC.Spec.DataSource.Kind {
 		case snapshotKind:
-			if *(options.PVC.Spec.DataSource.APIGroup) != snapshotAPIGroup {
+			if options.PVC.Spec.DataSource.APIGroup == nil ||
+				*(options.PVC.Spec.DataSource.APIGroup) != snapshotAPIGroup {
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -424,7 +424,8 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 
 		switch options.PVC.Spec.DataSource.Kind {
 		case snapshotKind:
-			if *(options.PVC.Spec.DataSource.APIGroup) != snapshotAPIGroup {
+			if options.PVC.Spec.DataSource.APIGroup == nil ||
+				*(options.PVC.Spec.DataSource.APIGroup) != snapshotAPIGroup {
 				return nil, controller.ProvisioningFinished, fmt.Errorf("the PVC source does not belong to the right APIGroup. Expected %s, Got %s", snapshotAPIGroup, *(options.PVC.Spec.DataSource.APIGroup))
```
jsafrane: you may dereference nil here

jsafrane:
Were you able to actually hit this nil dereference or are you just reading code? I am very interested in reproducer, because something must have gone terribly wrong in validation and it should be fixed also there.

cwdsuzhou:
> Were you able to actually hit this nil dereference or are you just reading code? I am very interested in reproducer, because something must have gone terribly wrong in validation and it should be fixed also there.

Thanks, I found it has been validated at apiserver.  

I am just reading the codes

/close

PR #383, opened 2019-11-28
Use v1.CSINode in the external-provisioner

### LINKED ISSUE #385: Update to use v1.CSINode
This change will require a major version bump because we would have to bump the min K8s version to 1.17 when the object was introduced.

We should make this change at least 1, if not 2 releases before the v1beta gets removed from Kubernetes.

/lifecycle frozen

bertinatto:
**What type of PR is this?**

/kind feature

This PR updates the external provisioner to use the GA version of the CSINode object.

Fixes #385

```release-note
Use GA version of CSINode object.
```

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -2,4 +2,26 @@
 
 . release-tools/prow.sh
 
+# This check assumes that the current configuration uses a driver deployment
+# which has been updated to use v1 APIs that aren't available in Kubernetes < 1.17.
+# TODO: The check can be removed when all Prow jobs for Kubernetes < 1.17 are removed.
+if ! (version_gt "${CSI_PROW_KUBERNETES_VERSION}" "1.16.255" || [ "${CSI_PROW_KUBERNETES_VERSION}" == "latest" ]); then
+    filtered_tests=
+    skipped_tests=
+    for test in ${CSI_PROW_TESTS}; do
+	case "$test" in
+	    parallel | parallel | serial | parallel-alpha | serial-alpha)
+		skipped_tests="$skipped_tests $test"
+		;;
+	    *)
+		filtered_tests="$filtered_tests $test"
+		;;
+	esac
+    done
+    if [ "$skipped_tests" ]; then
+	info "Testing on Kubernetes ${CSI_PROW_KUBERNETES_VERSION} is no longer supported. Skipping CSI_PROW_TESTS: $skipped_tests."
+	CSI_PROW_TESTS="$filtered_tests"
+    fi
+fi
+
 main
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -45,7 +45,7 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
 	v1 "k8s.io/client-go/listers/core/v1"
-	storagelistersv1beta1 "k8s.io/client-go/listers/storage/v1beta1"
+	storagelistersv1 "k8s.io/client-go/listers/storage/v1"
 	utilflag "k8s.io/component-base/cli/flag"
 	csitrans "k8s.io/csi-translation-lib"
 )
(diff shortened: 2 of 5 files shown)


bertinatto:
This provisioner depends on the snapshotter, which needs to be updated first. @yuxiangqian, please let me know one you update the client-go dep in the snapshotter.

bertinatto:
> This provisioner depends on the snapshotter, which needs to be updated first. @yuxiangqian, please let me know one you update the client-go dep in the snapshotter.

@yuxiangqian I hope you don't mind I created https://github.com/kubernetes-csi/external-snapshotter/pull/202 to address that. PTAL if you can.

msau42:
Can you try using override in the meantime?

bertinatto:
> Can you try using override in the meantime?

Do you mean pinning the *snapshotter*? If I use the lastest snapshotter release, go mod complains:

```
# github.com/kubernetes-csi/external-snapshotter/pkg/client/clientset/versioned/typed/volumesnapshot/v1beta1
/home/fjb/pkg/mod/github.com/kubernetes-csi/external-snapshotter@v2.0.0-rc1.0.20191024235139-bc6e42db5bc6+incompatible/pkg/client/clientset/versioned/typed/volumesnapshot/v1beta1/volumesnapshot_client.go:84:32: undefined: serializer.DirectCodecFactory
```

Just to clarify, to use the new `v1.CSINode`, we need to update `go-client`. But doing that will also update the `api` and `apimachinery` deps (otherwise it won't compile). The problem is that the current version of the snaphotter (which the provisioner depends on), depends on a previous version of `go-client` (which we're trying to get rid of in this PR).

msau42:
I meant overriding all the k8s.io/* versions. It looks with go mod, you can do this by modifying the "replace" statement:
https://stackoverflow.com/questions/57266010/how-to-override-a-dependency-in-go-modules

pohly:
> I meant overriding all the k8s.io/* versions. 

We have script for that: https://github.com/kubernetes-csi/external-provisioner/blob/master/release-tools/go-get-kubernetes.sh

yuxiangqian:
> > This provisioner depends on the snapshotter, which needs to be updated first. @yuxiangqian, please let me know one you update the client-go dep in the snapshotter.
> 
> @yuxiangqian I hope you don't mind I created [kubernetes-csi/external-snapshotter#202](https://github.com/kubernetes-csi/external-snapshotter/pull/202) to address that. PTAL if you can.

I have been working on this PR: https://github.com/kubernetes-csi/external-snapshotter/pull/203
CI is failing, will continue tomorrow.

bertinatto:
> > I meant overriding all the k8s.io/* versions.
> 
> We have script for that: https://github.com/kubernetes-csi/external-provisioner/blob/master/release-tools/go-get-kubernetes.sh

Yeah, the script is already pinning (replacing) the k8s.io/* deps

pohly:
> Yeah, the script is already pinning (replacing) the k8s.io/* deps

What I meant is that it can also be used to update those replacements with a newer Kubernetes.

yuxiangqian:
@bertinatto shamelessly stole your original PR as well to resolve the code-generator issue. Its now ready for review. I will update once its merged. PTAL if it solves your problem as well.

bertinatto:
/hold

Checking how to support older k8s releases

yuxiangqian:
@bertinatto [PR](https://github.com/kubernetes-csi/external-snapshotter/pull/203) merged.

msau42:
The next upcoming release is going to be a 2.0. So we can continue with this PR and remove v1beta support.

bertinatto:
`pull-kubernetes-csi-external-provisioner-1-16-on-kubernetes-1-16 ` fails because k8s 1.16 doesn't know `v1.CSINode` object:

 ```
I0421 09:36:24.962471       1 reflector.go:188] Listing and watching *v1.CSINode from k8s.io/client-go/informers/factory.go:135
E0421 09:36:24.964109       1 reflector.go:156] k8s.io/client-go/informers/factory.go:135: Failed to list *v1.CSINode: the server could not find the requested resource
```

msau42:
The simplest way I see around this is to add some logic to `.prow.sh` to skip the job if kubernetes version is < 1.17

msau42:
whoops accidentally clicked the close button

pohly:
Michelle Au <notifications@github.com> writes:
> The simplest way I see around this is to add some logic to `.prow.sh`
> to skip the job if kubernetes version is < 1.17

Agreed. It was part of the original design that the jobs might be
defined in Prow "just in case" and then the components themselves decide
that they have nothing to test.

The job then may still be needed for some other branch of the component.

bertinatto:
@pohly @msau42 
I used the code suggested by @pohly to skip the v1.16 job. PTAL

msau42:
/lgtm
/approve

Thanks!

PR #388, opened 2019-12-24
Add prometheus metrics to CSI external-provisioner using new csi-lib-utils library

saad-ali:
**What type of PR is this?**

/kind feature

This PR adds metrics to the external-provisioner CSI gRPC calls:

* `csi_sidecar_operations_seconds` [ALPHA] Container Storage Interface operation duration with gRPC error code status total

Fixes #

```release-note
Add prometheus metrics to CSI external-provisioner 
```


cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -68,6 +69,9 @@ var (
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "Namespace where the leader election resource lives. Defaults to the pod namespace if not set.")
 	strictTopology          = flag.Bool("strict-topology", false, "Passes only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.")
 
+	metricsAddress = flag.String("metrics-address", "", "The TCP network address address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
```
msau42: Can you also update README.md with these new options?

go.mod
```diff
@@ -4,75 +4,68 @@ go 1.12
...
+	k8s.io/client-go v0.17.0
+	k8s.io/component-base v0.17.0
+	k8s.io/csi-translation-lib v0.17.0
 	k8s.io/klog v1.0.0
 	k8s.io/kubernetes v1.14.0
 	sigs.k8s.io/sig-storage-lib-external-provisioner v4.0.1+incompatible
 )
 
-replace k8s.io/api => k8s.io/api v0.0.0-20190918195907-bd6ac527cfd2
+replace k8s.io/api => k8s.io/api v0.17.0
```
msau42: Were these forced overrides or automatic?  I'm wondering if we can actually remove most of these replace statements now that most of the repos have switched to semantic versioning.

cmd/csi-provisioner/csi-provisioner.go
saad-ali: Will do.

go.mod
saad-ali: Automatic via go-get-kubernetes.sh
I'll do some clean up.

saad-ali:
Feedback addressed. PTAL

README.md
```diff
@@ -61,6 +61,10 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--worker-threads <num>`: Number of simultaneously running `ControllerCreateVolume` and `ControllerDeleteVolume` operations. Default value is `100`.
 
+* `metrics-address`: The TCP network address address where the prometheus metrics endpoint will run (example: `:8080` which corresponds to port 8080 on local host). The default is empty string, which means metrics endpoint is disabled.
```
msau42: Add "--" similar to the other options

saad-ali: Fixed

saad-ali:
Feedback addressed PTAL

saad-ali:
Rebased to official 0.7.0 csi-lib-utils release.

go.mod
pohly: > I'm wondering if we can actually remove most of these replace statements now that most of the repos have switched to semantic versioning.

We can't, because the semantic versioning for Kubernetes components is too weak: `go get -u ./...` for updating to the latest "minor and patch upgrades" of other dependencies would also update all Kubernetes components if we removed the replace statements. This is bad because there's no guarantee that 0.18.0 is really API compatible with 0.17.0. I think it's better to update Kubernetes via the script and use `go get -u ./...` for the rest.

verult:
@mattcary @saikat-royc

PR #393, opened 2020-01-03
issue-384, A single "default" storage class secret is added

### LINKED ISSUE #384: A single "default" storage class secret
## Current Issue

Currently storage classes have each operation have a reference to the secret:

```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: fast-storage-all
provisioner: csi-driver.team.example.com
parameters:
  type: pd-ssd
  csi.storage.k8s.io/provisioner-secret-name: ${pvc.name}
  csi.storage.k8s.io/provisioner-secret-namespace: ${pvc.namespace}-fast-storage
  csi.storage.k8s.io/node-publish-secret-name: ${pvc.name}-${pvc.annotations['team.example.com/key']}
  csi.storage.k8s.io/node-publish-secret-namespace: ${pvc.namespace}-fast-storage
```

See [Docs](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html#multiple-operation-secrets)

## Proposal 

To have a default that can be overridden by the specific operation secret references, that way as new operations are supported, we do not need to have a large set of repeating information for each operation. The new storageClass would look like this:

```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: fast-storage-all
provisioner: csi-driver.team.example.com
parameters:
  type: pd-ssd
  csi.storage.k8s.io/secret-name: ${pvc.name}
  csi.storage.k8s.io/secret-namespace: ${pvc.namespace}-fast-storage
  csi.storage.k8s.io/node-publish-secret-name: ${pvc.name}-${pvc.annotations['team.example.com/key']}
  csi.storage.k8s.io/node-publish-secret-namespace: ${pvc.namespace}-fast-storage
``` 

Where

```
  csi.storage.k8s.io/secret-name: ${pvc.name}
  csi.storage.k8s.io/secret-namespace: ${pvc.namespace}-fast-storage
```

Are the new _default_ secrets if none are specified.

taaraora:
**What type of PR is this?**
/kind feature

Fixes #384

```release-note
Default StorageClass secrets are added:
 csi.storage.k8s.io/secret-name: ${pvc.name}
 csi.storage.k8s.io/secret-namespace: ${pvc.namespace}

Note: Default secrets for storage class feature does work only when both parameters are added.
```

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -213,6 +213,8 @@ func TestStripPrefixedCSIParams(t *testing.T) {
 				prefixedNodePublishSecretNamespaceKey:       "csiBar",
 				prefixedControllerExpandSecretNameKey:       "csiBar",
 				prefixedControllerExpandSecretNamespaceKey:  "csiBar",
+				prefixedDefaultSecretNameKey:                "csiBar",
+				prefixedDefaultSecretNamespaceKey:           "csiBar",
 			},
 			expectedParams: map[string]string{},
 		},
@@ -475,13 +477,17 @@ func provisionFromPVCCapabilities() (rpc.PluginCapabilitySet, rpc.ControllerCapa
 		}
 }
 
-// Minimal PVC required for tests to function
-func createFakePVC(requestBytes int64) *v1.PersistentVolumeClaim {
+func createFakeNamedPVC(requestBytes int64, name string, userAnnotations map[string]string) *v1.PersistentVolumeClaim {
+	annotations := map[string]string{annStorageProvisioner: driverName}
+	for k, v := range userAnnotations {
+		annotations[k] = v
+	}
+
 	return &v1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			UID:         "testid",
-			Name:        "fake-pvc",
-			Annotations: map[string]string{annStorageProvisioner: driverName},
+			Name:        name,
+			Annotations: annotations,
 		},
 		Spec: v1.PersistentVolumeClaimSpec{
 			Selector: nil, // Provisioner doesn't support selector
@@ -492,6 +498,11 @@ func createFakePVC(requestBytes int64) *v1.PersistentVolumeClaim {
 			},
 		},
 	}
+}
+
+// Minimal PVC required for tests to function
+func createFakePVC(requestBytes int64) *v1.PersistentVolumeClaim {
+	return createFakeNamedPVC(requestBytes, "fake-pvc", nil)
 }
 
 // createFakePVCWithVolumeMode returns PVC with VolumeMode
@@ -778,7 +789,7 @@ type provisioningTestcase struct {
 	volWithLessCap    bool
 	volWithZeroCap    bool
 	expectedPVSpec    *pvSpec
-	withSecretRefs    bool
+	clientSetObjects  []runtime.Object
 	createVolumeError error
 	expectErr         bool
 	expectState       controller.ProvisioningState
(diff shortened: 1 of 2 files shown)


msau42:
/assign @misterikkit 
to help review

pkg/controller/controller_test.go
```diff
@@ -213,6 +213,8 @@ func TestStripPrefixedCSIParams(t *testing.T) {
 				prefixedNodePublishSecretNamespaceKey:       "csiBar",
 				prefixedControllerExpandSecretNameKey:       "csiBar",
 				prefixedControllerExpandSecretNamespaceKey:  "csiBar",
+				prefixedDefaultSecretNameKey:                "csiBar",
+				prefixedDefaultSecretNamespaceKey:           "csiBar",
```
misterikkit: Does this change cover the new logic, or just fix an existing test? I think we should add a case that gets the default value.

taaraora: This was just fix of existing test. I've added a few new tests for default secret.

taaraora:
> Can you add some unit tests that use the new default secret?
> 
> Another thing to keep in mind, allowable secret templates are different for provisioning compared to the other operations. We should make sure we can maintain that behavior for the default secret.

- I've added tests for default secrets.
- Yep, default secrets support only those template variables which are common for all call templates. We definitely should describe this behavior in the documentation.

pkg/controller/controller_test.go
```diff
@@ -1229,6 +1382,31 @@ func TestProvision(t *testing.T) {
 			expectErr:         true,
 			expectState:       controller.ProvisioningNoChange,
 		},
+		"fail to get secret reference for default secret parameter template with annotation": {
```
misterikkit: Is the failure caused by mismatch between `secret-from-annotation` and `default-secret-from-annotation`? If so, could you make the names more distinct? Thanks.

pkg/controller/controller_test.go
```diff
@@ -477,13 +477,18 @@ func provisionFromPVCCapabilities() (rpc.PluginCapabilitySet, rpc.ControllerCapa
 		}
 }
 
-// Minimal PVC required for tests to function
-func createFakePVC(requestBytes int64) *v1.PersistentVolumeClaim {
+// createFakeNamedPVC creates PVC for tests, should be extended with new args when needed
+func createFakeNamedPVC(requestBytes int64, name string, userAnnotations ...struct{ Key, Value string }) *v1.PersistentVolumeClaim {
```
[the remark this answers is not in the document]
taaraora: Yes, the current signature makes `userAnnotations` argument optional. If we change it to map it will require to add an empty map to each place where this function is invoked now.

pkg/controller/controller_test.go
taaraora: As @msau42 pointed out the default secrets can support only that set of template variables which is common for all calls. In this test, I showed that the annotation variable is not supported by the default secret. Should I rename this test?

pkg/controller/controller_test.go
```diff
@@ -477,13 +477,18 @@ func provisionFromPVCCapabilities() (rpc.PluginCapabilitySet, rpc.ControllerCapa
 		}
 }
 
-// Minimal PVC required for tests to function
-func createFakePVC(requestBytes int64) *v1.PersistentVolumeClaim {
+// createFakeNamedPVC creates PVC for tests, should be extended with new args when needed
```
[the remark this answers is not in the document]
taaraora: Thank you, that makes sense, I can just delete it.

pkg/controller/controller_test.go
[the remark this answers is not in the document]
misterikkit: Seems like not too many call sites? Alternatively, since you only use `userAnnotations` in one place, you could make it a different helper function with a longer, more specific name.

The most comprehensive solution would be to adopt the standard "function options" pattern, but that may be overkill.

pkg/controller/controller_test.go
misterikkit: I think after that explanation, the test name makes sense. I just find the existing logic with secret references and template evaluation to be confusing. :cry:

misterikkit: Now that I re-read the test, maybe we can update the name to mention invalid/illegal template variable.

pkg/controller/controller_test.go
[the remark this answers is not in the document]
taaraora: Hm, I agree with you, a different helper function is the better way to handle this.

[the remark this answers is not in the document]
taaraora: Fixed, it is `map[string]string` now.

pkg/controller/controller_test.go
taaraora: Fixed.

msau42:
/approve
Thank you!

PR #399, opened 2020-01-17
add pvc metadata to createvolume req

### LINKED ISSUE #370: Pass k8s API metadata through parameters in CreateVolume
/kind feature

There is a use case for passing in k8s API metadata through parameters in CreateVolume. For example this information could be used by drivers to label underlying resources and associate them with k8s objects.

This could be done through a reserved parameter namespace e.g. `kubernetes.io/pvc.name`, `kubernetes.io/pvc.namespace`
It could be enabled by a flag so that it is not a breaking change (since some drivers may do strict validation on their own well-known parameters).

/cc @jsafrane @msau42

zetsub0u:
Co-authored-by: abrden <agustina.barbetta@gmail.com>

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Add PVC metadata to plugin call in the CreateVolume operation for it to be able to query for more information if needed during volume creation.

Fixes #370 

```release-note
* Adds --extraCreateMetadata flag which, when enabled, will inject parameters onto CreateVolume driver requests with PVC and PV metadata.
  Injected keys:
  - csi.storage.k8s.io/pvc/name
  - csi.storage.k8s.io/pvc/namespace
  - csi.storage.k8s.io/pv/name
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -68,6 +68,7 @@ var (
 	leaderElectionType      = flag.String("leader-election-type", "endpoints", "the type of leader election, options are 'endpoints' (default) or 'leases' (strongly recommended). The 'endpoints' option is deprecated in favor of 'leases'.")
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "Namespace where the leader election resource lives. Defaults to the pod namespace if not set.")
 	strictTopology          = flag.Bool("strict-topology", false, "Passes only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.")
+	extraCreateMetadata     = flag.Bool("extra-create-metadata", false, "If set, add pv/pvc metadata to plugin create requests as parameters.")
 
 	metricsAddress = flag.String("metrics-address", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
 	metricsPath    = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
@@ -221,7 +222,9 @@ func main() {
 		translator,
 		scLister,
 		csiNodeLister,
-		nodeLister)
+		nodeLister,
+		*extraCreateMetadata,
+	)
 
 	provisionController = controller.NewProvisionController(
 		clientset,
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -104,6 +104,11 @@ const (
 	nodePublishSecretNameKey      = "csiNodePublishSecretName"
 	nodePublishSecretNamespaceKey = "csiNodePublishSecretNamespace"
 
+	// PV and PVC metadata, used for sending to drivers in the  create requests, added as parameters, optional.
+	pvcNameKey      = "csi.storage.k8s.io/pvc/name"
+	pvcNamespaceKey = "csi.storage.k8s.io/pvc/namespace"
+	pvNameKey       = "csi.storage.k8s.io/pv/name"
+
 	// Defines parameters for ExponentialBackoff used for executing
 	// CSI CreateVolume API call, it gives approx 4 minutes for the CSI
 	// driver to complete a volume creation.
@@ -214,6 +219,7 @@ type csiProvisioner struct {
 	scLister                              storagelistersv1.StorageClassLister
 	csiNodeLister                         storagelistersv1beta1.CSINodeLister
 	nodeLister                            corelisters.NodeLister
+	extraCreateMetadata                   bool
 }
 
 var _ controller.Provisioner = &csiProvisioner{}
@@ -276,7 +282,9 @@ func NewCSIProvisioner(client kubernetes.Interface,
 	translator ProvisionerCSITranslator,
 	scLister storagelistersv1.StorageClassLister,
 	csiNodeLister storagelistersv1beta1.CSINodeLister,
-	nodeLister corelisters.NodeLister) controller.Provisioner {
+	nodeLister corelisters.NodeLister,
+	extraCreateMetadata bool,
+) controller.Provisioner {
 
 	csiClient := csi.NewControllerClient(grpcClient)
 	provisioner := &csiProvisioner{
@@ -297,6 +305,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 		scLister:                              scLister,
 		csiNodeLister:                         csiNodeLister,
 		nodeLister:                            nodeLister,
+		extraCreateMetadata:                   extraCreateMetadata,
 	}
 	return provisioner
 }
@@ -574,6 +583,13 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 	req.Parameters, err = removePrefixedParameters(options.StorageClass.Parameters)
 	if err != nil {
 		return nil, controller.ProvisioningFinished, fmt.Errorf("failed to strip CSI Parameters of prefixed keys: %v", err)
+	}
+
+	if p.extraCreateMetadata {
+		// add pvc and pv metadata to request for use by the plugin
+		req.Parameters[pvcNameKey] = options.PVC.GetName()
+		req.Parameters[pvcNamespaceKey] = options.PVC.GetNamespace()
+		req.Parameters[pvNameKey] = pvName
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--metrics-path`: The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.
 
+* `--extraCreateMetadata`: Enables the injection of extra PVC and PV metadata as parameters when calling `CreateVolume` on the driver (keys: "csi.storage.k8s.io/pvc/name", "csi.storage.k8s.io/pvc/namespace", "csi.storage.k8s.io/pv/name") 
+
 #### Other recognized arguments
 * `--feature-gates <gates>`: A set of comma separated `<feature-name>=<true|false>` pairs that describe feature gates for alpha/experimental features. See [list of features](#feature-status) or `--help` output for list of recognized features. Example: `--feature-gates Topology=true` to enable Topology feature that's disabled by default.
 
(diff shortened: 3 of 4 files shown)


davidz627:
@zetsub0u please sign the CLA
/ok-to-test

zetsub0u:
Hi, yes i followed the steps but i'm not getting the document to sign via email and i'm trying to figure out what to do now. Also i'll check the unittests, i completely forgot to run them as it was a simple change.

zetsub0u:
CLA signed, i guess need to wait for the bot to retest.

pkg/controller/controller.go
```diff
@@ -101,6 +101,11 @@ const (
 	nodePublishSecretNameKey      = "csiNodePublishSecretName"
 	nodePublishSecretNamespaceKey = "csiNodePublishSecretNamespace"
 
+	// PVC's name and namespace passed on create requests, useful for the plugin to query for extra data
+	// prior to performing operations
+	pvcNameKey      = "kubernetes.io/created-for/pvc/name"
```
davidz627: can we also add pv name as one of these?

pkg/controller/controller.go
```diff
@@ -567,6 +572,10 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 		return nil, controller.ProvisioningFinished, fmt.Errorf("failed to strip CSI Parameters of prefixed keys: %v", err)
 	}
 
+	// add pvc metadata to request for use by the plugin
+	req.Parameters[pvcNameKey] = options.PVC.GetName()
```
davidz627: we should gate this behind a feature flag or binary flag of some sort. This is a breaking change for drivers that do validation on parameters and don't expect extra stuff coming in. (this should default to off and you can turn it on if you want this information)

pkg/controller/controller.go
zetsub0u: We thought about this but, the PV name is generated in the provisioner and passed in the request as the Name field already and always, do you see any need for adding it into the parameters also?

pkg/controller/controller.go
zetsub0u: It seems like the featureFlag gate used here is being pulled from the api server, i'm not too sure if i should extend it with custom provisioner-only feature gates or just add a boolean flag say `--inject-pvc-metadata`, what do you think is the best approach?

pkg/controller/controller.go
davidz627: would prefer it is added to the parameters as well. Using it as the name is just an implementation choice in the provisioner not an explicit contract so it could technically change.

pkg/controller/controller.go
davidz627: I think a boolean flag is fine. I suck at naming so I can't come up with another solution but I don't think it should contain `pvc` in it since we will inject `pv/pvc` name. It should also have some sort of information of "where" it is injecting the metadata. `extra-create-params-metadata`?

Can you add a comment disclaimer as a flag comment along the lines of:
`This metadata should only be used for tagging or descriptive purposes and should in no circumstances be consumed and acted on programatically by a CSI Driver.`

yoesmat: Are we going to follow some sort of Alpha->Beta->GA plan for this flag?
If so can we include the stage in the name '--extra-create-params=alpha-pvc'?

davidz627: no need for a stage - just a single flag should be fine

yoesmat: What I am worried about in that case is that we will keep adding flags and the command to run the container will be really large by default.

It would be nice to have this be the default behavior at some point and the flag would be used to turn it off.

zetsub0u:
```      --feature-gates mapStringBool            A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
                                               AllAlpha=true|false (ALPHA - default=false)
                                               AllBeta=true|false (BETA - default=false)
                                               Topology=true|false (BETA - default=false)
                                               enable-extra-metadata-for-create-req=true|false (ALPHA - default=false)
```
I added this feature flag, does this sound reasonable? I accept name suggestions btw...

zetsub0u: Or do you prefer
```
 --enable-extra-metadata-for-create-req   If set, add pv/pvc metadata to plugin create requests as parameters.
```
I have both implement, just let me know which makes more sense.

yoesmat: I personally prefer the feature gate and have it go to default at some point.

davidz627: we can't set this as a default unless we bump the major version of the provisioner - it is a breaking change to drivers as this is adding "unexpected" parameters to their CreateVolume API call. I don't really see the value of feature gating this as the intention for that is to slowly "graduate" features to minimize risk. I wouldn't really consider this a risky feature in terms of potential impact or implementation

zetsub0u: Ok, i'll add a standard flag for this, any suggestions on naming? Anything else needing change? I couldn't find any place to document this feature for consumption...

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -64,10 +64,11 @@ var (
...
+	enableLeaderElection            = flag.Bool("enable-leader-election", false, "Enables leader election. If leader election is enabled, additional RBAC rules are required. Please refer to the Kubernetes CSI documentation for instructions on setting up these RBAC rules.")
+	leaderElectionType              = flag.String("leader-election-type", "endpoints", "the type of leader election, options are 'endpoints' (default) or 'leases' (strongly recommended). The 'endpoints' option is deprecated in favor of 'leases'.")
+	leaderElectionNamespace         = flag.String("leader-election-namespace", "", "Namespace where the leader election resource lives. Defaults to the pod namespace if not set.")
+	strictTopology                  = flag.Bool("strict-topology", false, "Passes only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.")
+	enableExtraMetadataForCreateReq = flag.Bool("enable-extra-metadata-for-create-req", false, "If set, add pv/pvc metadata to plugin create requests as parameters.")
```
davidz627: nit: `enable` not necessary (implied by boolean flag), `req` not required (implied by create)
`extra-create-metadata` is probably sufficient

pkg/controller/controller.go
```diff
@@ -101,6 +101,11 @@ const (
 	nodePublishSecretNameKey      = "csiNodePublishSecretName"
 	nodePublishSecretNamespaceKey = "csiNodePublishSecretNamespace"
 
+	// PV and PVC metadata, used for sending to drivers in the  create requests, added as parameters, optional.
+	pvcNameKey      = "kubernetes.io/created-for/pvc/name"
+	pvcNamespaceKey = "kubernetes.io/created-for/pvc/namespace"
+	pvNameKey       = "kubernetes.io/created-for/pv/name"
```
davidz627: cc @saad-ali is this an acceptable prefix? maybe it should be one of the volume specific prefixes?

pohly: I would prefer to use the same prefix as for [pod info on mount](https://kubernetes-csi.github.io/docs/pod-info.html): `csi.storage.k8s.io/...`

zetsub0u: Sounds good, updated.

cmd/csi-provisioner/csi-provisioner.go
zetsub0u: Done, i hate naming things :)

pkg/controller/controller.go
msau42: Also open an issue to track changing it to default, and we can add it to the 2.0 milestone

zetsub0u:
@davidz627 Let me know if you need any other changes. Thanks.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -68,6 +68,7 @@ var (
 	leaderElectionType      = flag.String("leader-election-type", "endpoints", "the type of leader election, options are 'endpoints' (default) or 'leases' (strongly recommended). The 'endpoints' option is deprecated in favor of 'leases'.")
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "Namespace where the leader election resource lives. Defaults to the pod namespace if not set.")
 	strictTopology          = flag.Bool("strict-topology", false, "Passes only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.")
+	extraCreateMetadata     = flag.Bool("extra-create-metadata", false, "If set, add pv/pvc metadata to plugin create requests as parameters.")
```
jsafrane: Please add the new option to README.md, together with list of parameters we pass.

pkg/controller/controller.go
```diff
@@ -104,6 +104,11 @@ const (
 	nodePublishSecretNameKey      = "csiNodePublishSecretName"
 	nodePublishSecretNamespaceKey = "csiNodePublishSecretNamespace"
 
+	// PV and PVC metadata, used for sending to drivers in the  create requests, added as parameters, optional.
+	pvcNameKey      = "csi.storage.k8s.io/created-for/pvc/name"
+	pvcNamespaceKey = "csi.storage.k8s.io/created-for/pvc/namespace"
+	pvNameKey       = "csi.storage.k8s.io/created-for/pv/name"
```
jsafrane: Would it make sense to use [the same schema we use in NodePublish](https://github.com/kubernetes/kubernetes/blob/7bbc06fcd436fda2bc82205598e1a91ce568b36f/pkg/volume/csi/csi_mounter.go#L317-L320)?

```
csi.storage.k8s.io/pvc/name
csi.storage.k8s.io/pvc/namespace
csi.storage.k8s.io/pv/name
```

I know CSI drivers that support migration need to translate it to `created-for/` format, still, it would make life of all the non-migrating drivers a bit easier.

jsafrane:
Oh, and please edit the first comment and add useful release note there (instead of `None`) - this is user visible change and actually it's probably going to be highlight of the new release ;-).

davidz627:
@zetsub0u Hi! Is this PR still active?

zetsub0u:
Yes, sorry. I was waiting to see if there was any discussion/change happening, i'll add the last requests soon.

andromedarabbit:
It would be nice if `KubernetesCluster` tag is supported also. Newer versions of Kubernetes might use the tag prefix `kubernetes.io/cluster/` instead:

https://github.com/kubernetes/kubernetes/blob/103e926604de6f79161b78af3e792d0ed282bc06/staging/src/k8s.io/legacy-cloud-providers/aws/tags.go

Thanks for the great PR. This should make my life very easier :)

msau42:
I don't think cluster tag is available on the PVC object. It's probably something that could be passed directly to the CSI driver since its generally deployed once per cluster

zetsub0u:
Updated the keys for the parameters and the Release notes.

davidz627:
@zetsub0u Thanks for the change! Please squash commits to 1 well described commit :)
/approve

jsafrane:
/lgtm
thanks a lot!

zetsub0u:
Thanks! We really needed this feature and didn't want to fork :)

Issue #328, opened 2019-08-13
Make fstype configurable while starting the external-provisioner


Madhu-1:
currently in external-provisioner fstype is set to `ext4` if nothing is passed in storage-class, we have a requirement to make it configurable so that the SP who is using the external-provisioner can choose the default fstype which they want to have it. This will avoid always mentioning of the fstype in the storage-class.

 in rbd csi plugin, we have seen better performance with `xfs` type, as external-provisioner is defaulting to `ext4`, in rbd csi we are asking users to mention the fstype as `xfs` in storage class. if fstype can be configured in external-provisioner user don't need to bother about the default fstype used.

CC @msau42

Madhu-1:
@humblec FYI

msau42:
I don't think fstype should be defaulted at all. There are filer types like nfs that expect a "" as the fstype

humblec:
@msau42 As @Madhu-1 mentioned, fstype is set to `ext4`  in sidecar container if not passed in SC param  https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L437 , so the only way to get around this is via `SC` parameter which is actually on overhead as all SC has to set this field.

msau42:
That doesn't seem right. Filer types like nfs/gluster/ceph should be failing when they validate fstype

msau42:
See https://github.com/kubernetes/kubernetes/issues/65122

humblec:
Have to go through in detail, but at glance , CSI pv source has been corrected. This has to be corrected here too. Isnt it ?

msau42:
Yes I don't think external provisioner should default. Now the question is will we break drivers that expect a non empty value? Need to double check CSI spec wording in this

humblec:
> Yes I don't think external provisioner should default. Now the question is will we break drivers that expect a non empty value? Need to double check CSI spec wording in this>

Yeah, we have to cross check on this, and also ( may be ) on backward compatibility of sidecar  ?

msau42:
I spoke with @jsafrane and he suggested we add a `--default-fs` flag, which defaults to `ext4` and then immediately deprecate it. When we have a 2.0, we can remove the defaulting behavior.

Does this sound reasonable to you? Would you have time to work on this?

jsafrane:
[...]

PR #400, opened 2020-02-04
Make fstype configurable in external provisioner

humblec:
At present the fstype is set to `ext4` if nothing is passed in storage-class.
However a SP could prefer to have different fstype for many reasons for their
driver/volumes. This patch enables SP who is using the external-provisioner
to choose the  fstype which they want to have it.

Fix# https://github.com/kubernetes-csi/external-provisioner/issues/328

> /kind bug

```release-note
 The `defaultFStype` arg is no longer default to "ext4" filesystem type. This arg is available for SP to use when they deploy provisioner sidecar. Admins can also specify this `fstype` via storage class parameter. If `fstype` is set in storage class parameter, it will be used. The sidecar arg is only checked if fstype is not set in the SC param.
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -71,7 +71,7 @@ var (
 	metricsAddress = flag.String("metrics-address", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
 	metricsPath    = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
 
-	defaultFSType = flag.String("default-fstype", "ext4", "Specify the filesystem type of the volume. If not specified,  external-provisioner will set default as `ext4`.")
+	defaultFSType = flag.String("default-fstype", "", "Specify the filesystem type of the volume. If not specified,  external-provisioner will not set any default filesystem type.")
 
 	featureGates        map[string]bool
 	provisionController *controller.ProvisionController
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -71,7 +71,7 @@ var (
 	metricsAddress = flag.String("metrics-address", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
 	metricsPath    = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
 
-	defaultFSType = flag.String("default-fstype", "", "Specify the filesystem type of the volume. If not specified,  external-provisioner will not set any default filesystem type.")
+	defaultFSType = flag.String("default-fstype", "", "The default filesystem type of the volume to provision when fstype is unspecified in the StorageClass. If the default is not set and fstype is unset in the StorageClass, then no fstype will be set")
 
 	featureGates        map[string]bool
 	provisionController *controller.ProvisionController
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -26,28 +26,25 @@ import (
 	"strings"
 	"time"
 
-	flag "github.com/spf13/pflag"
-
 	"github.com/kubernetes-csi/csi-lib-utils/deprecatedflags"
 	"github.com/kubernetes-csi/csi-lib-utils/leaderelection"
 	"github.com/kubernetes-csi/csi-lib-utils/metrics"
 	ctrl "github.com/kubernetes-csi/external-provisioner/pkg/controller"
 	snapclientset "github.com/kubernetes-csi/external-snapshotter/pkg/client/clientset/versioned"
-	"sigs.k8s.io/sig-storage-lib-external-provisioner/v5/controller"
-
+	flag "github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/util/wait"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	v1 "k8s.io/client-go/listers/core/v1"
+	storagelistersv1 "k8s.io/client-go/listers/storage/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/workqueue"
-	"k8s.io/klog"
-
-	utilfeature "k8s.io/apiserver/pkg/util/feature"
-	"k8s.io/client-go/informers"
-	v1 "k8s.io/client-go/listers/core/v1"
-	storagelistersv1 "k8s.io/client-go/listers/storage/v1"
 	utilflag "k8s.io/component-base/cli/flag"
 	csitrans "k8s.io/csi-translation-lib"
+	"k8s.io/klog"
+	"sigs.k8s.io/sig-storage-lib-external-provisioner/v5/controller"
 )
 
 var (
@@ -73,6 +70,8 @@ var (
 
 	metricsAddress = flag.String("metrics-address", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
 	metricsPath    = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
+
+	defaultFSType = flag.String("default-fstype", "ext4", "Specify the filesystem type of the volume. If not specified,  external-provisioner will set default as `ext4`.")
 
 	featureGates        map[string]bool
 	provisionController *controller.ProvisionController
@@ -240,6 +239,7 @@ func main() {
 		claimLister,
 		vaLister,
 		*extraCreateMetadata,
+		*defaultFSType,
 	)
 
 	provisionController = controller.NewProvisionController(
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -116,8 +116,6 @@ const (
 	backoffFactor   = 1.2
 	backoffSteps    = 10
 
-	defaultFSType = "ext4"
-
 	snapshotKind     = "VolumeSnapshot"
 	snapshotAPIGroup = snapapi.GroupName       // "snapshot.storage.k8s.io"
 	pvcKind          = "PersistentVolumeClaim" // Native types don't require an API group
@@ -211,6 +209,7 @@ type csiProvisioner struct {
 	timeout                               time.Duration
 	identity                              string
 	volumeNamePrefix                      string
+	defaultFSType                         string
 	volumeNameUUIDLength                  int
 	config                                *rest.Config
 	driverName                            string
@@ -292,6 +291,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 	claimLister corelisters.PersistentVolumeClaimLister,
 	vaLister storagelistersv1.VolumeAttachmentLister,
 	extraCreateMetadata bool,
+	defaultFSType string,
 ) controller.Provisioner {
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartLogging(klog.Infof)
@@ -307,6 +307,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 		timeout:                               connectionTimeout,
 		identity:                              identity,
 		volumeNamePrefix:                      volumeNamePrefix,
+		defaultFSType:                         defaultFSType,
 		volumeNameUUIDLength:                  volumeNameUUIDLength,
 		driverName:                            driverName,
 		pluginCapabilities:                    pluginCapabilities,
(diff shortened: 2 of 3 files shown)


humblec:
@jsafrane @msau42 ptal .

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -89,6 +90,8 @@ func main() {
 	flag.Var(utilflag.NewMapStringBool(&featureGates), "feature-gates", "A set of key=value pairs that describe feature gates for alpha/experimental features. "+
 		"Options are:\n"+strings.Join(utilfeature.DefaultFeatureGate.KnownFeatures(), "\n"))
 
+	flag.StringVar(&provisionercontroller.DefaultFSType, "default-fstype", "ext4", "Specify the filesystem type of the volume. If not specified,  external-provisioner will set default as `ext4`.")
```
davidz627: If we're going to do this I would prefer we follow the style of defining the other flags like `metricsPath`

humblec: @davidz627 the thing to note here is, `defaultFStype` is already defined here https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L119 unlike other variables in this section like `metricsPath`. So, isnt it good to be here ?

davidz627: that can be changed, it's just a variable

humblec: @davidz627 this comment is addressed. PTAL 
@jsafrane 
@msau42

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -73,6 +73,8 @@ var (
 	metricsAddress = flag.String("metrics-address", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
 	metricsPath    = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
 
+	defaultFSType = flag.String("default-fstype", "ext4", "Specify the filesystem type of the volume. If not specified,  external-provisioner will set default as `ext4`.")
```
xing-yang: This would still force the driver to specify a default fstype ahead of time.  Can we just let driver determine and set this value internally?

Madhu-1: https://github.com/kubernetes-csi/external-provisioner/issues/328#issuecomment-528442910

pohly: We could introduce `--default-fstype=none` as a backwards-compatible way of letting a CSI driver specify "don't set a default in external-provisioner, I'll handle that myself". "ext4" would still be the default, so nothing changes for drivers that don't use this new option.

humblec: Yeah,  defaulting to `ext4` would be the way for now. Isnt this PR good to go then? @jsafrane @msau42 @xing-yang ?

pohly: That's not quite what I meant. We should default to `ext4` to avoid the need for a major release, but *also* allow a CSI driver to use the modified external-provisioner without any defaulting. We can enable the latter with `--default-fstype=none`.

pkg/controller/controller.go
```diff
@@ -496,7 +497,7 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 		return nil, controller.ProvisioningFinished, fmt.Errorf("fstype specified in parameters with both \"fstype\" and \"%s\" keys", prefixedFsTypeKey)
 	}
 	if len(fsType) == 0 {
-		fsType = defaultFSType
+		fsType = p.defaultFSType
```
pohly: Here's the implementation of `--default-fstype=none`:

```
if fsType == "" && p.defaultFSType != "none" {
    fsType = p.defaultFSType
}
```

humblec: Done.

cmd/csi-provisioner/csi-provisioner.go
humblec: Done..

xing-yang: Can you add a brief explanation of what `none` means in the help message and also add a unit test for `--default-fstype=none`?

msau42: I am actually thinking of releasing a 2.0 this cycle.  There are a number of 2.0 changes that I want to target and I think we have gathered enough that it's time to do them all.

So I think we can actually make the change to remove the defaulting altogether and not have a new argument for it.

(I do want to release a 1.6.0 first, so let's not merge this until I cut a 1.6.0)

xing-yang: Removing the defaulting will be even better.

humblec: > I am actually thinking of releasing a 2.0 this cycle. There are a number of 2.0 changes that I want to target and I think we have gathered enough that it's time to do them all.
> 
> So I think we can actually make the change to remove the defaulting altogether and not have a new argument for it.
> 
> (I do want to release a 1.6.0 first, so let's not merge this until I cut a 1.6.0)

@msau42 iiuc, the suggestion here is to remove the `defaultFSType` arg completely from this PR and keep it ready for `2.0` , Isnt it ?

msau42: sorry to flip back and forth on this. We discussed this in the csi standup today and we may still want to provide an option to set a default so that we can do fstype source/target matching enforcement for cloning and potentially snapshots. cc @jsafrane @xing-yang @j-griffith

humblec: @msau42 done.. nw :).. ptal .

pkg/controller/controller_test.go
```diff
@@ -66,6 +66,7 @@ var (
 
 	driverNameAnnotation = map[string]string{annStorageProvisioner: driverName}
 	translatedKey        = "translated"
+	defaultfsType        = "ext4"
```
msau42: Add a test case with the "none" option?

cmd/csi-provisioner/csi-provisioner.go
msau42: Do we want to wait for a 2.0 so that we can add the option without a default?

Will it be confusing if we add an option now and immediately deprecate it and change it in the next release?

msau42: 1.6 is released. And I expect this next release to be a 2.0. So let's go ahead and add the new option without a default. I have mentioned this in the deprecation notes.

humblec: @msau42 sure . thanks!

msau42:
@humblec can you rebase? We're ready to merge this.

humblec:
> @humblec can you rebase? We're ready to merge this.

Sure @msau42 .. just went out of radar due to other clutter of PRs..

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -74,6 +71,8 @@ var (
...
```
msau42, line 74: Can we make this opt-in so that plugins don't need a "none" value?

msau42:
Can you also add a release note?

cmd/csi-provisioner/csi-provisioner.go
xing-yang: Can you add a comment here about "none"?  Otherwise driver developer won't know that they can set "none" here unless if they read the code.

xing-yang: @msau42 do you mean we should allow plugin to leave this as "" instead of setting to "none" to indicate there is no default fstype?

msau42: Correct

humblec:
@msau42 @xing-yang I have to admit that, due to various comments in different times, I am confused a bit here. however the lastpatchset is carved out in such a way, the `defaultFStype` arg is available for SP's to use when they deploy sidecar with a default value `"" `. If not set, the fsType value will be unfilled  Admins can also specify this fstype via SC param. If both SC param and provisioner sidecar arg has been filled , the preference goes to the sidecar arg. 

Is this the desired thing we want to achieve? Did I miss anything here?

xing-yang:
> If both SC param and provisioner sidecar arg has been filled , the preference goes to the sidecar arg.

I'm looking at the code in this PR.  It is not exactly what you described above.  If fstype is set in SC param, it will be used.  The sidecar arg is only checked if fstype is not set in the SC param.  This looks right to me.

humblec:
> > If both SC param and provisioner sidecar arg has been filled , the preference goes to the sidecar arg.
> 
> I'm looking at the code in this PR. It is not exactly what you described above. If fstype is set in SC param, it will be used. The sidecar arg is only checked if fstype is not set in the SC param. This looks right to me.

Ah.. exactly. Corrected my explanation too!

xing-yang:
Can you add this description to the release note?

humblec:
> Can you add this description to the release note?

I have added it. Please let me know if it need to be modified @xing-yang

xing-yang:
Can you move the release notes to the "Does this PR introduce a user-facing change?" section?

humblec:
> Can you move the release notes to the "Does this PR introduce a user-facing change?" section?

@xing-yang may be both can be set, I mean `release-note` and also `user-facing-*` section ?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -74,6 +71,8 @@ var (
 	metricsAddress = flag.String("metrics-address", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
 	metricsPath    = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
 
+	defaultFSType = flag.String("default-fstype", "", "Specify the filesystem type of the volume. If not specified,  external-provisioner will not set any default filesystem type.")
```
msau42, line 74: How about something like:
```
The default filesystem type of the volume to provision when fstype is unspecified in the StorageClass. If the default is not set and fstype is unset in the StorageClass, then no fstype will be set.
```

pkg/controller/controller.go
```diff
@@ -518,8 +519,8 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 	if fsTypesFound > 1 {
 		return nil, controller.ProvisioningFinished, fmt.Errorf("fstype specified in parameters with both \"fstype\" and \"%s\" keys", prefixedFsTypeKey)
 	}
-	if len(fsType) == 0 {
-		fsType = defaultFSType
+	if fsType == "" && p.defaultFSType != "" {
```
msau42, line 522: Can you add unit tests that test the various combinations?
- StorageClass set + default set
- StorageClass unset + default set
- StorageClass set + default unset
- StorageClass unset + default unset

humblec:
Running tool: /usr/local/go/bin/go test -timeout 30s github.com/kubernetes-csi/external-provisioner/pkg/controller -run ^(TestProvision)$

ok  	github.com/kubernetes-csi/external-provisioner/pkg/controller	0.040s

/usr/local/go/bin/go test -timeout 30s github.com/kubernetes-csi/external-provisioner/pkg/controller -run ^(TestFSTypeProvision)$

ok  	github.com/kubernetes-csi/external-provisioner/pkg/controller	(cached)

cmd/csi-provisioner/csi-provisioner.go
humblec: Sure. Done.

humblec:
@msau42 @jsafrane @xing-yang any thoughts on why few of the existing tests ( For ex: TestProvision) are failing randomly when package-level test is triggered ?

pkg/controller/controller_test.go
```diff
@@ -1709,9 +1838,12 @@ func runProvisionTest(t *testing.T, k string, tc provisioningTestcase, requested
 	clientSet := fakeclientset.NewSimpleClientset(tc.clientSetObjects...)
 
 	pluginCaps, controllerCaps := provisionCapabilities()
-	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5, csiConn.conn,
-		nil, provisionDriverName, pluginCaps, controllerCaps, supportsMigrationFromInTreePluginName, false, csitrans.New(), nil, nil, nil, nil, nil, tc.withExtraMetadata)
 
+	if tc.skipDefaultFSType {
+		defaultfsType = ""
```
msau42, line 1843: I think the unit test is failing because this is modifying a global variable.

humblec: @msau42 Thanks! Looks like that.. after untangling the test case , its working as expected!

pkg/controller/controller.go
humblec: Added the test cases ! ..

humblec:
@msau42 I believe this is good to go in!

pkg/controller/controller_test.go
```diff
@@ -1709,9 +1851,88 @@ func runProvisionTest(t *testing.T, k string, tc provisioningTestcase, requested
...
+
+	if tc.expectState == "" {
+		tc.expectState = controller.ProvisioningFinished
+	}
+	if tc.expectState != state {
+		t.Errorf("test %q: expected ProvisioningState %s, got %s", k, tc.expectState, state)
+	}
+
+	if tc.expectedPVSpec != nil {
+		if pv.Name != tc.expectedPVSpec.Name {
```
msau42, line 1888: I think the test logic would be cleaner if we did a deepequal of the entire pv spec, instead of matching individual fields one by one. But we can consider cleaning that up separately.

humblec: Sure, I will do that in a follow-up PR. There are some more places it need to be done in this file anyway.

PR #404, opened 2020-02-12
master: update release-tools

pohly:
Commit summary:
5f74333a4 prow.sh: also configure feature gates for kubelet
84f78b120 prow.sh: generic driver installation
fa90abd07 fix incorrect link
ac8a0212b Document the process for releasing a new sidecar
6582f2ff3 Update hostpath driver version to get fix for connection-timeout
8191eab6f Update snapshotter to version v2.0.0
8b0316c7e Fix overriding of junit results by using unique names for each e2e run
af9549b5a Update prow hostpath driver version to 1.3.0-rc2
fc8097595 Fix version_gt to work with kubernetes prefix
b98b2aed0 Enable snapshot tests in 1.17 to be run in non-alpha jobs.
a4e629966 fix syntax for ppc64le build
4ad69492c Improve snapshot pod running checks and improve version_gt
53888ae7d Improve README by adding an explicit Kubernetes dependency section
9a7a685ee Create a kind cluster with two worker nodes so that the topology feature can be tested. Test cases that test accessing volumes from multiple nodes need to be skipped
80bba1fe2 Use kind v0.6.0
003c14b2d Add snapshotter CRDs after cluster setup
83a4ef15d Adding build for ppc64le

```release-note
NONE
```

### DIFF
--- a/release-tools/README.md
+++ b/release-tools/README.md
@@ -81,7 +81,7 @@ on what is enabled in Prow, see
 https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-csi
 
 Test results for periodic jobs are visible in
-https://testgrid.k8s.io/sig-storage-csi
+https://testgrid.k8s.io/sig-storage-csi-ci
 
 It is possible to reproduce the Prow testing locally on a suitable machine:
 - Linux host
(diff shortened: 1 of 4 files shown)


pohly:
This seems to work. But let's do a 1.3.0 release of the CSI hostpath driver first and then refresh all repos.

PR #405, opened 2020-02-13
unset selected node when storage is exhausted for topology segment

pohly:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This makes sense under some specific circumstances:
- volume is supposed to be created only in a certain topology segment
- that segment is chosen via the pod scheduler via late binding
- the CSI driver supports topology
- the CSI driver reports that it ran out of storage

Previously, external-provisioner would keep retrying to create the
volume instead of notifying the scheduler to pick a node anew.

Related-to: https://github.com/kubernetes/kubernetes/issues/72031

One downside of this change is that if the conditions above are met
and some other final error occurs, rescheduling also is triggered. A
richer API for the sig-storage-lib-external-provisioner library would
be needed to avoid this.

It's okay to treat ResourceExhausted as final error, the CSI spec
explicitly describes this case. Previously it was treated as non-final
error merely because retrying made more sense (resources might become
available).

The check for "strict topology" was added because without it, a CSI driver should already go ahead and pick some other topology segment than the one preferred for the selected node. It seems to make little sense to do another scheduling round if no topology segment has enough space.

```release-note
If a CSI driver returns ResourceExhausted for CreateVolume, supports topology, the volume uses late binding, and the driver has requested strict topology, then external-scheduler now asks for a rescheduling of the pod that triggered the volume creation. This may then result in a retry with a different selected node.
```


pkg/controller/controller.go
```diff
@@ -1165,11 +1189,17 @@ func isFinalError(err error) bool {
 		return false
 	}
 	switch st.Code() {
+	case codes.ResourceExhausted: // CSI: operation not pending, "Unable to provision in `accessible_topology`"
+		if mayReschedule {
+			// may succeed elsewhere -> give up for now
+			return true
+		}
+		// may still succeed at a later time -> continue
+		return false
```
jsafrane: This is unfortunate case of mixing gRPC transport error codes with application / CSI errors.

*Go gRPC implementation* returns `ResourceExhausted` when it processes too big messages (both when sending and receiving). This is the logic behind the current code - the operation itself may be in progress or even successfully finished (too big success response may have been just thrown away) and the provisioner should retry.

Looking at the code, gRPC default max message size is [sometimes defaulted to 4KiB](https://github.com/kubernetes-csi/external-provisioner/blob/e73400e556e73c0c43448fc497aefe26b6f2dc4f/vendor/google.golang.org/grpc/server.go#L56)

When a *CSI driver* intentionally returns `ResourceExhausted` when it gets out of space on the storage backend, the error is final and we can be 100% sure that the volume is being provisioned. Rescheduling the pod / PVC to a different node is correct.

Questions are:
* How to recognize *who* sent the response. IMO it's impossible to distinguish CSI driver from gRPC implementation.
* Whether we can assume that `ResourceExhausted` returned by gRPC implementation is safe **not** to retry. 
  * If CreateVolumeResponse / CreateVolumeRequest was too big, retrying the request should yield the same response and one of them is going to be too big again. Retrying is useless and there is a leaked volume. Trying on a different node won't harm.

pohly: This is a problem in the CSI spec, right? Instead of re-using status codes, it would have been better to pick new ones to avoid such ambiguity.

I agree that it is impossible to determine the sender. If sending the request already fails, then trying elsewhere is safe and in fact the only way how (theoretically) the volume might still succeed, because retrying locally will just send the same message over and over again.

When receiving the reply fails with `ResourceExhausted`, then the situation is worse in the sense that the volume probably has been provisioned, external-provisioner just never gets to know that. Retrying won't help here much either unless an admin steps in an reconfigures the external-provisioner with larger max message sizes.

I think we should bring this up as a problem with spec. But given what it is right now, avoiding the gRPC `ResourceExhausted` by using large enough message size limits and then assuming that the error really comes from the CSI driver seems viable to me.

pohly:
@jsafrane I've updated this to use sig-storage-lib-external-provisioner v5.0.0. I also filed an issue against the CSI spec about the `RESOURCE_EXHAUSTED` ambiguity and added that as comment to the code.

What's missing is a local test in `pkg/controller` for the new code path. Do we need one and if so, any suggestions what a good place for that might be?

E2E testing will cover this (https://github.com/kubernetes/kubernetes/pull/88114), too.

pkg/controller/controller.go
```diff
@@ -614,10 +614,31 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
...
 	if err != nil {
-		if isFinalError(err) {
-			return nil, controller.ProvisioningFinished, err
-		}
-		return nil, controller.ProvisioningInBackground, err
+		// Giving up after an error and telling the pod scheduler to retry with a different node
+		// only makes sense if:
+		// - The CSI driver supports topology: without that, the next CreateVolume call after
+		//   rescheduling will be exactly the same.
+		// - We are using strict topology: otherwise the CSI driver is already allowed
```
msau42: I think we can remove the check for strict topology. I'm not aware of any drivers that retry provisioning on subsequent topologies in the list if the first fails.

pkg/controller/controller.go
```diff
@@ -614,10 +614,31 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
...
+		//   provisioning be retried if we give up for now.
+		// - The error is one where rescheduling is
+		//   a) allowed (i.e. we don't have to keep calling CreateVolume because the operation might be running) and
+		//   b) it makes sense (typically local resource exhausted).
+		//   isFinalError is going to check this.
+		mayReschedule := p.supportsTopology() &&
+			p.strictTopology &&
+			options.SelectedNode != nil
+		state := checkError(err, mayReschedule)
+		klog.V(5).Infof("CreateVolume failed, supports topology = %v, strict topology %v, node selected %v => may reschedule = %v => state = %v: %v",
```
msau42: should this be a warning?

go.mod
```diff
@@ -8,18 +8,18 @@ require (
...
+	k8s.io/api v0.17.3
+	k8s.io/apimachinery v0.17.3
 	k8s.io/apiserver v0.17.0
 	k8s.io/client-go v0.17.0
 	k8s.io/component-base v0.17.0
 	k8s.io/csi-translation-lib v0.17.0
 	k8s.io/klog v1.0.0
 	k8s.io/kubernetes v1.14.0
-	sigs.k8s.io/sig-storage-lib-external-provisioner v4.1.0+incompatible
+	sigs.k8s.io/sig-storage-lib-external-provisioner v4.1.0+incompatible // indirect
```
msau42, line 21: This is kind of strange, we have dependencies on both versions?

pohly: I had noticed that, too, but because we never actually pull in any code from it (= it's not in `vendor`) ignored it.

This might be a bug in `sig-storage-lib-external-provisioner`:
```
$ go mod why -m sigs.k8s.io/sig-storage-lib-external-provisioner
# sigs.k8s.io/sig-storage-lib-external-provisioner
github.com/kubernetes-csi/external-provisioner/cmd/csi-provisioner
sigs.k8s.io/sig-storage-lib-external-provisioner/v5/controller
sigs.k8s.io/sig-storage-lib-external-provisioner/v5/controller.test
sigs.k8s.io/sig-storage-lib-external-provisioner/controller/metrics
```

=> controller.test (which we don't use) pulls in the unversioned `sigs.k8s.io/sig-storage-lib-external-provisioner` here: https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/7d9e8b4c678a803070309639110a4494a4010efe/controller/controller_test.go#L26

Not sure how that works; perhaps `go test` simply maps that to the checked out module source code. Will fix it.

pohly: Fix is in https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/71

msau42: Thanks! Since we're not actually using it here, we don't need to wait for.

PR #409, opened 2020-02-21
Add more tests for provsionsing from PVC as DataSource

ShyamsundarR:
**What type of PR is this?**
/kind bug

This commit adds 6 more tests to the existing unit tests
for creating PVC given a DataSource as another PVC, IOW
cloning feature.

The tests are added in 2 categories,
- When source PVCs Status.Phase is different than,
  ClaimBound
- When clone PV source's Status.Phase is other than,
  VolumeBound

The tests did open up a code issue, where PV Status was
not checked for VolumeBound status. This is also fixed
in this commit.

Tested using `make test-go` against go compiler
version go1.13.6 linux/amd64

NA, Added to help move clone feature to GA.

- If PV `Spec.ClaimRef` is non-nil, can `Status.Phase` be anything other than bound? IOW, is this test valid and needed?

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -822,6 +822,11 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
 		return nil, fmt.Errorf("claim in dataSource not bound or invalid")
 	}
 
+	if sourcePV.Status.Phase != v1.VolumeBound {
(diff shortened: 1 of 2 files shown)


jsafrane:
lgtm-ish, but refactoring to #411 - style would be helpful.

ShyamsundarR:
> lgtm-ish, but refactoring to #411 - style would be helpful.

Considering the order in which PRs were submitted, I was thinking if #409 and #410 were merged first, I would rebase #411 PR against the merged master and refactor the newly added tests as well.

We could also get #411 in first and rework the other two based on that. Let me know what is preferred and I can sequence it accordingly. Thanks.

ShyamsundarR:
> > lgtm-ish, but refactoring to #411 - style would be helpful.

As #411 is merged, rebased this to include the same and reduce TC clutter.

PR #410, opened 2020-02-21
Add volumeMode checks to clone operations

j-griffith:
/kind cleanup

This PR add checks against the volume mode between source and newly created
clone of a pvc.  If the source PVC does not match the requested volume mode of
the newly created clone the create operation will fail.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -832,6 +832,18 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
 		return nil, fmt.Errorf("claim in dataSource not bound or invalid")
 	}
 
+	if options.PVC.Spec.VolumeMode == nil || *options.PVC.Spec.VolumeMode == v1.PersistentVolumeFilesystem {
+		if sourcePV.Spec.VolumeMode != nil && *sourcePV.Spec.VolumeMode != v1.PersistentVolumeFilesystem {
+			return nil, fmt.Errorf("the source PVC and destination PVCs must have the same volume mode for cloning.  Source is Block, but new PVC requested Filesystem")
+		}
+	}
+
+	if options.PVC.Spec.VolumeMode != nil && *options.PVC.Spec.VolumeMode == v1.PersistentVolumeBlock {
+		if sourcePV.Spec.VolumeMode == nil || *sourcePV.Spec.VolumeMode != v1.PersistentVolumeBlock {
+			return nil, fmt.Errorf("the source PVC and destination PVCs must have the same volume mode for cloning.  Source is Filesystem, but new PVC requested Block")
+		}
+	}
+
 	volumeSource := csi.VolumeContentSource_Volume{
 		Volume: &csi.VolumeContentSource_VolumeSource{
 			VolumeId: sourcePV.Spec.CSI.VolumeHandle,
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -513,7 +513,7 @@ func createFakePVCWithVolumeMode(requestBytes int64, volumeMode v1.PersistentVol
 }
 
 // fakeClaim returns a valid PVC with the requested settings
-func fakeClaim(name, namespace, claimUID string, capacity int64, boundToVolume string, phase v1.PersistentVolumeClaimPhase, class *string) *v1.PersistentVolumeClaim {
+func fakeClaim(name, namespace, claimUID string, capacity int64, boundToVolume string, phase v1.PersistentVolumeClaimPhase, class *string, mode string) *v1.PersistentVolumeClaim {
 	claim := v1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            name,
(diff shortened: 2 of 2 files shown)


pkg/controller/controller_test.go
```diff
@@ -3444,6 +3597,55 @@ func TestProvisionFromPVC(t *testing.T) {
 			},
 		}
 
+		nilModePV := &v1.PersistentVolume{
```
ShyamsundarR: `nilModePV` is the same as `pv` above, and `pv` can be reused instead.

pkg/controller/controller_test.go
```diff
@@ -3378,6 +3380,157 @@ func TestProvisionFromPVC(t *testing.T) {
 			cloneSupport:   true,
 			expectErr:      true,
 		},
+		"provision new block volume with pvc data source mode as filesystem": {
```
ShyamsundarR: The intention of this test case seems to be to point to a source PV whose VolumeMode is explicitly set to "Filesystem", however PVC data source is `srcName` which gets 'pvName' claim bound to it. This source PVs VolumeMode is hence `nil` by default. As a result, this and the next test case at line 3421 are the same.

pkg/controller/controller_test.go
```diff
@@ -3444,6 +3597,55 @@ func TestProvisionFromPVC(t *testing.T) {
...
+					Namespace: srcNamespace,
+					Name:      "nil-mode-claim",
+					UID:       types.UID("fake-nil-claim-uid"),
+				},
+				StorageClassName: fakeSc1,
+			},
+		}
+
+		//clonePVName: "pvSrcBlockMode",
+		blkModePV := &v1.PersistentVolume{
```
ShyamsundarR: (nit) `blkModePV` can be instead defined as,
```
                blkModePV := pv.DeepCopy()
		blkModePV.Name = pvSrcBlockMode
		blkModePV.Spec.VolumeMode = &volumeModeBlock
```

And corresponding test cases can be updated to reflect `tc.clonePVName` as `pvSrcBlockMode`

This also removes the need to add `blockModeClaim` below.

j-griffith: Actually I noticed some bad behavior in the tests (that I didn't track down).  It seems that we're missing cleanup steps in the tests and the result of trying to do something like you suggest above is things like invalid claims, and unexpected status on PVs.  I'll take a look again, maybe it was something else I had incorrect, or maybe I can more easily identify the problem.  Using debug statements in the catches though, I did notice that our tests don't always fail for the reasons we're expecting them to (sort of a problem), we set `expectErr: true` and let it rip, but it's not always failing for what we expect.

pkg/controller/controller_test.go
j-griffith: Yes, you're correct good catch, I'll get it fixed up

pkg/controller/controller_test.go
j-griffith: Yeah, I'll go through these again and verify the modes again; thanks.

pkg/controller/controller_test.go
j-griffith: I think we should open an issue against the tests, and not necessarily failing for what we expect.  I verified this when trying to copy the pv object into a block object, I suspect we might be doing this in other places.

ShyamsundarR: I agree, there were a few flakes and incorrect errors, I have put up this [PR](https://github.com/kubernetes-csi/external-provisioner/pull/411) as an initial cleanup for the tests. It does not ensure the errors are what is expected, when returned are what is expected though.

pkg/controller/controller.go
```diff
@@ -819,6 +819,23 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
 		return nil, fmt.Errorf("claim in dataSource not bound or invalid")
 	}
 
+	if options.PVC.Spec.VolumeName == "my-block-pvc" {
```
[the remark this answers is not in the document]
ShyamsundarR: Resolved by the latest update

pkg/controller/controller_test.go
ShyamsundarR: Resolved by the latest update

ShyamsundarR:
Barring the issue discussed in [this](https://github.com/kubernetes-csi/external-provisioner/pull/410#discussion_r382999744) comment, which needs a separate resolution anyway, this PR is good to go, IMO.

pkg/controller/controller_test.go
```diff
@@ -3378,6 +3379,119 @@ func TestProvisionFromPVC(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			pvcStatusReady: true,
+			expectedPVSpec: nil,
+			cloneSupport:   true,
+			expectErr:      true,
+		},
+		"provision with pvc source is block request nil mode": {
```
msau42: Can you also add more positive and negative test cases with nil mode?

jsafrane:
@j-griffith, can you please rebase and rework the unit tests to the new style? It should be easier now.

pkg/controller/controller.go
```diff
@@ -819,6 +819,19 @@ func (p *csiProvisioner) getPVCSource(options controller.ProvisionOptions) (*csi
...
+		}
+	}
+
+	if options.PVC.Spec.VolumeMode != nil && *options.PVC.Spec.VolumeMode == v1.PersistentVolumeBlock {
+		if sourcePV.Spec.VolumeMode == nil || *sourcePV.Spec.VolumeMode != v1.PersistentVolumeBlock {
+			return nil, fmt.Errorf("the source PVC and destination PVCs must have the same volume mode for cloning.  Source is Filesystem, but new PVC requested Block")
+
+		}
+	}
+
```
xing-yang: These checks look good to me.

PR #411, opened 2020-02-22
Cleanup and fix, clone unit tests

ShyamsundarR:
**What type of PR is this?**
/kind cleanup

There was quite a bit of repetitive code in each test case definition,
this has been cleaned up. Further following changes were done as some failures
during the cleanup were noted,

- Corrected size checks in controller.go, to use sourcePVC size instead
of gRPC payload size for comparison
- Updated fakeClaim in controller_test.go, to reflect correct sourcePVC
size (instead of defaulting to 1Gi for a 1000 byte request)
- Defined and reused a defaultPVCRequest across tests
- Removed unused test case struct elements
- Removed restating default test case parameters per test
- Refactored core test case loop into phases, and removed any duplicate
calls/variables
- Corrected expected RPC calls when sourcePVC size is greater than
requestedPVC

Tested for flakes by running the following bash loop,
for i in {1..50}; do go clean -testcache; make test-go; done

Tested using, go version go1.13.6 linux/amd64

None

- Should we add checks in the code to prevent a CreateVolume->DeleteVolume set of calls when the requested PVC size is greater than source PVC?
    - IOW, avoid hitting [this](https://github.com/kubernetes-csi/external-provisioner/blob/40297715f5409a617c35b2aa29f1c812df63dc29/pkg/controller/controller.go#L619-L630) condition when DataSource is a PVC and we know the sizes mismatch, by checking ahead [here](https://github.com/kubernetes-csi/external-provisioner/blob/40297715f5409a617c35b2aa29f1c812df63dc29/pkg/controller/controller.go#L786-L790).

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -526,7 +526,7 @@ func fakeClaim(name, namespace, claimUID string, capacity int64, boundToVolume s
 			AccessModes: []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce, v1.ReadOnlyMany},
 			Resources: v1.ResourceRequirements{
 				Requests: v1.ResourceList{
-					v1.ResourceName(v1.ResourceStorage): resource.MustParse(strconv.FormatInt(capacity, 10)),
+					v1.ResourceName(v1.ResourceStorage): *resource.NewQuantity(capacity, resource.BinarySI),
 				},
 			},
 			VolumeName:       boundToVolume,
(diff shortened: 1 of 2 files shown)


ShyamsundarR:
/assign @msau42 
/assign @j-griffith 
/assign @jsafrane 

I do not expect this patch to be merged before, either of the following 2 PRs (unless of course maintainers think otherwise), but setting this up for a cleanup of the tests and reviews on approach/changes, (hopefully) before the clone GA release,
- https://github.com/kubernetes-csi/external-provisioner/pull/410
- https://github.com/kubernetes-csi/external-provisioner/pull/409

pkg/controller/controller_test.go
```diff
@@ -3508,6 +3152,16 @@ func TestProvisionFromPVC(t *testing.T) {
 			t.Errorf("test %q: Expected error, got none", k)
 		}
 
+		// Phase: process test responses
+		// reset tc.VolOpts back to defaults
+		if resetSize != nil {
```
msau42: Instead of having tc carry over across test cases, we should wrap everything around t.Run()

ShyamsundarR: Updated the PR adding another commit, that prevents per-tc modification of common content. Also, added the required t.Run() call, and added a t.Parallel() within it (more as proof of each tc not stomping on each others state, rather than for performance).

Assuming this addresses this comment and helps with the cleanup.

pkg/controller/controller_test.go
```diff
@@ -527,7 +526,7 @@ func fakeClaim(name, namespace, claimUID, capacity, boundToVolume string, phase
 			AccessModes: []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce, v1.ReadOnlyMany},
 			Resources: v1.ResourceRequirements{
 				Requests: v1.ResourceList{
-					v1.ResourceName(v1.ResourceStorage): resource.MustParse(capacity),
+					v1.ResourceName(v1.ResourceStorage): resource.MustParse(strconv.FormatInt(capacity, 10)),
```
jsafrane: Use `NewQuantity()` instead of converting int to string and parsing it.

ShyamsundarR: Addressed here and also in the `generatePVCForProvisionFromPVC` function.

When addressing this I did find that the addition of t.Parallel required a local assignment of tc, as otherwise every parallel test was executing the same tc (the last one as per the random order of tests). Would like a review of the change specific to this a bit more carefully, https://github.com/kubernetes-csi/external-provisioner/pull/411/files#diff-a846267007f026d95b52f6baaf249948R2974-R2978

jsafrane:
/lgtm
/approve
Thanks for the cleanup!

PR #412, opened 2020-02-24
Accept provision/delete calls from PVCs that have the 'migrated-to' annotation of the provisioner name

davidz627:
/kind feature

Final piece to fix: https://github.com/kubernetes/kubernetes/issues/79043

```release-note
CSI Migration: Fixes CSI migration issue where PVCs provisioned before migration was turned on could not be deleted
```


davidz627:
/cc @msau42 @jsafrane 
this PR is ready for review

pkg/controller/controller.go
```diff
@@ -403,12 +404,16 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 		return nil, controller.ProvisioningFinished, errors.New("storage class was nil")
 	}
 
-	if options.PVC.Annotations[annStorageProvisioner] != p.driverName {
+	if options.PVC.Annotations[annStorageProvisioner] != p.driverName && options.PVC.Annotations[annMigratedTo] != p.driverName {
```
msau42: do we have a unit test for this?

davidz627: now we do :)

msau42:
/lgtm
/approve

Can you make the release note more easily understandable by someone that doesn't have background on the issue? Something like "Fixes CSI migration issue where PVCs provisioned before migration was turned on could not be deleted"
/hold

PR #419, opened 2020-03-04
WIP: build for multiple platforms only in CI, add s390x

pohly:
Developers should not be forced to build for all platforms by
default. We also don't want to copy-and-paste the go invocation for
each new platform.

To address both, the target platform(s) are now configurable via
BUILD_PLATFORMS and additional platforms are only enabled in the Prow
CI.

### DIFF
--- a/release-tools/build.make
+++ b/release-tools/build.make
@@ -60,18 +60,25 @@ else
 TESTARGS =
 endif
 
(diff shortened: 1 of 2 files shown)


pohly:
Was merged into csi-release-tools.

Issue #420, opened 2020-03-10
single topology key is not working for waitforfirstconsumer, if driver supports multiple topology key


pawanpraka1:
If waitforfirstconsumer is used and if the driver supports multiple keys, mentioning single key in the storageclass is not working.

The csi-provisioner will get list of all the topology keys from the csinode object and gets the values for those keys from the node object. Now this key-value *term* is matched against the topology mentioned in the storageclass and it should match entirely which is not correct as if topology mentioned in the storageclass is subset of the key-value *term* supported by the driver, then the node is a valid condidate.

We should probably change this equal check with subset check.

https://github.com/kubernetes-csi/external-provisioner/blob/efcaee79e47446e38910a3c1a024824387fcf235/pkg/controller/topology.go#L483

pawanpraka1:
fixed via https://github.com/kubernetes-csi/external-provisioner/pull/421.

PR #421, opened 2020-03-11
subset of supported topology keys are not working for waitforfirstconsumer

pawanpraka1:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**

fixes:- https://github.com/kubernetes-csi/external-provisioner/issues/420

If waitforfirstconsumer is used and if the driver supports multiple topology  keys,
mentioning single key(or less than supported keys) in the storageclass is not working.
The csi-provisioner gets list of all the topology keys from the csinode object
and gets the values for those keys from the node object. Now this key-value term
is matched against the topology mentioned in the storageclass and it should match
entirely which is not correct as even if topology mentioned in the storageclass
is subset of the key-value term supported by the driver, then the node is a valid condidate.

Instead of strictly checking for the equality of the selected node's topology with
the allowedTopologies mentioned in the storageclass, we should check if allowedTopologies
is subset of selected node's topology.

**Does this PR introduce a user-facing change?:**
```release-note
We can now provide few keys in the allowedtopologies of all the keys supported by the driver in late binding case.
```

### DIFF
--- a/pkg/controller/topology.go
+++ b/pkg/controller/topology.go
@@ -200,7 +200,7 @@ func GenerateAccessibilityRequirements(
 				allowedTopologiesFlatten := flatten(allowedTopologies)
 				found := false
 				for _, t := range allowedTopologiesFlatten {
-					if t.equal(selectedTopology) {
+					if t.subset(selectedTopology) {
 						found = true
 						break
 					}
@@ -480,7 +480,7 @@ func sortAndShift(terms []topologyTerm, primary topologyTerm, shiftIndex uint32)
 		preferredTerms = append(terms[shiftIndex:], terms[:shiftIndex]...)
 	} else {
 		for i, t := range terms {
-			if t.equal(primary) {
+			if t.subset(primary) {
 				preferredTerms = append(terms[i:], terms[:i]...)
 				break
 			}
@@ -560,6 +560,17 @@ func (t topologyTerm) hash() string {
 
 func (t topologyTerm) less(other topologyTerm) bool {
 	return t.hash() < other.hash()
+}
+
+func (t topologyTerm) subset(other topologyTerm) bool {
+	for key, tv := range t {
+		v, ok := other[key]
+		if !ok || v != tv {
+			return false
+		}
+	}
+
+	return true
 }
 
 func (t topologyTerm) equal(other topologyTerm) bool {
--- a/pkg/controller/topology_test.go
+++ b/pkg/controller/topology_test.go
@@ -1323,6 +1323,73 @@ func TestPreferredTopologies(t *testing.T) {
 			},
 			expectError: true,
 		},
+		"allowedTopologies is subset of selected node's topology": {
+			allowedTopologies: []v1.TopologySelectorTerm{
+				{
+					MatchLabelExpressions: []v1.TopologySelectorLabelRequirement{
+						{
+							Key:    "com.example.csi/disk",
+							Values: []string{"ssd"},
+						},
+						{
+							Key:    "com.example.csi/zone",
+							Values: []string{"zone1"},
+						},
+					},
+				},
+			},
+			nodeLabels: []map[string]string{
+				{"com.example.csi/zone": "zone1", "com.example.csi/rack": "rack1", "com.example.csi/disk": "ssd"},
+				{"com.example.csi/zone": "zone2", "com.example.csi/rack": "rack2", "com.example.csi/disk": "ssd"},
+				{"com.example.csi/zone": "zone2", "com.example.csi/rack": "rack1", "com.example.csi/disk": "nvme"},
+			},
+			topologyKeys: []map[string][]string{
+				{testDriverName: []string{"com.example.csi/zone", "com.example.csi/rack", "com.example.csi/disk"}},
+				{testDriverName: []string{"com.example.csi/zone", "com.example.csi/rack", "com.example.csi/disk"}},
+				{testDriverName: []string{"com.example.csi/zone", "com.example.csi/rack", "com.example.csi/disk"}},
+			},
+			expectError: false,
+			expectedPreferred: []*csi.Topology{
+				{
+					Segments: map[string]string{
+						"com.example.csi/disk": "ssd",
+						"com.example.csi/zone": "zone1",
+					},
+				},
+			},
+			expectedStrictPreferred: []*csi.Topology{
+				{
+					Segments: map[string]string{
+						"com.example.csi/disk": "ssd",
+						"com.example.csi/rack": "rack1",
+						"com.example.csi/zone": "zone1",
+					},
+				},
+			},
+		},
+		"allowedTopologies is not the subset of selected node's topology": {
+			allowedTopologies: []v1.TopologySelectorTerm{
+				{
+					MatchLabelExpressions: []v1.TopologySelectorLabelRequirement{
+						{
+							Key:    "com.example.csi/zone",
+							Values: []string{"zone1"},
+						},
+					},
+				},
+			},
+			nodeLabels: []map[string]string{
+				{"com.example.csi/zone": "zone10", "com.example.csi/rack": "rack1"},
+				{"com.example.csi/zone": "zone11", "com.example.csi/rack": "rack2"},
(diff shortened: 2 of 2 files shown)


pawanpraka1:
I signed it

pawanpraka1:
I signed it

pawanpraka1:
I signed it

pawanpraka1:
I signed it

pkg/controller/topology_test.go
```diff
@@ -1323,6 +1323,44 @@ func TestPreferredTopologies(t *testing.T) {
...
+				},
+			},
+			nodeLabels: []map[string]string{
+				{"com.example.csi/zone": "zone1", "com.example.csi/rack": "rack1"},
+				{"com.example.csi/zone": "zone2", "com.example.csi/rack": "rack2"},
+				{"com.example.csi/zone": "zone2", "com.example.csi/rack": "rack1"},
+			},
+			topologyKeys: []map[string][]string{
+				{testDriverName: []string{"com.example.csi/zone", "com.example.csi/rack"}},
+				{testDriverName: []string{"com.example.csi/zone", "com.example.csi/rack"}},
```
msau42: these two entries can be removed, they are duplicates of the first

pkg/controller/topology_test.go
```diff
@@ -1323,6 +1323,44 @@ func TestPreferredTopologies(t *testing.T) {
...
+				{testDriverName: []string{"com.example.csi/zone", "com.example.csi/rack"}},
+			},
+			expectError: false,
+			expectedPreferred: []*csi.Topology{
+				{
+					Segments: map[string]string{
+						"com.example.csi/zone": "zone1",
```
msau42: For the case where selectedNode is set, I would expect the fully qualified topology to be generated (with zone + rack) in strict and non-strict cases.

pkg/controller/topology_test.go
```diff
@@ -1323,6 +1323,44 @@ func TestPreferredTopologies(t *testing.T) {
 			},
 			expectError: true,
 		},
+		"allowedTopologies is subset of selected node's topology": {
```
msau42: can you add test cases for when selectedNode is not set?

pkg/controller/topology.go
```diff
@@ -562,6 +562,10 @@ func (t topologyTerm) less(other topologyTerm) bool {
 	return t.hash() < other.hash()
 }
 
+func (t topologyTerm) subset(other topologyTerm) bool {
+	return strings.Contains(other.hash(), t.hash())
```
verult: What if `rackA` and `rackAB` both exist?

What if the two hashes are `...#rackA,...#rackC` and `...#rackA,...#rackB,...#rackC`?

pkg/controller/topology_test.go
pawanpraka1: right, we need one entry for each node.

pkg/controller/topology_test.go
pawanpraka1: there are already test cases for that where we will pass everything aftter flattening the list, we don;t bother about supported keys by the driver.  (for e.g. https://github.com/kubernetes-csi/external-provisioner/blob/39dcde6e61ead3a5b796cf6be6d4adb2e70c1088/pkg/controller/topology_test.go#L741).

As this PR touches the code path for waitforfirstconsumer only, so added test cases for selected node only.

pkg/controller/topology.go
pawanpraka1: good point. I should probably rely on checking individual key value pair. Thanks. Will fix this. I will also update the test case for this.

pkg/controller/topology_test.go
pawanpraka1: That was case introduced with strict-topology only (since 1.4). Which we can use to put strictness on the topology information to pass  fully qualified topology list.

If we want to change that dehavior and nothing will break because of this, then I can make that change but would prefer a separate PR for that as this PR focuses on "subset of the leys are not working". WDYT?

verult:
/lgtm

Just thinking out loud and throwing in another perspective: What happens on driver upgrade when the new version of the driver introduces a new key? Say there was an `AllowedTopologies` restricting only on the old key, and the admin forgets to update the StorageClass to include the new key. With the subset logic, k8s will match all previously matched nodes before the upgrade. IMO this is great and better than before, where the volume provision will fail.

pkg/controller/topology_test.go
msau42: Ok, let's fix it in a followup PR.

pkg/controller/topology.go
msau42: Was a test case added for this?

pawanpraka1: @msau42, Yes, I have modified the test case to cover this.

msau42: I'm not seeing the case where an implementation using "Contains" would have incorrectly allowed "rackA" when the allowedTopologies was "rackAB".

msau42:
Can you also add a release note for this?

pkg/controller/topology.go
pawanpraka1: It is the other way around @msau42, it will not allow which should be allowed in few cases. The problem is since in the hash we sort all the key value pair and generate one string, so contains logic may fail when becasue of sorting, something comes in between even though all the key values are present but not contiguously. Check the unit test case, I have modified it to cover this case.  With the conatins logic that will fail.

Storageclass
allowedTopologies:
- matchLabelExpressions:
  - key: com.example.csi/disk
    values:
      - ssd
  - key: com.example.csi/zone
    values:
      - zone1

and nodes are like this:- 

node1: {"com.example.csi/zone": "zone1", "com.example.csi/rack": "rack1", "com.example.csi/disk": "ssd"}

node2: {"com.example.csi/zone": "zone2", "com.example.csi/rack": "rack2", "com.example.csi/disk": "ssd"}

node3: {"com.example.csi/zone": "zone2", "com.example.csi/rack": "rack1", "com.example.csi/disk": "nvme"}


As per storage class node1 should qualify that but the two hashes will be like :
storageclass :  com.example.csi/disk#ssd,com.example.csi/zone#zone1
node1:- com.example.csi/disk#ssd,com.example.csi/rack#rack1,com.example.csi/zone#zone1

Since rack is coming in between, so contains will work here even though it qualifies the storageclass requirement.

pawanpraka1:
> Can you also add a release note for this?

done @msau42 .

pkg/controller/topology.go
msau42: I'm looking for a test case where:

allowedTopologies is something like:
```
					MatchLabelExpressions: []v1.TopologySelectorLabelRequirement{
						{
							Key:    "com.example.csi/disk",
							Values: []string{"ssd"},
						},
						{
							Key:    "com.example.csi/zone",
							Values: []string{"zone1"},
						},

```
and the node labels is something like:

```
{"com.example.csi/zone": "zone10", "com.example.csi/rack": "rack1", "com.example.csi/disk": "ssd"},
```

Note very carefully that the zone value in allowedTopologies is `zone1` and the zone value in the node label is `zone10`. They are different, but the string is a subset of the other.

The first iteration of this PR would have matched these two, which was incorrect. The current implementation correctly does not match them.

pawanpraka1: Yes, good to add that test case.

pawanpraka1: @msau42, @verult, added one negative test case for this.

msau42:
/lgtm
/approve
Thanks for working on this!

pawanpraka1:
@msau42  travis has passed but status has not updated yet, Please check.

msau42:
Can you try force pushing your branch again? I'm not sure why travis reporting is stuck

pawanpraka1:
@msau42 can you restart the travis build?

msau42:
I'm not able to restart it, I think it is triggered based on git pushes. Maybe try squashsing your commits and then force pushing?

pawanpraka1:
@msau42 done, forced push the changes.

pawanpraka1:
@msau42 same issue, again.

msau42:
I will just force merge the PR, all the other checks have passed. Thanks!

pawanpraka1:
thanks @msau42 .

Issue #414, opened 2020-02-26
Volume can be deleted in the middle of cloning


jsafrane:
We use finalizers to prevent volumes from being deleted while a **snapshot** is being taken. Should we use the same precaution to protect  volumes that are being **cloned**? Or do we assume that the cloning is faster / easier and storage backend will sort it out somehow?

@msau42 @xing-yang, what do you think?

xing-yang:
Actually snapshotting is usually faster than cloning if no upload is needed.  I think it is a good idea to add a finalizer to prevent the source volume from being deleted in cloning.

jsafrane:
Looking at the external provisioner, it's simple to add a finalizer to PVC that is being cloned.

But it's quite hard to remove it reliably. Here is a brief design:

* During `ProvisionExt()`, the provisioner adds a finalizer to PVC before calling cloning `CreateVolumeRequest`.

* CSI external-provisioner must watch PVCs via a new workqueue.
  * When PVC X has `DeletionTimestamp` && it has the finalizer:
      * If all PVCs that have X as its data source are *not* Pending, the finalizer is removed.
      * If there is a Pending PVC that has X as its data source (i.e. cloning may be in progress), the PVC is added back to the workqueue with exp. backoff, so the PVC is re-evaluated periodically, until it is cloned.
  * Therefore CSI external-provisioner must instantiate and run its own PV/PVC/StorageClass shared informers and pass them to provisioning library as options.
  * As side effect, if a PVC that has X as data source is wrong (has wrong storage class, size, ...), it effectively blocks X from being deleted, even though there is no cloning in progress.
    * Checking if cloning is in progress for a PVC is currently impossible, as the cloning may have been started by a previous provisioner instance that has just been restarted. Any in-memory cache will not survive this.

* During `ProvisionExt()`, the provisioner clones a PVC even if the source PVC has DeletionTimestamp: the cloning operation may have been started when the source PVC was "ok" and the operation timed out. The provisioner must call `CreateVolume` again to get the final error code or provisioned volume. [This is actually current behavior, posting here for completeness].

PR #422, opened 2020-03-11
Added controller for PVC deletion

Danil-Grigorev:
/kind feature

This PR addresses an attempt to split PVC handling between `external-provisioner` and `sig-storage-lib-external-provisioner`, also introducing a way to handle finalizers, related to in-progress PVC cloning.

PVC referenced in a data source is now updated with finalizer `provisioner.storage.kubernetes.io/cloning-protectionr` at the moment when a `ProvisionExt` method is called, used to dynamically provision a PV to which the content will be cloned at.

When cloning is finished - all PVC referencing the one as
a data source goes from `Pending` to `Bound` state, which allows the finalizer to be removed before deletion.
 
Fixes https://github.com/kubernetes-csi/external-provisioner/issues/414

```release-note
finalizers:
- provisioner.storage.kubernetes.io/cloning-protection

is now set on PVCs which are the source ones for the cloning,
preventing their removal before the cloning finishes.
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -255,15 +255,16 @@ func main() {
 	)
 
 	run := func(context.Context) {
-		factory.Start(context.Background().Done())
-		cacheSyncResult := factory.WaitForCacheSync(context.Background().Done())
+		stopCh := context.Background().Done()
+		factory.Start(stopCh)
+		cacheSyncResult := factory.WaitForCacheSync(stopCh)
 		for _, v := range cacheSyncResult {
 			if !v {
 				klog.Fatalf("Failed to sync Informers!")
 			}
 		}
 
-		go csiClaimController.Run(int(*workerThreads), context.Background().Done())
+		go csiClaimController.Run(int(*workerThreads), stopCh)
 		provisionController.Run(wait.NeverStop)
 	}
 
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -247,7 +247,7 @@ func main() {
 		provisionerOptions...,
 	)
 
-	csiClaimController := ctrl.NewCSIClaimController(
+	csiClaimController := ctrl.NewCSIController(
 		clientset,
 		claimLister,
 		claimInformer,
@@ -263,12 +263,10 @@ func main() {
 				klog.Fatalf("Failed to sync Informers!")
 			}
 		}
-		klog.Info("Starting 2PVC controller")
 
 		go csiClaimController.Run(int(*workerThreads), stopCh)
 		provisionController.Run(wait.NeverStop)
 	}
-	klog.Info("Starting 1PVC controller")
 
 	if !*enableLeaderElection {
 		run(context.TODO())
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -175,9 +175,10 @@ func main() {
 	factory := informers.NewSharedInformerFactory(clientset, ctrl.ResyncPeriodOfCsiNodeInformer)
 
 	// -------------------------------
-	// StorageClass lister
+	// Listers
 	// Create informer to prevent hit the API server for all resource request
 	scLister := factory.Storage().V1().StorageClasses().Lister()
+	claimLister := factory.Core().V1().PersistentVolumeClaims().Lister()
 
 	var csiNodeLister storagelistersv1beta1.CSINodeLister
 	var nodeLister v1.NodeLister
@@ -190,9 +191,7 @@ func main() {
 	// PersistentVolumeClaims informer
 	rateLimiter := workqueue.NewItemExponentialFailureRateLimiter(*retryIntervalStart, *retryIntervalMax)
 	claimQueue := workqueue.NewNamedRateLimitingQueue(rateLimiter, "claims")
-
 	claimInformer := factory.Core().V1().PersistentVolumeClaims().Informer()
-	claimLister := factory.Core().V1().PersistentVolumeClaims().Lister()
 
 	// Setup options
 	provisionerOptions := []func(*controller.ProvisionController) error{
@@ -256,16 +255,15 @@ func main() {
 	)
 
 	run := func(context.Context) {
-		stopCh := context.Background().Done()
-		factory.Start(stopCh)
-		cacheSyncResult := factory.WaitForCacheSync(stopCh)
+		factory.Start(context.Background().Done())
+		cacheSyncResult := factory.WaitForCacheSync(context.Background().Done())
 		for _, v := range cacheSyncResult {
 			if !v {
 				klog.Fatalf("Failed to sync Informers!")
 			}
 		}
 
-		go csiClaimController.Run(int(*workerThreads), stopCh)
+		go csiClaimController.Run(int(*workerThreads), context.Background().Done())
 		provisionController.Run(wait.NeverStop)
 	}
 
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -172,13 +172,37 @@ func main() {
 	timeStamp := time.Now().UnixNano() / int64(time.Millisecond)
 	identity := strconv.FormatInt(timeStamp, 10) + "-" + strconv.Itoa(rand.Intn(10000)) + "-" + provisionerName
 
+	factory := informers.NewSharedInformerFactory(clientset, ctrl.ResyncPeriodOfCsiNodeInformer)
+
+	// -------------------------------
+	// StorageClass lister
+	// Create informer to prevent hit the API server for all resource request
+	scLister := factory.Storage().V1().StorageClasses().Lister()
+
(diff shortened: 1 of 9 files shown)


pkg/controller/controller.go
```diff
@@ -531,6 +582,13 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 		req.VolumeContentSource = volumeContentSource
 	}
 
+	if options.PVC.Spec.DataSource != nil && rc.clone {
+		err = p.setCloneFinalizer(options.PVC.Spec.DataSource.Name, options.PVC.Namespace)
```
jsafrane: Is there reason you pass name/namespace instead of full PVC? `setCloneFinalizer` does not need to query API server at all.

pkg/controller/controller.go
```diff
@@ -997,6 +1070,150 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
...
+		var key string
+		var ok bool
+		if key, ok = obj.(string); !ok {
+			p.claimQueue.Forget(obj)
+			return fmt.Errorf("expected string in workqueue but got %#v", obj)
+		}
+
+		if err := p.syncClaimHandler(key); err != nil {
+			klog.Warningf("Retrying syncing claim %q after %v failures", key, p.claimQueue.NumRequeues(obj))
+			p.claimQueue.AddRateLimited(obj)
```
jsafrane: You should not call `Forget` below if you want the obj in the queue.

pkg/controller/controller.go
```diff
@@ -997,6 +1070,150 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
...
+	claim, ok := obj.(*v1.PersistentVolumeClaim)
+	if !ok {
+		return fmt.Errorf("expected claim but got %+v", obj)
+	}
+
+	if !checkFinalizer(claim, pvcCloneFinalizer) {
+		return nil
+	}
+
+	pvcList, err := p.client.CoreV1().PersistentVolumeClaims("").List(metav1.ListOptions{})
```
jsafrane: IMo you should use informer here, we don't want to load the API server with every PVC being cloned.

pkg/controller/controller.go
```diff
@@ -997,6 +1070,150 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
...
+	// Check for pvc state with DataSource pointing to claim
+	for _, pvc := range pvcList.Items {
+		if pvc.Spec.DataSource == nil {
+			continue
+		}
+
+		if pvc.Spec.DataSource.Kind == pvcKind &&
+			pvc.Spec.DataSource.Name == claim.Name &&
+			pvc.Status.Phase == v1.ClaimPending {
+			return apierrs.NewBadRequest(fmt.Sprintf("PVC '%s' is in pending state, cloning in progress", pvc.Name))
```
jsafrane: Why `apierrs.NewBadRequest`? Simple `fmt.Errorf()` would probably do.

pkg/controller/controller.go
```diff
@@ -997,6 +1070,150 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
...
+		}
+	}
+
+	// Remove clone finalizer
+	// need to get the pvc again because the delete has updated the object with a deletion timestamp
+	claim, err = p.client.CoreV1().PersistentVolumeClaims(claim.Namespace).Get(claim.Name, metav1.GetOptions{})
```
jsafrane: I don't understand why extra `Get()` is necessary. `claim` already has `DeletionTimestamp`. Yes, it could have changed since the informer got update event from API server, but it could have changed also since you get it from API server here and use the information few lines below.

pkg/controller/controller.go
```diff
@@ -997,6 +1070,150 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
...
+
+	finalizers := make([]string, 0)
+	for _, finalizer := range claim.ObjectMeta.Finalizers {
+		if finalizer != pvcCloneFinalizer {
+			finalizers = append(finalizers, finalizer)
+		}
+	}
+
+	// Only update the finalizers if we actually removed something
+	if checkFinalizer(claim, pvcCloneFinalizer) {
```
jsafrane: You already checked the finalizer above.

pkg/controller/controller.go
```diff
@@ -311,6 +325,43 @@ func NewCSIProvisioner(client kubernetes.Interface,
 	return provisioner
 }
 
+// NewCSIController creates new controller for additional CSI capabilities
```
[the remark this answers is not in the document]
jsafrane: And maybe call it CloningProtectionController, until someone adds another responsibility to it?

pkg/controller/clone_controller_test.go
```diff
@@ -0,0 +1,184 @@
...
+			)
+
+			go cloningProtector.Run(1, context.TODO().Done())
+
+			// Simulate Delete behavior
+			claim := tc.cloneSource.(*v1.PersistentVolumeClaim)
+			clientSet.CoreV1().PersistentVolumeClaims(claim.Namespace).Update(pvcDeletionMarked(claim))
+
+			// Wait for couple reconciles for controller to adjust finalizers
+			time.Sleep(2 * time.Second)
```
jsafrane: Is it necessary to wait for a couple of reconciles? For an unit test we typically call `syncClaim()` and don't run whole controller loop with workqueues.

(Yes, there will be some functions not covered by unit tests.)

Danil-Grigorev: No, not at all. Will fix this, thanks for suggestion. It will also allow to catch an error, which forces the reconcile to run again.

pkg/controller/controller.go
Danil-Grigorev: Fixed.

pkg/controller/controller.go
Danil-Grigorev: Thanks, didn't notice that!

pkg/controller/controller.go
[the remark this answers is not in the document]
Danil-Grigorev: Sounds good, just added tests which should describe the behaviour and moved it all to a separate file.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -234,16 +247,23 @@ func main() {
...
 
+	csiClaimController := ctrl.NewCloningProtectionController(
+		clientset,
+		claimLister,
+		claimInformer,
+		claimQueue,
+	)
+
 	run := func(context.Context) {
-		stopCh := context.Background().Done()
```
jsafrane: Why did you remove `stopCh`?

pkg/controller/clone_controller_test.go
```diff
@@ -0,0 +1,226 @@
...
+		},
+		"delete source pvc when destination pvc status is claim pending": {
+			cloneSource: pvcFinalizers(baseClaim(), pvcCloneFinalizer),
+			initialClaims: []runtime.Object{
+				pvcPhase(
+					pvcDataSourceClone(
+						pvcNamed(baseClaim(), dstName),
+						srcName,
+					),
+					v1.ClaimPending)},
```
jsafrane: It would be more readable if you put `claim` as the last argument - function name + its arguments would be on the same line:

```
pvcPhase(v1.ClaimPending,
    pvcDataSourceClone(srcName,
        pvcNamed(dstName, baseClaim())
```

pkg/controller/controller.go
```diff
@@ -134,6 +134,8 @@ const (
 	annStorageProvisioner = "volume.beta.kubernetes.io/storage-provisioner"
 
 	snapshotNotBound = "snapshot %s not bound"
+
+	pvcCloneFinalizer = "external-provisioner.clone.kubernetes.io/finalizer"
```
jsafrane: provisioner.storage.kubernetes.io/cloning-protection

cmd/csi-provisioner/csi-provisioner.go
Danil-Grigorev: Accidentally left this change after seeing issues with a similar approach in tests, will undo this.

jsafrane:
Good work and thanks a lot!
/lgtm
/approve

pkg/controller/clone_controller.go
```diff
@@ -0,0 +1,206 @@
...
+	defer utilruntime.HandleCrash()
+	defer p.claimQueue.ShutDown()
+
+	claimHandler := cache.ResourceEventHandlerFuncs{
+		AddFunc:    p.enqueueClaimUpadate,
+		UpdateFunc: func(_ interface{}, newObj interface{}) { p.enqueueClaimUpadate(newObj) },
+	}
+	p.claimInformer.AddEventHandlerWithResyncPeriod(claimHandler, controller.DefaultResyncPeriod)
+
+	for i := 0; i < threadiness; i++ {
```
msau42: This is going to double the number of goroutines, and default value of worker-threads is 100 (so 200) total. Do we really need that many threads to process this finalizer? @jsafrane

jsafrane: hm, maybe not...

Danil-Grigorev: I'll address this in a follow up PR. Sorry for the issue. Could 5 be sufficient @msau42?

PR #425, opened 2020-03-17
extraCreateMetadata support labels/annotations

fredkan:
**What type of PR is this?**
>
> /kind feature

> Currently, CSI controller cannot get any extra configuration from pvc. But in some situations, user need define configurations in pvc, not only in storageclass.
>
> extraCreateMetadata provide the way to add extra pvc metadata to controller server, but not labels or annotations. This feature is to achieve that.

```release-note
no need to change current configuration. extraCreateMetadata is added info for controller server. No effect for plugin if user not use it.
```


msau42:
Thanks for the PR, but we explicitly don't allow this today because it breaks PVC portability across storage providers. See https://github.com/kubernetes-csi/external-provisioner/issues/86 for further discussion.

PR #434, opened 2020-04-16
Start clone protection only when capability matches

### LINKED ISSUE #431: Disable the clone controller when the driver does not have CLONE_VOLUME capability
/type bug

If the driver doesn't have the `CLONE_VOLUME` controller capability, the clone controller shouldn't need to run

Danil-Grigorev:
**What type of PR is this?**

 /kind bug

Adds a capability check, to prevent cloning protection from starting, when `CLONE_VOLUME` is not set on the driver.

Fixes #431 

```release-note
NONE
```


msau42:
Thank you for the fix! Can we also cherry pick this?

PR #438, opened 2020-05-13
Add VolumeAttachment Lister to CSI Provisioner

RaunakShah:
This change prevents calling DeleteVolume on the CSI plugin for volumes that are still attached to a kubernetes node.

**What type of PR is this?**

/kind bug

Currently, external-provisioner does not check if a volume is attached to a node before trying to delete it. 
This can happen when there's a race between Pod and PVC deletion (like deleting a ns). 
NOTE: PVC protection does not apply here because the Pod is deleted from k8s, but the volume is still attached. 

Based on discussion in https://github.com/kubernetes/kubernetes/issues/84226, this change adds a volumeattachment lister to verify if a volume is still attached to a node before attempting to delete it. 

Testing:

Deleting a volume that is not attached to a node works as expected:

```
{"log":"I0513 18:32:21.365637       1 controller.go:1422] delete \"pvc-80f3eaf0-e4f5-487a-a048-1e884a5ee13b\": started\n","stream":"stderr","time":"2020-05-13T18:32:21.372039744Z"}
{"log":"I0513 18:32:23.737215       1 controller.go:1450] delete \"pvc-80f3eaf0-e4f5-487a-a048-1e884a5ee13b\": volume deleted\n","stream":"stderr","time":"2020-05-13T18:32:23.739834928Z"}
{"log":"I0513 18:32:23.758050       1 controller.go:1496] delete \"pvc-80f3eaf0-e4f5-487a-a048-1e884a5ee13b\": persistentvolume deleted\n","stream":"stderr","time":"2020-05-13T18:32:23.758428374Z"}
{"log":"I0513 18:32:23.759194       1 controller.go:1498] delete \"pvc-80f3eaf0-e4f5-487a-a048-1e884a5ee13b\": succeeded\n","stream":"stderr","time":"2020-05-13T18:32:23.759331067Z"}
```

Deleting the namespace and creating a race between detach and delete:

```
{"log":"I0513 18:41:53.980570       1 controller.go:1422] delete \"pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0\": started\n","stream":"stderr","time":"2020-05-13T18:41:53.98129317Z"}
{"log":"E0513 18:41:53.986309       1 controller.go:1445] delete \"pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0\": volume deletion failed: persistentvolume pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0 is still attached to node k8s-node-0572\n","stream":"stderr","time":"2020-05-13T18:41:53.986566278Z"}
{"log":"W0513 18:41:53.986780       1 controller.go:965] Retrying syncing volume \"pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0\", failure 0\n","stream":"stderr","time":"2020-05-13T18:41:53.986890274Z"}
{"log":"E0513 18:41:53.987122       1 controller.go:983] error syncing volume \"pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0\": persistentvolume pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0 is still attached to node k8s-node-0572\n","stream":"stderr","time":"2020-05-13T18:41:53.987286288Z"}
{"log":"I0513 18:41:53.987990       1 event.go:281] Event(v1.ObjectReference{Kind:\"PersistentVolume\", Namespace:\"\", Name:\"pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0\", UID:\"79fa52d2-ccdd-49cf-a3ed-e5eb081528af\", APIVersion:\"v1\", ResourceVersion:\"5383\", FieldPath:\"\"}): type: 'Warning' reason: 'VolumeFailedDelete' persistentvolume pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0 is still attached to node k8s-node-0572\n","stream":"stderr","time":"2020-05-13T18:41:53.988209321Z"}
...
{"log":"I0513 18:42:01.006230       1 controller.go:1422] delete \"pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0\": started\n","stream":"stderr","time":"2020-05-13T18:42:01.007199938Z"}
{"log":"I0513 18:42:03.968729       1 controller.go:1450] delete \"pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0\": volume deleted\n","stream":"stderr","time":"2020-05-13T18:42:03.969879653Z"}
{"log":"I0513 18:42:04.025931       1 controller.go:1496] delete \"pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0\": persistentvolume deleted\n","stream":"stderr","time":"2020-05-13T18:42:04.02633573Z"}
{"log":"I0513 18:42:04.025978       1 controller.go:1498] delete \"pvc-fce3b92c-093f-4a74-86f7-1fc9cb6d3be0\": succeeded\n","stream":"stderr","time":"2020-05-13T18:42:04.026375078Z"}
```

Fixes https://github.com/kubernetes/kubernetes/issues/84226

```release-note
Add VolumeAttachment Lister to  prevent calling DeleteVolume on the CSI plugin for volumes that are still attached to a kubernetes node.
```

### DIFF
--- a/deploy/kubernetes/rbac.yaml
+++ b/deploy/kubernetes/rbac.yaml
@@ -50,6 +50,9 @@ rules:
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["volumeattachments"]
+    verbs: ["get", "list", "watch"]
 
 ---
 kind: ClusterRoleBinding
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -38,6 +38,7 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	_ "k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
@@ -219,6 +220,7 @@ type csiProvisioner struct {
 	csiNodeLister                         storagelistersv1.CSINodeLister
 	nodeLister                            corelisters.NodeLister
 	claimLister                           corelisters.PersistentVolumeClaimLister
+	vaLister                              storagelistersv1.VolumeAttachmentLister
 	extraCreateMetadata                   bool
 }
 
@@ -284,6 +286,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 	csiNodeLister storagelistersv1.CSINodeLister,
 	nodeLister corelisters.NodeLister,
 	claimLister corelisters.PersistentVolumeClaimLister,
+	vaLister storagelistersv1.VolumeAttachmentLister,
 	extraCreateMetadata bool,
 ) controller.Provisioner {
 
@@ -307,6 +310,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 		csiNodeLister:                         csiNodeLister,
 		nodeLister:                            nodeLister,
 		claimLister:                           claimLister,
+		vaLister:                              vaLister,
 		extraCreateMetadata:                   extraCreateMetadata,
 	}
 	return provisioner
@@ -1006,7 +1010,6 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
 	req := csi.DeleteVolumeRequest{
 		VolumeId: volumeId,
 	}
-
 	// get secrets if StorageClass specifies it
 	storageClassName := util.GetPersistentVolumeClass(volume)
 	if len(storageClassName) != 0 {
@@ -1034,6 +1037,18 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
 	defer cancel()
+
+	// Verify if volume is attached to a node before proceeding with deletion
+	vaList, err := p.vaLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list volumeattachments: %v", err)
+	}
+
+	for _, va := range vaList {
+		if va.Spec.Source.PersistentVolumeName != nil && *va.Spec.Source.PersistentVolumeName == volume.Name {
+			return fmt.Errorf("persistentvolume %s is still attached to node %s", volume.Name, va.Spec.NodeName)
+		}
+	}
 
 	_, err = p.csiClient.DeleteVolume(ctx, &req)
 
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -409,7 +409,7 @@ func TestCreateDriverReturnsInvalidCapacityDuringProvision(t *testing.T) {
 
 	pluginCaps, controllerCaps := provisionCapabilities()
 	csiProvisioner := NewCSIProvisioner(nil, 5*time.Second, "test-provisioner", "test",
-		5, csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, csitrans.New(), nil, nil, nil, nil, false)
+		5, csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, csitrans.New(), nil, nil, nil, nil, nil, false)
 
 	// Requested PVC with requestedBytes storage
 	deletePolicy := v1.PersistentVolumeReclaimDelete
@@ -1674,7 +1674,7 @@ func runProvisionTest(t *testing.T, k string, tc provisioningTestcase, requested
 
 	pluginCaps, controllerCaps := provisionCapabilities()
 	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5, csiConn.conn,
-		nil, provisionDriverName, pluginCaps, controllerCaps, supportsMigrationFromInTreePluginName, false, csitrans.New(), nil, nil, nil, nil, tc.withExtraMetadata)
+		nil, provisionDriverName, pluginCaps, controllerCaps, supportsMigrationFromInTreePluginName, false, csitrans.New(), nil, nil, nil, nil, nil, tc.withExtraMetadata)
 
 	out := &csi.CreateVolumeResponse{
 		Volume: &csi.Volume{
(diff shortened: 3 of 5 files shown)


pkg/controller/controller.go
```diff
@@ -1035,6 +1038,18 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
...
 	defer cancel()
 
+	// Verify if volume is attached to a node before proceeding with deletion
+	vaList, err := p.vaLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list volumeattachments: %v", err)
+	}
+
+	for _, va := range vaList {
+		if va.Spec.Source.PersistentVolumeName != nil && *va.Spec.Source.PersistentVolumeName == volume.Name && va.Status.Attached {
```
msau42, line 1048: We don't need to check status. Just the presence of the object is enough to not proceed.

pkg/controller/controller_test.go
```diff
@@ -2838,6 +2840,47 @@ func TestDelete(t *testing.T) {
...
+			},
+			storageClass: &storagev1.StorageClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "sc-name",
+				},
+				Parameters: map[string]string{
+					prefixedProvisionerSecretNameKey: "static-${pv.name}-${pvc.namespace}-${pvc.name}",
+				},
+			},
+			volumeAttachment: &storagev1.VolumeAttachment{
```
msau42, line 2868: Can you also add a test case where there's a VolumeAttachment that exists for a different PV?

msau42:
cc @jsafrane

xing-yang:
I see a few e2e tests failed complaining about PVs not deleted.  We probably need to fix the e2e tests first.

deploy/kubernetes/rbac.yaml
```diff
@@ -50,6 +50,9 @@ rules:
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["volumeattachments"]
+    verbs: ["list"]
```
jsafrane, line 55: Should there be also `"watch"`? I hope `vaLister.List()` below does not list all VolumeAttachments from the API server.

pkg/controller/controller.go
RaunakShah: done

pkg/controller/controller_test.go
RaunakShah: done

deploy/kubernetes/rbac.yaml
RaunakShah: Ah you're right, we only need to watch on the API server and list from the cache. However i'm wondering do we still need a List/Get to populate the cache?

RaunakShah:
Seeing if the e2e failure was intermittent or because of this change.
/retest

deploy/kubernetes/rbac.yaml
jsafrane: `list` is needed to populate the cache, I am not sure about `get`, IMO it won't harm to have it in the RBAC - all the other objects have it.

RaunakShah: done!

pkg/controller/controller_test.go
```diff
@@ -2838,6 +2840,47 @@ func TestDelete(t *testing.T) {
 			},
 			expectErr: true,
 		},
+		"fail - delete when attached to node": {
```
msau42, line 2843: Can you add 2 more test case like this, except:
- the VA.status.attached = false
- the VA has a deletion timestamp on it

Both of these cases should still behave as if the volume is still attached.

RaunakShah: done!

pkg/controller/controller.go
```diff
@@ -1035,6 +1038,18 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
 	ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
 	defer cancel()
 
+	// Verify if volume is attached to a node before proceeding with deletion
+	vaList, err := p.vaLister.List(labels.Everything())
```
SandeepPissay, line 1042: VolumeAttachment objects will be in thousands in a scaled environment. Listing all VA objects is quite expensive. Can we avoid such expensive call to API server?

jsafrane: `vaLister` lists VolumeAttachments from in-memory informer cache. "Thousands" should be ok, "millions" could be problematic.

SandeepPissay: Does informer limit how many objects are in the cache? How are we sure that all VolumeAttachment objects will fit into memory? If the lister ends up hitting the API server then this will introduce a performance regression, hence I'm a bit concerned with this code.

pohly:
I am seeing tests failures in PMEM-CSI with the new external-provisioner 2.0.0 because of "still attached to node" in a situation where NodeUnpublishVolume and NodeUnstageVolume have completed successfully, i.e. the volume shouldn't be attached anymore. This goes on for 5 minutes until the test gives up and redeploys the driver, which involves restarting the external-provisioner. The new instance then *does* try to delete. I wonder whether the informer can have an outdated VolumeAttachment object for over five minutes?

pohly:
> I see a few e2e tests failed complaining about PVs not deleted. We probably need to fix the e2e tests first.

Is that problem truly gone? The PMEM-CSI tests are passing locally, but not in the CI, which implies that it depends a bit on timing.

pkg/controller/controller.go
```diff
@@ -1035,6 +1038,18 @@ func (p *csiProvisioner) Delete(volume *v1.PersistentVolume) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to list volumeattachments: %v", err)
+	}
+
+	for _, va := range vaList {
+		if va.Spec.Source.PersistentVolumeName != nil && *va.Spec.Source.PersistentVolumeName == volume.Name {
+			return fmt.Errorf("persistentvolume %s is still attached to node %s", volume.Name, va.Spec.NodeName)
```
pohly, line 1049: Shouldn't this code be checking that `va.Status.Attached` is true?

As it stands, it checks for "is there a VolumeAttachment object for the PV", which is a bit broader than "is the PV attached". I don't know whether that makes a difference.

RaunakShah: Removed that check based on @msau42 comment - https://github.com/kubernetes-csi/external-provisioner/pull/438#discussion_r424775727
IIRC the precedent was another comment in another PR that the existence of `volumeattachment` was enough to fail the operation..

pohly: I think I understand now why: a ControllerPublishVolume call might have failed, in which case the call must be retried or ControllerUnpublishVolume needs to be called (according to the CSI spec). So it's correct to prevent DeleteVolume even when the volume is not really attached.

PR #442, opened 2020-06-04
Ignore unsupported DataSource in PVC

### LINKED ISSUE #441: Unsupported DataSource in PVC is not properly handled
If DataSource is neither PVC nor VolumeSnapshot, the provisioner goes ahead and creates an empty volume and waits for ContentSource to become not nil. Since the introduction of AnyVolumeDataSource, it is possible that we have such a DataSource, however, there is no way for a CSI Driver to set the VolumeContentSource to a type that is not supported by the CSI spec.  As a result, the provisioner will keep creating an empty volume, deleting it, and do it over and over again.

We should either allow such a DataSource and create an empty volume or return a final error early.

xing-yang:
**What type of PR is this?**
/kind bug

If DataSource is neither PVC nor VolumeSnapshot, the provisioner goes ahead and creates an empty volume and waits for ContentSource to become not nil. Since the introduction of AnyVolumeDataSource, it is possible that we have such a DataSource, however, there is no way for a CSI Driver to set the ContentSource to a type that is not supported by the CSI spec.  As a result, the provisioner will keep creating an empty volume, deleting it, and do it over and over again.

This PR allows an empty volume to be provisioned and allows a generic data populator to populate data to the volume after it is provisioned when that becomes available.

An alternative fix is to fail early and return a final error and controller.ProvisioningFinished when an DataSource type is neither PVC nor VolumeSnapshot.

Fixes #441 

```release-note
Allow an empty volume to be created if the DataSource is not VolumeSnapshot and PVC.
```


xing-yang:
Looking at @bswartz's sample data populator implementation, I think it is probably better for the external-provisioner not to create an empty volume and just ignore the request and return controller.ProvisioningFinished without an error. 
    https://github.com/NetApp/hello-populator/blob/master/controller.go

CC @msau42

msau42:
/assign @saad-ali 
I agree I think the behavior that we want is going to depend on what the design for data populators is going to look like.

I do think provisioning an empty volume and returning may cause problems because the moment we create the PV object, it can become bound and used by a Pod.

bswartz:
Yes, this is the main sticking point in the data populator design, and I've been struggling with it.

What we would like is for populators to work seamlessly, so that from the user's perspective there is no difference between data sources that the provisioner sidecar understands, vs data source that are handled externally by other populator controllers. However, we still need to reliably return errors when what the user is trying to do isn't going to work, so they don't get confused or frustrated.

Kubernetes makes this hard because of it's declarative model, where it's valid to define dependents before their dependencies. Whether using a particular data source is an error then changes over time.

I suspect the best we can do is to develop some kind of registry of data sources that should work at any given time within a cluster, and to consult that registry at provisioning time. Unregistered data sources should cause error immediately, because there's no reason to believe those requests will ever succeed, even though they conceivably could succeed later if someone installed a controller to handle them. Registered ones should not cause an error, and the thing that registered the data source kind should be responsible for taking a look at the request and either acting on it or generating an error, if appropriate.

I'm still thinking about what kind of registry could work here, and what the registration process could look like. It would have to be dynamic and specific to the cluster, and its contents will ultimately depend on what the admin installs on the cluster in the way of populators.

msau42:
We have a similar issue with external provisioners. The way we dealt with it was just to have an event that says "waiting for external provisioner X to provision". We can do something similar here.

bswartz:
I think that kind of event is necessary but not sufficient. Yes we should log when we see something we don't know how to respond to so the user knows what stage of the process that request has reached, but there's a big difference between knowing that the provisioner sidecar decided not to act on the request, and knowing that something else is in the system has promised to act on it vs possibly the request never getting acted on.

The big problem we're going to see as soon as there are real external data populators in use is that they'll only be installed on some minority of actual k8s clusters, and users won't have any good way to know if they're present or not on the cluster they're trying to use. We already see this problem with snapshots, where some clusters support them and other's don't, and it's nontrivial to figure out which case your cluster is.

xing-yang:
I agree that the event is not sufficient, but making this change will be a step forward and better than the current situation.  Currently the external-provisioner will create an empty volume and then delete it because VolumeContentSource is not set by the CSI driver and it will keep doing this forever.

pkg/controller/controller.go
```diff
@@ -475,7 +475,11 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 		case pvcKind:
 			rc.clone = true
 		default:
-			klog.Infof("Unsupported DataSource specified (%s), the provisioner won't act on this request", options.PVC.Spec.DataSource.Kind)
+			klog.Infof("DataSource specified (%s) is not supported by the provisioner, waiting for an external data populator to create the volume", options.PVC.Spec.DataSource.Kind)
+			// DataSource is not VolumeSnapshot and PVC
+			// Wait for an external data populator to create the volume
+			//controller.eventRecorder.Event(options.PVC, v1.EventTypeNormal, "Provisioning", fmt.Sprintf("Waiting for a volume to be created by an external data populator"))
+			return nil, controller.ProvisioningFinished, nil
```
msau42, line 482: Can we add a unit test for this?

pkg/controller/controller.go
```diff
@@ -475,7 +475,11 @@ func (p *csiProvisioner) ProvisionExt(options controller.ProvisionOptions) (*v1.
 		case pvcKind:
 			rc.clone = true
 		default:
-			klog.Infof("Unsupported DataSource specified (%s), the provisioner won't act on this request", options.PVC.Spec.DataSource.Kind)
+			klog.Infof("DataSource specified (%s) is not supported by the provisioner, waiting for an external data populator to create the volume", options.PVC.Spec.DataSource.Kind)
+			// DataSource is not VolumeSnapshot and PVC
+			// Wait for an external data populator to create the volume
+			//controller.eventRecorder.Event(options.PVC, v1.EventTypeNormal, "Provisioning", fmt.Sprintf("Waiting for a volume to be created by an external data populator"))
```
msau42, line 481: should we add the event? Even if it's not the complete solution, it's more end-user facing than a log.

pkg/controller/controller.go
xing-yang: Sure

pkg/controller/controller.go
xing-yang: Yeah, I'd like to add an event.  I just realized that eventRecorder is defined in the sig-storage-lib-external-provisioner controller, but not in the provisioner controller.  So I'll add a new eventRecorder param in the provisioner controller.

pkg/controller/controller.go
xing-yang: Done

pkg/controller/controller.go
```diff
@@ -289,6 +293,11 @@ func NewCSIProvisioner(client kubernetes.Interface,
 	vaLister storagelistersv1.VolumeAttachmentLister,
 	extraCreateMetadata bool,
 ) controller.Provisioner {
+	broadcaster := record.NewBroadcaster()
+	broadcaster.StartLogging(klog.Infof)
+	broadcaster.StartRecordingToSink(&corev1.EventSinkImpl{Interface: client.CoreV1().Events(v1.NamespaceAll)})
+	var eventRecorder record.EventRecorder
```
msau42, line 299: In this case, we don't need to declare it before initialization

PR #446, opened 2020-06-18
Increase timeout for create/delete volume operation

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Increase timeout for create/delete volume operation to reduce round trips of errors and retries

Fixes # 445

```release-note
Increase timeout for create/delete volume operation to reduce round trips of errors and retries
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -60,7 +60,7 @@ var (
 	retryIntervalMax     = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval of failed provisioning or deletion.")
 	workerThreads        = flag.Uint("worker-threads", 100, "Number of provisioner worker threads, in other words nr. of simultaneous CSI calls.")
 	finalizerThreads     = flag.Uint("cloning-protection-threads", 1, "Number of simultaniously running threads, handling cloning finalizer removal")
-	operationTimeout     = flag.Duration("timeout", 10*time.Second, "Timeout for waiting for creation or deletion of a volume")
+	operationTimeout     = flag.Duration("timeout", 120*time.Second, "Timeout for waiting for creation or deletion of a volume")
 	_                    = deprecatedflags.Add("provisioner")
 
 	enableLeaderElection = flag.Bool("leader-election", false, "Enables leader election. If leader election is enabled, additional RBAC rules are required. Please refer to the Kubernetes CSI documentation for instructions on setting up these RBAC rules.")


cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -60,7 +60,7 @@ var (
 	retryIntervalMax     = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval of failed provisioning or deletion.")
 	workerThreads        = flag.Uint("worker-threads", 100, "Number of provisioner worker threads, in other words nr. of simultaneous CSI calls.")
 	finalizerThreads     = flag.Uint("cloning-protection-threads", 1, "Number of simultaniously running threads, handling cloning finalizer removal")
-	operationTimeout     = flag.Duration("timeout", 10*time.Second, "Timeout for waiting for creation or deletion of a volume")
+	operationTimeout     = flag.Duration("timeout", 120*time.Second, "Timeout for waiting for creation or deletion of a volume")
```
msau42, line 63: Do we think 2 minutes is a reasonable default for most plugins? I suspect that 10 seconds is definitely too small for most plugins, but I don't really have much insight into other plugins to tell if 2 minutes is reasonable.

saikat-royc: That is a valid question. From GCE Persistent Disk point of view 2 mins is a reasonable timeout. However I am not sure about other plugins.

Another alternative to reduce the scope of the change would be to change the default timeout value by updating glag value here : https://github.com/saikat-royc/gcp-compute-persistent-disk-csi-driver/blob/master/deploy/kubernetes/base/controller.yaml

Thoughts?

msau42: For the immediate term, setting an explicit timeout in the driver is preferable.

We can continue to have the conversation if increasing the default timeout in general is desirable. This next release of external-provisioner is going to be a major release, so if we do want to change defaults, now is the time, although we may want to consider making the defaults the same across all sidecars? cc @jsafrane @xing-yang for thoughts

xing-yang: 10 sec is definitely too short.  We set the default timeout value in our driver deployment yaml file to 5 min for all sidecars.  We decided on that value after scale testing.

humblec: Indeed 10 seconds is definitely short, we mitigate the impact ( high number of retries) in our CSI driver( ceph) with adjusting this timeout option to a value of `150s`. That may be bit over as our case study says so, however considering the return from driver triggered as soon as its done, we shouldn't be worrying too much about the adjustment in the sidecar default. Also,  `+1` for keeping this consistent across the sidecars.

saikat-royc: The next steps agreed upon is to increase the timeouts in the csi driver sidecar deploy yamls

PR #449, opened 2020-06-19
Make driverName configurable in general functions like createFakePVC

humblec:
At present, createFakePVC call make use of global drivername "test-driver"
while creating a fake PVC. This is not flexible if we want to use this common
createFakePVC with a different drivername. This patch address the issue.
This patch also address one other inconsistency of variable name: ie
requestedBytes Vs requestBytes.
The callers make use of variable `requestedBytes` and callees( create*) functions
make use of `requestBytes` in its arg name for consistency.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -423,7 +423,7 @@ func TestCreateDriverReturnsInvalidCapacityDuringProvision(t *testing.T) {
 			Parameters:    map[string]string{},
 		},
 		PVName: "test-name",
-		PVC:    createFakePVC(requestedBytes),
+		PVC:    createFakePVC(requestedBytes, driverName),
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/controller_test.go
```diff
@@ -423,7 +423,7 @@ func TestCreateDriverReturnsInvalidCapacityDuringProvision(t *testing.T) {
 			Parameters:    map[string]string{},
 		},
 		PVName: "test-name",
-		PVC:    createFakePVC(requestedBytes),
+		PVC:    createFakePVC(requestedBytes, driverName),
```
msau42, line 426: What unit tests do we have that use different driver names?

Issue #374, opened 2019-10-26
Cache cluster topologies


msau42:
Once we move CSINode and Node to informers, we could potentially cache the topologies available in the cluster instead of having to List() and iterate through them every time we provision.

/kind feature

mucahitkurt:
@msau42 I think this refers to [these lines of codes](https://github.com/kubernetes-csi/external-provisioner/blob/0c56ab07a68cbe7b186d96f2afa8e00b5895066b/pkg/controller/topology.go#L372-L391)?

If I understand correctly, we would like to cache topology key/value pairs for a cluster, and they are all node labels, so we should cache all node labels and then select according to requested topology keys.

This will make unnecessary to extract requested topology keys from nodes that have at least one requested topology key, we may able to select values of requested topology keys from the informer with one list method call.

But I'm not sure about  benefits vs complexity tradeoff since we already use the informer for Nodes.

msau42:
Yes, basically anytime we have to list, like [here](https://github.com/kubernetes-csi/external-provisioner/blob/0c56ab07a68cbe7b186d96f2afa8e00b5895066b/pkg/controller/topology.go#L321) also.

The performance benefit should theoretically show up when: 1) you have a lot of nodes and 2) you have very few topologies compared to nodes (like zones). But agree, it may be good to write a performance benchmark test first so we can compare the performance when you have to List() nodes every time you provision vs listing topologies from a cache

PR #450, opened 2020-07-01
storage capacity producer

pohly:
**What type of PR is this?**
/kind feature

This is the producer side of https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1472-storage-capacity-tracking

TODO:
- [x] decide about command line interface
- [x] release notes
- [x] documentation
- [x] more tests
- [x] import Kubernetes 1.19.0 beta + sig-storage-lib-external-provisioner v6.0.0; right now I am using redirects instead
- [x] ignore storage classes with immediate binding, because CSIStorageCapacity information is not used for those anyway?
- [x] support making a Deployment the owner by walking up the owner chain

```release-note
optionally publish storage capacity information (alpha feature)
```


pkg/features/features.go
```diff
@@ -26,6 +26,10 @@ const (
 	// alpha: v0.4
 	// beta: v1.2
 	Topology featuregate.Feature = "Topology"
+
+	// owner: @pohly
+	// alpha: 2.0.0
+	CentralCapacity featuregate.Feature = "CentralCapacity"
```
pohly, line 32: In the KEP, I had described separate command line options for enabling the different storage capacity tracking modes. Conceptually those were similar (and inspired) by the feature mechanism in Kubernetes, with a single command line option taking different strings that form a set.

While working on the implementation, I found that external-provisioner is already using the Kubernetes feature infrastructure. I'd like to propose that we just use that also for the new feature capacity tracking. Each mode of producing CSIStorageCapacity then maps to a feature. Sp this is now the new option:
```
--feature-gates CentralCapacity=on
```

@jsafrane: what do you think?

pkg/capacity/topology/doc.go
```diff
@@ -0,0 +1,133 @@
...
+limitations under the License.
+*/
+
+// Package topology contains an abstract interface for discovering
+// topology segments for a storage backend and a specific implementation
+// which does that based on the CSINodeDriver.TopologyKeys and the
+// corresponding labels for the nodes.
+package topology
+
+import (
```
jsafrane, line 23: Anything below this line does not belong to `topology/doc.go`

pkg/features/features.go
jsafrane: Feature gates is for code that's graduating and they are eventually removed. If I understand the proposal correctly, we  still need a parameter that tells which capacity mode to use when the feature is GA and we remove its feature gate (if we have one).

IMO, we don't really need feature gate on this, we may add `-enable-capacity=central` and mark it as alpha.

go.mod
```diff
@@ -7,53 +7,70 @@ require (
...
-replace k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.17.0
+replace k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.19.0-beta.2
 
-replace k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.17.0
+replace k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.19.0-beta.2
 
-replace k8s.io/sample-controller => k8s.io/sample-controller v0.17.0
+replace k8s.io/sample-controller => k8s.io/sample-controller v0.19.0-beta.2
+
+replace k8s.io/api => /nvme/gopath/src/k8s.io/kubernetes/staging/src/k8s.io/api
```
jsafrane, line 60: `/nvme` looks like some debugging hack.

pkg/capacity/topology/doc.go
pohly: Okay. I wasn't sure whether doc.go was just for package-level documentation or also for interfaces of the package. I'll move it.

go.mod
pohly: That's the WIP part: I need it to compile against the modified Kubernetes until the API gets merged there.

If you want to compile yourself, then you can replace this something like - beware, untested:
```
go mod edit -replace k8s.io/api=github.com/pohly/kubernetes@csi-storage-capacity
```

pohly: Nope, that won't work. Better just check out my branch somewhere and edit the go.mod file.

pkg/features/features.go
pohly: Okay. So should I use a simple boolean `-enable-capacity-central` or `-enable-capacity=central`? I have the code for the latter in https://github.com/kubernetes-csi/external-provisioner/pull/450/files#diff-513b7bafc46ec7994a7f0bc31fd799cc

go.mod
jsafrane: I ended up deleting go.mod completely and my editor uses stuff from `vendor/`

pkg/features/features.go
jsafrane: I like the `-enable-capacity=central`, as described in the KEP.

pkg/capacity/controller.go
```diff
@@ -0,0 +1,506 @@
...
+		capacity, err = c.client.StorageV1alpha1().CSIStorageCapacities().Update(ctx, capacity, metav1.UpdateOptions{})
+		if err != nil && apierrs.IsConflict(err) {
+			// Handle the case where we had a stale copy of the object. Can only happen
+			// when someone else was making changes to it, which should be rare.
+			capacity, err = c.client.StorageV1alpha1().CSIStorageCapacities().Get(ctx, capacity.Name, metav1.GetOptions{})
+			if err != nil {
+				return fmt.Errorf("getting fresh copy of CSIStorageCapacity for %+v: %v", item, err)
+			}
+			capacity.Capacity = quantity
+			capacity, err = c.client.StorageV1alpha1().CSIStorageCapacities().Update(ctx, capacity, metav1.UpdateOptions{})
```
jsafrane, line 410: We usually propagate `Conflict` errors back to the workqueue, it gets retried after exp. backoff or when event that CSIStorageCapacity was updated arrives (if you had informer on CSIStorageCapacity...)

Or there is `RetryOnConflict()` if you really want to handle it here. https://github.com/kubernetes/client-go/blob/6f3bc4ba9215358ea3ceff566cff7480a5d23a7b/util/retry/util.go#L68

pkg/capacity/controller.go
```diff
@@ -0,0 +1,506 @@
...
+			NodeTopology:     item.segment.GetLabelSelector(),
+			Capacity:         quantity,
+		}
+		if c.enumerateObjects >= 0 {
+			// Workaround for testing with a fake client: it doesn't
+			// set the name, so we have to make up something ourselves.
+			c.enumerateObjects++
+			capacity.Name = fmt.Sprintf("csisc-test-%d", c.enumerateObjects)
+			capacity.GenerateName = ""
+		}
```
jsafrane, line 432: I don't like testing code in the controller. Use a reactor to generate name in controller_test.go, as suggested in https://github.com/kubernetes/client-go/issues/439 and implemented in https://github.com/bitnami-labs/sealed-secrets/pull/301

pkg/capacity/controller.go
pohly: At the moment, the CSIStorageCapacity object never gets updated in memory because there is no informer, so I have to deal with it here. `RetryOnConflict` looks like a simpler way to do it. Let me think about it this.

pkg/capacity/controller.go
pohly: I didn't like it either, but didn't know how to get name generation working. Thanks for the pointer, I'll add that.

pkg/capacity/controller.go
```diff
@@ -0,0 +1,506 @@
...
+		return err
+	}
+	c.capacitiesLock.Lock()
+	defer c.capacitiesLock.Unlock()
+outer:
+	for _, capacity := range capacities.Items {
+		if !c.isControlledByUs(capacity.OwnerReferences) {
+			continue
+		}
+		for item, capacity2 := range c.capacities {
```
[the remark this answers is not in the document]
pohly: `prepare` starts with `cache.WaitForCacheSync(ctx.Done(), c.topologyInformer.HasSynced, c.scInformer.Informer().HasSynced)`. That ensures that the various `on*` functions are called and `c.capacities` is fully populated and represents the desired set of objects.

I just realized that I must make my tests a bit stricter: I need to verify that existing objects that don't need to be updated really don't get updated.

pkg/capacity/topology/nodes_test.go
```diff
@@ -0,0 +1,535 @@
...
+
+			var objects []runtime.Object
+			objects = append(objects, makeNodes(tc.initialNodes)...)
+			clientSet := fakeclientset.NewSimpleClientset(objects...)
+			nt := fakeNodeTopology(ctx, testDriverName, clientSet)
+
+			added, removed, _ := addTestCallback(nt)
+			go nt.Run(ctx)
+
+			time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
pohly: Me neither, but I wasn't sure how to avoid it because the informers run in goroutines. `HasSynced()` might work, let me try that.

> In addition, tc.update needs to update the informer caches too... Or should the tc include initialSegments to fill the nodeTopology and no update call is then needed?

That's where I currently need the delay: because the update method just uses the client API, the informers need to time to catch up. Updating the informers directly may be a way to avoid that delay.

pkg/capacity/topology/nodes_test.go
```diff
@@ -0,0 +1,535 @@
...
+					name: node2,
+					driverKeys: map[string][]string{
+						driverName: localStorageKeys,
+					},
+					labels: localStorageLabelsNode2,
+				},
+			},
+			expectedSegments: []*Segment{localStorageNode2},
+		},
+		"deep-topology": {
```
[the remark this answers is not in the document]
pohly: okay

pkg/capacity/controller.go
pohly: Done.

pkg/capacity/topology/doc.go
```diff
@@ -0,0 +1,133 @@
...
+	"sort"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Segment represents a topology segment. Entries are always sorted by
+// key and keys are unique. In contrast to a map, segments therefore
+// can be compared efficiently. A nil segment matches no nodes
+// nodes in a cluster, an empty segment all of them.
```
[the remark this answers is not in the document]
pohly: fixed

pkg/features/features.go
pohly: Done.

pkg/capacity/controller.go
```diff
@@ -0,0 +1,506 @@
...
+		func() int {
+			scs, _ := c.scInformer.Lister().List(labels.Everything())
+			return len(scs)
+		}(),
+		len(c.capacities))
+
+	// Retrieve all existing CSIStorageCapacity objects that are owned by us
+	// *before* working on items (because that would then create duplicates).
+	// There's no reconciliation at runtime because changes not made by us
+	// should never happen.
```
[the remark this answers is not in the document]
pohly: I'm undecided whether I should use a CSIStorageCapacity informer or just list them periodically.

If I use an informer, the controller will get all of its own changes sent back to it, which adds a permanent, fixed overhead. The advantage is that the controller can react to undesirable changes quickly.

If I list, probably with a long delay, then the overhead might be lower (depending on how often it polls), but undesirable changes will be detected more slowly.

@jsafrane : What would you prefer?

go.mod
pohly: The code pushed today uses officially released upstream components.

pkg/capacity/controller.go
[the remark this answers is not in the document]
pohly: Done.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -266,6 +277,50 @@ func main() {
 		controllerCapabilities,
 	)
 
+	var capacityController *capacity.Controller
+	if (*capacityFeatures)[capacity.FeatureCentral] {
+		// TODO: use parameters instead?
```
pohly, line 282: @jsafrane: do you have an opinion whether it is worth making POD_NAME and POD_NAMESPACE real parameters? The advantage is that they become more visible. The disadvantage is that the YAML file becomes more complex because the variables are still needed (we need the env/valueFrom/fieldRef and the parameter which uses it).

Also, is it worth adding a parameter for how far the ownership chain should be walked up? It's not needed for StatefulSet where the desired owner is one level up (current code, hard-coded), but it may be useful for a Deployment->ReplicaSet->Pod scenario, because then the owner should be the Deployment object.

The problem is that in order to retrieve arbitrary parent objects, I need to pull in code like sigs.k8s.io/controller-runtime/pkg/client which can get objects of arbitrary types.

pkg/capacity/controller.go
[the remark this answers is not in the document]
pohly: @jsafrane: ping? ^^^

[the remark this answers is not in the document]
pohly: I've convinced myself that an informer is better because:
- a constant overhead is simpler to measure and plan for
- the code for handling the update conflict can be removed

Implemented (but not pushed yet).

pkg/capacity/controller.go
pohly: I added the informer, so this special case could be removed.

cmd/csi-provisioner/csi-provisioner.go
pohly: I decided to stick with just the env variables.

pohly:
@jsafrane: I've pushed a version of the code with documentation and all of the inline TODOs resolved. Can you perhaps take another look?

I intend to add two more things before merging:
- walking up the owner chain
- skipping storage classes with immediate binding

deploy/kubernetes/rbac.yaml
```diff
@@ -53,6 +53,9 @@ rules:
   - apiGroups: ["storage.k8s.io"]
     resources: ["volumeattachments"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["storage.k8s.io"]
```
pohly, line 56: I've been debating with myself whether this should be really enabled or listed here as a comment block (similar to secrets). I decided to enable it because the rbac.yaml then is simpler to use and the risk is lower compared to enabled access to secrets.

deploy/kubernetes/rbac.yaml
```diff
@@ -53,6 +53,9 @@ rules:
...
+    resources: ["csistoragecapacities"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
pohly, line 58: "patch" is currently not needed, but I'd rather have it listed to avoid a major new release when we have to add it later.

pkg/capacity/controller_test.go
```diff
@@ -0,0 +1,1123 @@
...
+			defer mutex.Unlock()
+			uidCounter++
+			s.UID = types.UID(fmt.Sprintf("CSISC-UID-%d", uidCounter))
+		}
+		s.ResourceVersion = csiscRev + "0"
+		return false, nil, nil
+	}
+}
+
+// updateCSIStorageCapacityReactor implements the logic required for the GenerateName and UID fields to work when using
```
jsafrane, line 873: It manages `ResourceVersion`, not UID nor GenerateName.

pkg/capacity/controller_test.go
```diff
@@ -0,0 +1,1123 @@
...
+			Name:            fmt.Sprintf("csisc-%d", capacityCounter),
+			OwnerReferences: owners,
+		},
+		NodeTopology:     in.segment.GetLabelSelector(),
+		StorageClassName: in.storageClassName,
+		Capacity:         in.getCapacity(),
+	}
+}
+
+func makeCapacities(in []testCapacity) (items []runtime.Object) {
```
jsafrane, line 1095: This seems to be unused

cmd/csi-provisioner/csi-provisioner.go
jsafrane: Env. variables are pretty common. For ownership, I can see that using Deployment is far better than using ReplicaSet, however, I don't know how to make it generic and nice. Perhaps a special case for ReplicaSet?

deploy/kubernetes/rbac.yaml
pohly: Done.

pohly: I also made this more restrictive: we only need permissions inside the namespace.

cmd/csi-provisioner/csi-provisioner.go
pohly: I've implemented it using sigs.k8s.io/controller-runtime/pkg/client: that API works with arbitrary object types.

I just pushed that updated code.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -76,6 +81,13 @@ var (
 	kubeAPIQPS   = flag.Float32("kube-api-qps", 5, "QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0.")
 	kubeAPIBurst = flag.Int("kube-api-burst", 10, "Burst to use while communicating with the kubernetes apiserver. Defaults to 10.")
 
+	capacityFeatures = func() *capacity.Features {
```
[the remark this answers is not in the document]
pohly: I wanted to keep all argument definitions at the top and missed that there was already one being defined further below. I'd rather move `feature-gates` up instead of `capacity-features` down into main, but I don't feed strongly about this.

deploy/kubernetes/rbac.yaml
```diff
@@ -53,6 +53,9 @@ rules:
   - apiGroups: ["storage.k8s.io"]
```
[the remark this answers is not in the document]
pohly: Right, I also found this in the meantime.

> IMO a comment would be more than enough here.

A comment explaining the permissions, or commented out permissions?

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	k8s.io/component-base v0.19.0-rc.2
 	k8s.io/csi-translation-lib v0.19.0-rc.2
 	k8s.io/klog v1.0.0
+	k8s.io/klog/v2 v2.2.0
```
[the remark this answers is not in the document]
pohly: I wanted to avoid future changes once we switch, but I can also use just klog for the sake of consistency. It doesn't change much because we already pull klog/v2.

Fixed.

pkg/capacity/controller.go
```diff
@@ -0,0 +1,562 @@
...
+	"context"
+	"fmt"
+	"reflect"
+	"sync"
+	"time"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc"
+	// "github.com/kubernetes-csi/csi-lib-utils/rpc"
+	// v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
pohly: Oh, yes.

BTW, I renamed "controller.go" -> "capacity.go" to have some unique file name in log messages.

pkg/capacity/controller.go
```diff
@@ -0,0 +1,562 @@
...
+	// objects: callbacks handle future updates and iterating
+	// avoids the assumumption that the callback will be invoked
+	// for all objects immediately when adding it.
+	klog.V(3).Info("Checking for existing CSIStorageCapacity objects")
+	handler := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) { c.onCAddOrUpdate(ctx, obj.(*storagev1alpha1.CSIStorageCapacity)) },
+		UpdateFunc: func(_ interface{}, newObj interface{}) {
+			c.onCAddOrUpdate(ctx, newObj.(*storagev1alpha1.CSIStorageCapacity))
+		},
+		DeleteFunc: func(obj interface{}) { c.onCDelete(ctx, obj.(*storagev1alpha1.CSIStorageCapacity)) },
```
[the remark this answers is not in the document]
pohly: Can that ever happen? I've seen the checks elsewhere but didn't add them myself because the callbacks should never get any other kind of object.

pkg/capacity/controller_test.go
```diff
@@ -0,0 +1,1123 @@
...
+					"foo": "1Gi",
+				},
+			},
+			initialCapacities: []testCapacity{
+				{
+					segment:          layer0,
+					storageClassName: "other-sc",
+					quantity:         "1Gi",
+				},
+			},
```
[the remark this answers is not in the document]
pohly: Added `expectedCapacities: []testCapacity{}` wherever it had been omitted.

pkg/capacity/controller_test.go
pohly: Fixed.

pkg/capacity/controller_test.go
pohly: Removed.

pohly:
@jsafrane: I pushed once more to address some of your review comments.

Open:
- content of RBAC file
- argument initialization
- sanity check when casting callback parameters

pkg/owner/owner.go
```diff
@@ -0,0 +1,97 @@
...
+func Lookup(config *rest.Config, namespace, name string, gkv schema.GroupVersionKind, levels int) (*metav1.OwnerReference, error) {
+	c, err := client.New(config, client.Options{})
+	if err != nil {
+		return nil, fmt.Errorf("build client: %v", err)
+	}
+
+	return lookupRecursive(c, namespace, name, gkv, levels)
+}
+
+func lookupRecursive(c client.Client, namespace, name string, gkv schema.GroupVersionKind, levels int) (*metav1.OwnerReference, error) {
```
jsafrane, line 47: Would it make sense to have `apiVersion` and `kind` as parameters instead of `GroupVersionKind`? The recursive loop parses `gvk` from an object via `ParseGroupVersion` only to assemble it back into `APIVersion` / `Kind` in the next iteration.

In reality the loop will be executed twice at max, so I don't have a strong opinion.

pkg/owner/owner.go
```diff
@@ -0,0 +1,97 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Lookup walks up the ownership chain zero or more levels and returns an OwnerReference for the
+// object. The object identified by name, namespace and type is the starting point and is
+// returned when levels is zero. Only APIVersion, Kind, Name, and UID will be set.
+// IsController is always true.
+func Lookup(config *rest.Config, namespace, name string, gkv schema.GroupVersionKind, levels int) (*metav1.OwnerReference, error) {
```
jsafrane, line 38: Please add an unit tests with a Deployment + StatefulSet

pkg/capacity/controller.go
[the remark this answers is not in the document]
jsafrane: At least DeleteFunc used to be called with deletion tombstone - not the latest version of the API object, but the latest known by the informer.

https://github.com/kubernetes/kubernetes/blob/ff3e5e06a79bc69ad3d7ccedd277542b6712514b/pkg/controller/volume/persistentvolume/pv_controller_base.go#L180

deploy/kubernetes/rbac.yaml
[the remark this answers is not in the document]
jsafrane: A comment explaining the permissions. So driver vendors can do qualified judgement and delete those they don't need.

cmd/csi-provisioner/csi-provisioner.go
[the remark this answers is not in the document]
jsafrane: ack

pkg/owner/owner.go
pohly: Even if that adds yet another dependency, specifically on sigs.k8s.io/controller-runtime/pkg/client/fake?

 I wasn't quite sure whether that would be worthwhile, but I can add it.

pohly: It won't cover RBAC rules, though.

pkg/owner/owner.go
pohly: Yes, that makes sense. I picked gkv for the exported API because it is strongly typed instead of three loosely type strings, but here the code is local and strings are better.

pohly:
> I can see an unit test failing, please take a look at it.

I don't get that locally. Perhaps the 5 second timeout was too aggressive? It shouldn't need many CPU cycles, but perhaps the entire process really wasn't getting scheduled for an extended period of time. I've extended that to a minute.

pkg/owner/owner.go
pohly: Done and pushed.

pkg/owner/owner.go
pohly: Done.

pkg/owner/owner.go
jsafrane: Awesome!

pkg/capacity/controller.go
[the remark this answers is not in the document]
pohly: Interesting, I learned something new, thanks. I wonder whether I could have learned that from reading some API description.

I added cast checking. I see that as something that needs to be handled in the glue code rather than the actual callback, but it's now a bit repetitive, so I could be convinced to move casting into the callback...

pkg/capacity/capacity.go
```diff
@@ -0,0 +1,627 @@
...
+		if err != nil {
+			return fmt.Errorf("create CSIStorageCapacity for %+v: %v", item, err)
+		}
+		klog.V(5).Infof("Capacity Controller: created %s for %+v with capacity %v", capacity.Name, item, quantity)
+	} else if capacity.Capacity.Value() == quantity.Value() {
+		klog.V(5).Infof("Capacity Controller: no need to update %s for %+v, same capacity %v", capacity.Name, item, quantity)
+		return nil
+	} else {
+		// Update existing object.
+		capacity.Capacity = quantity
```
msau42, line 518: I  think this capacity object can point to the object that's stored in the informer, since it's populated from the event handlers. We should not modify objects stored in the informer cache. Instead we need to make a copy of object before doing an update.

pkg/capacity/capacity.go
```diff
@@ -0,0 +1,627 @@
...
+		var err error
+		klog.V(5).Infof("Capacity Controller: updating %s for %+v, new capacity %v", capacity.Name, item, quantity)
+		capacity, err = c.client.StorageV1alpha1().CSIStorageCapacities(capacity.Namespace).Update(ctx, capacity, metav1.UpdateOptions{})
```
msau42, line 521: Why update instead of patch? There's less chance of conflicts when doing a patch.

pkg/capacity/topology/nodes.go
```diff
@@ -0,0 +1,328 @@
...
+	nt.sync(ctx)
+	return true
+}
+
+func (nt *nodeTopology) sync(ctx context.Context) {
+	// For all nodes on which the driver is registered, collect the topology key/value pairs
+	// and sort them by key name to make the result deterministic. Skip all segments that have
+	// been seen before.
+	segments := nt.List()
+	removed := map[*Segment]bool{}
```
msau42, line 237: Can you name this something like removalCandidates since everything in here may not be actually removed.

pkg/capacity/topology/nodes.go
```diff
@@ -0,0 +1,328 @@
...
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	coreinformersv1 "k8s.io/client-go/informers/core/v1"
+	storageinformersv1 "k8s.io/client-go/informers/storage/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+)
+
+func NewNodeTopology(
```
msau42, line 38: Future todo but I think parts of this could be used to address https://github.com/kubernetes-csi/external-provisioner/issues/374

deploy/kubernetes/rbac.yaml
```diff
@@ -87,6 +87,19 @@ rules:
 - apiGroups: ["coordination.k8s.io"]
   resources: ["leases"]
   verbs: ["get", "watch", "list", "delete", "update", "create"]
+- apiGroups: ["storage.k8s.io"]
```
msau42, line 90: Add a note that this is only needed if you're using the storage capacity feature

msau42:
cc @verult

pkg/capacity/topology/nodes.go
```diff
@@ -0,0 +1,328 @@
...
+				// it in CSINode, but we haven't seen the corresponding
+				// node update yet as the label is not set. Ignore the node
+				// for now, we'll sync up when we get the node update.
+				continue node
+			}
+			newSegment = append(newSegment, SegmentEntry{key, value})
+		}
+
+		// Add it only if new, otherwise look at the next node.
+		for _, segment := range segments {
```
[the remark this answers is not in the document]
pohly: The problem is that a segment is a slice, and slices cannot be used as key in a map.

There are two approaches that would work:
- A tree with topology key/value pairs on the edges and the segment with those pairs in the leaf.
- Serialize the slice in a string, then use that string as key in a map.

I  was not sure which approach would work better in practice, so I kept it simple for now.

pkg/owner/owner.go
```diff
@@ -0,0 +1,97 @@
...
+			APIVersion: apiVersion,
+			Kind:       kind,
+			Name:       name,
+			UID:        u.GetUID(),
+			Controller: &isTrue,
+		}, nil
+	}
+	owners := u.GetOwnerReferences()
+	for _, owner := range owners {
+		if owner.Controller != nil && *owner.Controller {
```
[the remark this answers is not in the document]
pohly: Each object can have multiple owners, but not all of them are controlling the object. The rationale here is that we want to walk up the chain back to whatever created the pod and ignore all other owner references.

https://github.com/kubernetes/apimachinery/blob/2c6e837afb200aae8d14b68b0fa258ccecba58a6/pkg/apis/meta/v1/controller_ref.go#L42-L51 also ignores the non-controlling owners.

pkg/owner/owner.go
```diff
@@ -0,0 +1,97 @@
...
+			gv, err := schema.ParseGroupVersion(owner.APIVersion)
+			if err != nil {
+				return nil, fmt.Errorf("parse OwnerReference.APIVersion: %v", err)
+			}
+			// With this special case here we avoid one lookup and thus the need for
+			// more RBAC permissions.
```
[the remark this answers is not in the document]
pohly: Added together with an example.

pohly:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-provisioner/450/pull-kubernetes-csi-external-provisioner-1-18-on-kubernetes-1-18/1294189594291474435 failed because:
- The job wants to bring up a cluster and tries to build Kubernetes master for that.
- The Go version for master is determined to be 1.15.0.
- The download URL for .0 Go versions is without that digit, i.e. https://dl.google.com/go/go1.15.0.linux-amd64.tar.gz should be https://dl.google.com/go/go1.15.linux-amd64.tar.gz.

That download failure can be fixed. But what puzzles me is that the job tries to build Kubernetes from source. It shouldn't be doing that.

pohly:
> That download failure can be fixed. But what puzzles me is that the job tries to build Kubernetes from source. It shouldn't 
be doing that.

Gotcha! We don't configure which Kubernetes e2e to use for 1.18, so it falls back to building master.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -266,15 +281,72 @@ func main() {
...
 	)
 
+	var capacityController *capacity.Controller
+	if (*capacityFeatures)[capacity.FeatureCentral] {
+		podName := os.Getenv("POD_NAME")
+		namespace := os.Getenv("POD_NAMESPACE")
+		if podName == "" || namespace == "" {
+			klog.Fatalf("need POD_NAMESPACE/POD_NAME env variables, have only POD_NAMESPACE=%q and POD_NAME=%q", namespace, podName)
+		}
+		controller, err := owner.Lookup(config, namespace, podName,
```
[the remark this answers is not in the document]
msau42: While encouraged, we have not made it a requirement that a csi driver must run as a Kubernetes pod, and plenty of deployers have run CSI drivers and the sidecars as binaries, docker containers, and static pods (which do not show up in the api server).

Why must we have an owner? I see it as an optimization for cleaning up objects when you uninstall a driver. It doesn't really help the "run 2 of the same driver" since the scheduler can't distinguish.

We can look at adding a mode to support this in the future, but I would consider it a beta blocker. For now, let's make sure to document this requirement in the README.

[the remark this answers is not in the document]
pohly: I've added a paragraph about this and a tracking issue: https://github.com/kubernetes-csi/external-provisioner/issues/465

README.md
```diff
@@ -61,14 +62,26 @@ Note that the external-provisioner does not scale with more replicas. Only one e
...
 
 * `--extra-create-metadata`: Enables the injection of extra PVC and PV metadata as parameters when calling `CreateVolume` on the driver (keys: "csi.storage.k8s.io/pvc/name", "csi.storage.k8s.io/pvc/namespace", "csi.storage.k8s.io/pv/name")
 
+##### Storage capacity arguments
+
+See the [storage capacity section](#capacity-support) below for details.
+
+* `--capacity-threads <num>`: Number of simultaneously running threads, handling CSIStorageCapacity objects. Defaults to `1`.
+
+* `--capacity-poll-period <interval>`: How long the external-provisioner waits before checking for storage capacity changes. Defaults to `1m`.
```
msau42, line 79: Sorry I meant "interval" in the field name. For example `-retry-interval-max`, except in this case it's not exponential.

pkg/capacity/features.go
```diff
@@ -0,0 +1,83 @@
...
+// controller.
+type Feature string
+
+// Features are disabled by default.
+type Features map[Feature]bool
+
+const (
+	// FeatureCentral enables the mode where there is only one
+	// external-provisioner actively running in the cluster which
+	// talks to the CSI driver's controller.
```
msau42, line 36: add "service" to be more explicit we're talking about the csi service

pkg/capacity/capacity.go
```diff
@@ -0,0 +1,627 @@
...
+
+	// The caches are fully populated now, but the event handlers
+	// may or may not have been invoked yet. To be sure that we
+	// have all data, we need to list all resources. Here we list
+	// topology segments, onTopologyChanges lists the classes.
+	c.onTopologyChanges(c.topologyInformer.List(), nil)
+
+	klog.V(3).Infof("Initial number of topology segments %d, storage classes %d, potential CSIStorageCapacity objects %d",
+		len(c.topologyInformer.List()),
+		func() int {
```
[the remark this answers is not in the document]
msau42: Can we store this somewhere else? It would be nice to not have to do a List just for a log message.

README.md
```diff
@@ -61,14 +61,22 @@ Note that the external-provisioner does not scale with more replicas. Only one e
...
 
 * `--metrics-path`: The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.
 
 * `--extra-create-metadata`: Enables the injection of extra PVC and PV metadata as parameters when calling `CreateVolume` on the driver (keys: "csi.storage.k8s.io/pvc/name", "csi.storage.k8s.io/pvc/namespace", "csi.storage.k8s.io/pv/name")
 
+* `--capacity-threads <num>`: Number of simultaneously running threads, handling CSIStorageCapacity objects. Defaults to `1`.
+
+* `--capacity-poll-period <duration>`: How long the external-provisioner waits before checking for storage capacity changes. Defaults to `1m`.
+
+* `--enable-capacity <enumeration>`: Enables producing CSIStorageCapacity objects with capacity information from the driver's GetCapacity call. Can be given more than once and/or with comma-separated values. Currently supported: --enable-capacity=central,immediate-binding.
```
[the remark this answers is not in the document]
pohly: `--capacity-controller-deployment-mode` works for me. Default will be `none`.

I shortened the other option to `--capacity-for-immediate-binding` to keep the grouped together and shorter.

pkg/capacity/features.go
pohly: Done.

pkg/capacity/topology/nodes.go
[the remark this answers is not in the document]
pohly: Added a tracking issue for this, too: https://github.com/kubernetes-csi/external-provisioner/issues/466

pkg/capacity/capacity.go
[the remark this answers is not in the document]
pohly: It's in a function because I didn't want to introduce variables that are only used inside the log call. I've changed it.

Both `c.topologyInformer` and `c.scInformer` implement `List` as a local operation and this log message is only printed once, therefore I am not worried about the performance.

pkg/capacity/capacity.go
```diff
@@ -0,0 +1,627 @@
...
+	// Ensure that we have an entry for it...
+	capacity := c.capacities[item]
+	c.capacities[item] = capacity
+	// ... and then tell our workers to update
+	// or create that capacity object.
+	klog.V(5).Infof("Capacity Controller: enqueuing %+v", item)
+	c.queue.Add(item)
+}
+
+// addWorkItem ensures that the item gets removed from c.capacities. It
```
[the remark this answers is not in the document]
pohly: Fixed.

pkg/capacity/capacity.go
```diff
@@ -0,0 +1,627 @@
...
+			// longer needed.
+			return nil
+		}
+		return fmt.Errorf("retrieve storage class for %+v: %v", item, err)
+	}
+
+	req := &csi.GetCapacityRequest{
+		Parameters: sc.Parameters,
+		// The assumption is that the capacity is independent of the
+		// capabilities. The standard makes it mandatory to pass something,
```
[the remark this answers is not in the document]
pohly: Not sure. Perhaps it does make sense somewhere and it's just a simplifying assumption that is valid in Kubernetes, so let's gather feedback before proposing changes.

pkg/capacity/capacity.go
pohly: Okay, done.

pkg/capacity/capacity.go
pohly: There should be exactly a single writer for each object, so conflicts should never occur.

Performance is a more relevant concern. I'm not sure whether patching is always going to be faster than updating, in particular when the object is small. As with all performance optimizations, we should have a benchmark first -> https://github.com/kubernetes-csi/external-provisioner/issues/467

README.md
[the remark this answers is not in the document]
msau42: What does `none` do?

pkg/capacity/capacity.go
```diff
@@ -0,0 +1,627 @@
...
+	}
+	c.cInformer.Informer().AddEventHandler(handler)
+	capacities, err := c.cInformer.Lister().List(labels.Everything())
+	if err != nil {
+		// This shouldn't happen.
+		utilruntime.HandleError(err)
+		return
+	}
+	for _, capacity := range capacities {
+		c.onCAddOrUpdate(ctx, capacity)
```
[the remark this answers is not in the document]
msau42: Still outstanding. It would be nice to have an explanation why it's important we invoke all the event handlers initially.

pkg/capacity/capacity.go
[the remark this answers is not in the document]
msau42: Picking arbitrary values could get us in trouble if a plugin actually tries to validate any of them like they're supposed to. For now we're basically telling plugins "don't validate the capability for this call if you want it to work in Kubernetes". I think it would be a good to open an issue in the csi spec repo saying that when we tried to implement this, we could not find a meaningful use for this.

msau42:
/approve
Thank you!

Will let @jsafrane do another pass.

I just have one nit on adding a comment. And can you update the release note to point to the readme with more information on how to use the feature?

README.md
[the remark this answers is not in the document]
pohly: Disable CSIStorageCapacity support. I can also use the empty string instead.

pkg/capacity/capacity.go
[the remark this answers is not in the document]
pohly: Sorry, missed this one. There was already a comment further up (" Now that we know what we need, we can check what we have.") which I extended:

> This is important to avoid duplicates, i.e. creating a new
> object with the same topology and storage class as an
> existing one instead of updating that one.

The reason for explicitly iterating over the list was also explained there ("We do that both via callbacks *and* by iterating over all objects: callbacks handle future updates and iterating avoids the assumption that the callback will be invoked for all objects immediately when adding it.").

Without iterating over the list I had seen random test failures because the event handlers where getting called in parallel to and sometimes after the initial processing. This is consistent with the documentation of informers, which only states that the local state inside the cache is up-to-date after WaitForCacheSync.

pkg/capacity/capacity.go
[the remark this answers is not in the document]
pohly: Plugins can validate capabilities, because we are passing valid ones. What we are saying is "we expect capacity to be independent of this particular parameter". I created an issue for this: https://github.com/container-storage-interface/spec/issues/444

README.md
```diff
@@ -61,14 +62,28 @@ Note that the external-provisioner does not scale with more replicas. Only one e
...
 
+##### Storage capacity arguments
+
+See the [storage capacity section](#capacity-support) below for details.
+
+* `--capacity-threads <num>`: Number of simultaneously running threads, handling CSIStorageCapacity objects. Defaults to `1`.
+
+* `--capacity-poll-interval <interval>`: How long the external-provisioner waits before checking for storage capacity changes. Defaults to `1m`.
+
+* `--capacity-controller-deployment-mode central|none`: Enables producing CSIStorageCapacity objects with capacity information from the driver's GetCapacity call. 'central' is currently the only supported mode. Use it when there is just one active provisioner in the cluster. Defaults to `none`.
```
msau42, line 81: Sorry I think this merged before I got replied. I think it's not obvious that "none" means disabled. Could you clarify that in the doc, or just have unset as disabled?

Maybe also put this argument first since it basically controls the remaining args.

pohly: Fixed in https://github.com/kubernetes-csi/external-provisioner/pull/468

PR #457, opened 2020-07-25
Update README with qps and burst argument description

RaunakShah:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**:
https://github.com/kubernetes-csi/external-provisioner/pull/447 did not update the README with information about how to use these parameters. Adding a one liner for `--kube-api-qps` and `--kube-api-burst` as part of this PR in the README.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Update README with qps and burst argument description
```

PR #461, opened 2020-08-10
Add link to QPS and Burst documentation

RaunakShah:
**What type of PR is this?**

> /kind documentation

Follow up to https://github.com/kubernetes-csi/external-provisioner/pull/457, addressing @pohly's comment - https://github.com/kubernetes-csi/external-provisioner/pull/457#discussion_r464445846

Couldn't find a decent document that describes QPS and Burst explicitly, so just linking to the code documentation. 

Fixes #

```release-note
Adding documentation link to describe QPS and Burst parameters. 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,9 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--worker-threads <num>`: Number of simultaneously running `ControllerCreateVolume` and `ControllerDeleteVolume` operations. Default value is `100`.
 
-* `--kube-api-qps <num>`: QPS for clients that communicate with the kubernetes apiserver. Defaults to `5.0`.
+* `--kube-api-qps <num>`: The number of requests per second sent by a Kubernetes client to the Kubernetes API server. Defaults to `5.0`.
 
-* `--kube-api-burst <num>`: Burst for clients that communicate with the kubernetes apiserver. Defaults to `10`.
+* `--kube-api-burst <num>`: The number of requests to the Kubernetes API server, exceeding the QPS, that can be sent at any given time. Defaults to `10`.
 
 * `--cloning-protection-threads <num>`: Number of simultaneously running threads, handling cloning finalizer removal. Defaults to `1`.
 


README.md
```diff
@@ -57,9 +57,9 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--worker-threads <num>`: Number of simultaneously running `ControllerCreateVolume` and `ControllerDeleteVolume` operations. Default value is `100`.
 
-* `--kube-api-qps <num>`: QPS for clients that communicate with the kubernetes apiserver. Defaults to `5.0`.
+* `--kube-api-qps <num>`: [QPS](https://github.com/kubernetes/client-go/blob/master/rest/config.go#L115) for clients that communicate with the Kubernetes apiserver. Defaults to `5.0`.
```
xing-yang, line 60: Can you point to code in a specific release?  Otherwise the line of code in the master branch could change any time.

RaunakShah: good point

README.md
```diff
@@ -57,9 +57,9 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--worker-threads <num>`: Number of simultaneously running `ControllerCreateVolume` and `ControllerDeleteVolume` operations. Default value is `100`.
 
-* `--kube-api-qps <num>`: QPS for clients that communicate with the kubernetes apiserver. Defaults to `5.0`.
+* `--kube-api-qps <num>`: [QPS](https://github.com/kubernetes/client-go/blob/release-1.19/rest/config.go#L115) for clients that communicate with the Kubernetes apiserver. Defaults to `5.0`.
```
pohly, line 60: `release-1.19` is still a branch. Better use a tag like `v0.19.0-rc.4`. Highlighting also becomes nicer when using a range: https://github.com/kubernetes/client-go/blob/5521967004d84d9e6f89df86dfeb5977f993bcbd/rest/config.go#L115-L117

But that's mute because that documentation still doesn't explain what QPS and burst are. Explaining "QPS" with "QPS indicates the maximum QPS" doesn't help when it's unknown what "QPS" is in the first place :cry: 

Can you ask around whether there's any better documentation of these concepts?

msau42: If we can't find better documentation, can you just add a high level description?

RaunakShah: How about this:

```
QPS - "The number of requests per second processed by a client to the Kubernetes API server"
Burst - "The number of requests to the Kubernetes API server, exceeding the QPS, that can be processed at any given time"
```

and i'll add a link to https://github.com/kubernetes/client-go/blob/master/util/flowcontrol/throttle.go#L48

msau42: Sounds good to me. Just a nit, instead of "a client", what about "the Kubernetes client"

pohly: I've been looking into this a bit more recently myself. I think a more accurate verb is "sent" instead of "processed", because that's really where the rate check happens (https://github.com/kubernetes/client-go/blob/5521967004d84d9e6f89df86dfeb5977f993bcbd/rest/request.go#L847-L852): before sending the request to the server, the client voluntarily slows itself down.

RaunakShah: @msau42 @pohly PTAL!

PR #469, opened 2020-08-21
Set READONLY flag in CSI PV based on PVC accessmode

humblec:
Refer: https://github.com/kubernetes/kubernetes/issues/70505

```release-note
The provisioner sidecar now has an argument called `controller-publish-readonly`
which set the value of CSI PV spec `readonly` field value based on the PVC access mode. 

If this flag has been set to `true` and pvc access mode carry only `ROX` access mode,
the controller automatically set PersistentVolume.spec.CSIPersistentVolumeSource.readOnly
field  to `true`.
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -103,6 +103,7 @@ var (
 	nodeDeploymentImmediateBinding = flag.Bool("node-deployment-immediate-binding", true, "Determines whether immediate binding is supported when deployed on each node.")
 	nodeDeploymentBaseDelay        = flag.Duration("node-deployment-base-delay", 20*time.Second, "Determines how long the external-provisioner sleeps initially before trying to own a PVC with immediate binding.")
 	nodeDeploymentMaxDelay         = flag.Duration("node-deployment-max-delay", 60*time.Second, "Determines how long the external-provisioner sleeps at most before trying to own a PVC with immediate binding.")
+	controllerPublishReadOnly      = flag.Bool("controller-publish-readonly", false, "This option enables PV to be marked as readonly at controller publish volume call if PVC accessmode has been set to ROX.")
 
 	featureGates        map[string]bool
 	provisionController *controller.ProvisionController
@@ -387,6 +388,7 @@ func main() {
 		*extraCreateMetadata,
 		*defaultFSType,
 		nodeDeployment,
+		*controllerPublishReadOnly,
 	)
 
 	var capacityController *capacity.Controller
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -109,11 +109,6 @@ var (
 	version             = "unknown"
 )
 
-type leaderElection interface {
-	Run() error
-	WithNamespace(namespace string)
-}
-
 func main() {
 	var config *rest.Config
 	var err error
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -259,6 +259,7 @@ type csiProvisioner struct {
 	extraCreateMetadata                   bool
 	eventRecorder                         record.EventRecorder
 	nodeDeployment                        *internalNodeDeployment
+	controllerPublishReadOnly             bool
 }
 
 var _ controller.Provisioner = &csiProvisioner{}
@@ -337,6 +338,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 	extraCreateMetadata bool,
 	defaultFSType string,
 	nodeDeployment *NodeDeployment,
+	controllerPublishReadOnly bool,
 ) controller.Provisioner {
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartLogging(klog.Infof)
@@ -369,6 +371,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 		vaLister:                              vaLister,
 		extraCreateMetadata:                   extraCreateMetadata,
 		eventRecorder:                         eventRecorder,
+		controllerPublishReadOnly:             controllerPublishReadOnly,
 	}
 	if nodeDeployment != nil {
 		provisioner.nodeDeployment = &internalNodeDeployment{
@@ -810,9 +813,17 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 			return nil, controller.ProvisioningInBackground, sourceErr
 		}
 	}
+	pvReadOnly := false
+	volCaps := req.GetVolumeCapabilities()
+	// if the request only has one accessmode and if its ROX, set readonly to true
+	// TODO: check for the driver capability of MULTI_NODE_READER_ONLY capability from the CSI driver
+	if len(volCaps) == 1 && volCaps[0].GetAccessMode().GetMode() == csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY && p.controllerPublishReadOnly {
+		pvReadOnly = true
+	}
 
 	result.csiPVSource.VolumeHandle = p.volumeIdToHandle(rep.Volume.VolumeId)
 	result.csiPVSource.VolumeAttributes = volumeAttributes
+	result.csiPVSource.ReadOnly = pvReadOnly
 	pv := &v1.PersistentVolume{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: pvName,
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -417,7 +417,7 @@ func TestCreateDriverReturnsInvalidCapacityDuringProvision(t *testing.T) {
 
 	pluginCaps, controllerCaps := provisionCapabilities()
 	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test",
-		5, csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, false, defaultfsType, nil)
+		5, csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, false, defaultfsType, nil, true)
 
 	// Requested PVC with requestedBytes storage
 	deletePolicy := v1.PersistentVolumeReclaimDelete
@@ -835,6 +835,7 @@ type provisioningTestcase struct {
 	expectSelectedNode            string // a specific selected-node of the PVC in the apiserver after the test, same as before if empty
 	expectNoProvision             bool   // if true, then ShouldProvision should return false
 	supportsSingleNodeMultiWriter bool   // if true, then provision with single node multi writer capabilities
+	controllerPublishReadOnly     bool
 }
 
 type provisioningFSTypeTestcase struct {
@@ -1216,7 +1217,7 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 			},
 			expectState: controller.ProvisioningFinished,
 		},
-		"provision with access mode multi node multi readonly": {
+		"provision with access mode multi node multi readonly with sidecar arg false": {
 			volOpts: controller.ProvisionOptions{
 				StorageClass: &storagev1.StorageClass{
 					ReclaimPolicy: &deletePolicy,
@@ -1239,6 +1240,7 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 					},
 				},
 			},
+			controllerPublishReadOnly: false,
 			expectedPVSpec: &pvSpec{
 				Name:          "test-testi",
 				ReclaimPolicy: v1.PersistentVolumeReclaimDelete,
@@ -1250,6 +1252,61 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 					Driver:       "test-driver",
 					VolumeHandle: "test-volume-id",
 					FSType:       "ext4",
+					ReadOnly:     false,
+					VolumeAttributes: map[string]string{
+						"storage.kubernetes.io/csiProvisionerIdentity": "test-provisioner",
+					},
+				},
+			},
+			expectCreateVolDo: func(t *testing.T, ctx context.Context, req *csi.CreateVolumeRequest) {
+				if len(req.GetVolumeCapabilities()) != 1 {
+					t.Errorf("Incorrect length in volume capabilities")
+				}
+				if req.GetVolumeCapabilities()[0].GetAccessMode() == nil {
+					t.Errorf("Expected access mode to be set")
+				}
+				if req.GetVolumeCapabilities()[0].GetAccessMode().GetMode() != csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY {
+					t.Errorf("Expected multi_node_reader_only")
+				}
+			},
+			expectState: controller.ProvisioningFinished,
+		},
+		"provision with access mode multi node multi readonly with sidecar arg true": {
+			volOpts: controller.ProvisionOptions{
+				StorageClass: &storagev1.StorageClass{
+					ReclaimPolicy: &deletePolicy,
+					Parameters:    map[string]string{},
+				},
+				PVName: "test-name",
+				PVC: &v1.PersistentVolumeClaim{
+					ObjectMeta: metav1.ObjectMeta{
+						UID:         "testid",
+						Annotations: driverNameAnnotation,
+					},
+					Spec: v1.PersistentVolumeClaimSpec{
+						Selector: nil,
+						Resources: v1.ResourceRequirements{
+							Requests: v1.ResourceList{
+								v1.ResourceName(v1.ResourceStorage): resource.MustParse(strconv.FormatInt(requestedBytes, 10)),
+							},
+						},
+						AccessModes: []v1.PersistentVolumeAccessMode{v1.ReadOnlyMany},
+					},
+				},
+			},
+			controllerPublishReadOnly: true,
+			expectedPVSpec: &pvSpec{
+				Name:          "test-testi",
+				ReclaimPolicy: v1.PersistentVolumeReclaimDelete,
+				AccessModes:   []v1.PersistentVolumeAccessMode{v1.ReadOnlyMany},
+				Capacity: v1.ResourceList{
+					v1.ResourceName(v1.ResourceStorage): bytesToQuantity(requestedBytes),
+				},
+				CSIPVS: &v1.CSIPersistentVolumeSource{
+					Driver:       "test-driver",
+					VolumeHandle: "test-volume-id",
+					FSType:       "ext4",
+					ReadOnly:     true,
 					VolumeAttributes: map[string]string{
 						"storage.kubernetes.io/csiProvisionerIdentity": "test-provisioner",
 					},
@@ -2244,7 +2301,7 @@ func runFSTypeProvisionTest(t *testing.T, k string, tc provisioningFSTypeTestcas
 		myDefaultfsType = ""
 	}
 	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5, csiConn.conn,
-		nil, provisionDriverName, pluginCaps, controllerCaps, supportsMigrationFromInTreePluginName, false, true, csitrans.New(), nil, nil, nil, nil, nil, false, myDefaultfsType, nil)
+		nil, provisionDriverName, pluginCaps, controllerCaps, supportsMigrationFromInTreePluginName, false, true, csitrans.New(), nil, nil, nil, nil, nil, false, myDefaultfsType, nil, false)
 	out := &csi.CreateVolumeResponse{
 		Volume: &csi.Volume{
 			CapacityBytes: requestedBytes,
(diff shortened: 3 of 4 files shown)


pkg/controller/controller.go
```diff
@@ -689,6 +689,15 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
...
 
+	pvReadOnly := false
+	volCaps := req.GetVolumeCapabilities()
+	// if the request only has one accessmode and if its ROX, set readonly to true
+	if len(volCaps) == 1 {
+		if volCaps[0].GetAccessMode().GetMode() == csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY || volCaps[0].GetAccessMode().GetMode() == csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY {
+			pvReadOnly = true
+		}
+	}
+
```
gnufied, line 700: I am not sure if presence of ROX accessMode means volume should be controller published as readonly.  To me, these they can be orthogonal concerns. cc @bswartz

humblec: true.. but regardless of how controller publish the PV source could still set it RO .

gnufied: yes but interpreting ROX as readonly in PV source should be done with caution. It could make the PV unmountable in case journal needs to be replayed or something else.

bswartz: Yes I was going to argue the same. Volumes should not be controller-published as read-only unless the PV is marked read-only. There are too many good reasons for publishing them writable to the node and then having the node re-mount them as read-only to the pod.

Journal-replays are one use case. Pending expand operations are another.

I interpret PVs with the readOnly: true flag to be PERMANENTLY read-only, which you almost never want in practice. We might want to make it easier to cause that PV flag to get set to true in some cases, but I think the decision for ControllerPublishVolume should look at that flag only and not look at access modes.

humblec: I agree the FS operations may need it RW while controller publish. But SP driver is free to do the operation or default to RW while controller publishing or node staging. Isnt it ? 

>I interpret PVs with the readOnly: true flag to be PERMANENTLY read-only, which you almost never want in practice. 

I could think of a scenario where user provision a new volume as "RO" with dataSource set to `PVC` or `Snapshot`. Isnt it a valid use case where user want a PV to be READONLY? 


@saad-ali @msau42 @jsafrane @xing-yang inviting your thoughts .

bswartz: I said _almost_ never, and you put your finger on one of the valid cases.

And I apologize for not reading the actual code when I last responded. I was just responding to the comment. I see now that this access mode check is deciding what to set for the PV's readOnly value, not the publish check. I think this is a relatively safe change to make.

The benefit of this change, as I see it:
- Making the PV read only means the nodes won't be able to write to the volume, which prevents data corruption in the case of a rogue node or a container break-out attack.

The downsides of this change are:
- Filesystem volumes created in this way won't be resizeable, ever.
- In the case that the snapshot from which the filesystem volume was cloned was taken with an unclean journal, the resulting volume will be unmountable and the user won't have any recourse.

humblec: >The downsides of this change are:
    Filesystem volumes created in this way won't be resizeable, ever.

If the FS volume is created with RO mode, I dont see a reason to expand the volume at all :).. That said, its actually an `empty volume got provisioned` + `no data` can be written any way. so I dont see a reason to expand it. more or less, we can block the RO volume resize as an additional measure too.


 >In the case that the snapshot from which the filesystem volume was cloned was taken with an unclean journal, the resulting volume will be unmountable and the user won't have any recourse.
>

If the volume got provisioned from a datasource with RO mode,  I am wondering what could cause an unclean journal? if that was the case, SP driver could have declared failure in very first attempt of provisioning itself or I would rather put it as a clone/restore failure from storage backend.  Also, even in this case, SP driver still have a provision of respecting the RO flag in the PV or discarding it while staging or publishing it.

humblec:
@msau42 @xing-yang @saad-ali @jsafrane ptal :+1:

xing-yang:
@humblec In the proposal described in kubernetes/kubernetes#70505, there is also the following.  I don't see that is addressed by this PR?
```
Add new csiControllerPublishVolumeReadOnly StorageClass parameter to CSI external-provisioner
```

humblec:
> @humblec In the proposal described in [kubernetes/kubernetes#70505](https://github.com/kubernetes/kubernetes/issues/70505), there is also the following. I don't see that is addressed by this PR?
> 
> ```
> Add new csiControllerPublishVolumeReadOnly StorageClass parameter to CSI external-provisioner
> ```

@xing-yang I was in an impression that, the SC param is an optional one. Or is it mandatory that, we will expose the SC param too ?

xing-yang:
>> @humblec In the proposal described in kubernetes/kubernetes#70505, there is also the following. I don't see that is addressed by this PR?

>>> Add new csiControllerPublishVolumeReadOnly StorageClass parameter to CSI external-provisioner

> @xing-yang I was in an impression that, the SC param is an optional one. Or is it mandatory that, we will expose the SC param too ?

@humblec based on  kubernetes/kubernetes#70505, "Add new csiControllerPublishVolumeReadOnly StorageClass parameter to CSI external-provisioner" should be addressed as well.  

@saad-ali can you chime in as you wrote the original proposal.

humblec:
> > > @humblec In the proposal described in [kubernetes/kubernetes#70505](https://github.com/kubernetes/kubernetes/issues/70505), there is also the following. I don't see that is addressed by this PR?
> 
> > > > Add new csiControllerPublishVolumeReadOnly StorageClass parameter to CSI external-provisioner
> 
> > @xing-yang I was in an impression that, the SC param is an optional one. Or is it mandatory that, we will expose the SC param too ?
> 
> @humblec based on [kubernetes/kubernetes#70505](https://github.com/kubernetes/kubernetes/issues/70505), "Add new csiControllerPublishVolumeReadOnly StorageClass parameter to CSI external-provisioner" should be addressed as well.
> 

Somehow I was in an impression that its  either 1 or 2 in that proposal  :), so I was trying clarify. I will wait for @saad-ali thought to double confirm.  

> @saad-ali can you chime in as you wrote the original proposal.

humblec:
While we wait confirmation, I have added an extra commit for SC param. Please take a look at it and share your thoughts @xing-yang

saad-ali:
This looks good to me. It uses the PVC access mode to auto set the value of `PV.spec.CSIPersistentVolumeSource.readOnly` which is always left `false` today. And it provides a StorageClass parameter to override the automation if desired.

/lgtm
/approve

Adding a hold to give others an opportunity to review.

/hold

pkg/controller/controller.go
```diff
@@ -689,6 +700,20 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 		}
 	}
 
+	pvReadOnly := false
+	volCaps := req.GetVolumeCapabilities()
+	// if the request only has one accessmode and if its ROX, set readonly to true
+	if len(volCaps) == 1 {
+		if volCaps[0].GetAccessMode().GetMode() == csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY || volCaps[0].GetAccessMode().GetMode() == csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY {
```
xing-yang, line 707: csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY is not really mapped to ReadOnlyMany by the external-provisioner, right?  So if you are checking it here, I'd suggest that you add a note to clarify that.

pkg/controller/controller.go
```diff
@@ -689,6 +700,20 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
...
+			if (scReadOnlyKeyFound == 1) && scReadOnlyEnabled {
```
msau42, line 708: Can we just check scReadOnlyEnabled?

Also, is this something that we think should be enabled on a per storageclass basis, or on a driver basis? If driver, then a flag to external provisioner would be better.

pkg/controller/controller.go
msau42: Do we ever set SINGLE_NODE_READER_ONLY?

pkg/controller/controller_test.go
```diff
@@ -1001,6 +1031,100 @@ func TestFSTypeProvision(t *testing.T) {
...
+					ReadOnly:     true,
+					VolumeAttributes: map[string]string{
+						"storage.kubernetes.io/csiProvisionerIdentity": "test-provisioner",
+					},
+				},
+				AccessModes: []v1.PersistentVolumeAccessMode{v1.ReadOnlyMany},
+			},
+			expectState: controller.ProvisioningFinished,
+		},
+		"readonly PVC with SC param key set to true": {
```
msau42, line 1096: can you add one more test case for it set to false?

pkg/controller/controller.go
xing-yang: Does not look like it is set anywhere: https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L407

pkg/controller/controller.go
bswartz: > If the volume got provisioned from a datasource with RO mode, I am wondering what could cause an unclean journal? if that was the case, SP driver could have declared failure in very first attempt of provisioning itself or I would rather put it as a clone/restore failure from storage backend. Also, even in this case, SP driver still have a provision of respecting the RO flag in the PV or discarding it while staging or publishing it.

This is the sequence of events that could lead to trouble:

1. Create an writable volume, with an ext4 (or other journaling) filesystem
2. Pod attaches to the volume, and begins writing data
3. Node containing that pod crashes, loses power, or otherwise dies suddenly
4. Take a snapshot of the volume before it is attached to another pod
5. Create a new volume from the snapshot, and mark it ROX
6. Due to this PR, the new volume is marked readonly=true
6. Try to attach a pod to the new volume, and because the volume is readonly, it's controller-published as readonly
7. When the pod tries to come up, the mount fails because the kernel is unable to replay the journal on a readonly device

It's likely that the node crash doesn't even need to happen. An ordinary snapshot taken at the wrong moment could result in having a snapshot of a volume with a journal that needs replaying to successfully mount the volume.

To reproduce this, you'd need to be taking snapshot of a volume undergoing relatively heavy write load, and then trying to mount a readonly clone of that snapshot. Multiple attempts might be needed because it will depend on exactly when the snapshot occurs relative to the kernel's FS flushes to disk.

pohly: > 8. the pod tries to come up, the mount fails because the kernel is unable to replay the journal on a readonly device

I'd like to question this part. I expect filesystems to be prepared for this case. Consider a normal device that degrades such that writes start to fail. The normal reaction of the kernel is to remount the filesystem as read-only. Being able to mount read-only after a system reboot to extract more data seems like a sensible feature to support - much better than refusing to mount completely.

Whether data and files are consistent is another question, but it's not different from protecting against a sudden power loss.

bswartz: First of all, I want to be clear that I'm not trying to block this issue. I think it's a good change, and I'd like to see it merge.

I was just trying to answer the question of what potential downsides this change could have, because it does introduce this new corner case with journaled filesystems. We shouldn't be surprised to see support cases from people with problems mounting readonly volumes that were cloned from in-use filesystems or snapshots taken of in-use filesystems, if journaling is in use.

There are some workarounds we can either recommend to users, or possibly even build into our CSI drivers to automatically overcome this class of problems. The most obvious workaround is to create a volume that's writable and use that instead. Or, create a writable volume, attach that to a pod that does nothing (to let the journal replay), then clone that volume after the pod has terminated cleanly (so the journal is empty).

The specific kernel error I see when I reproduce this problem with ext4 is:
[587843.214656] EXT4-fs (nbd2): INFO: recovery required on readonly filesystem
[587843.214657] EXT4-fs (nbd2): write access unavailable, cannot proceed (try mounting with noload)

Possibly CSI drivers could use the -o noload option with ext4 filesystems when mounting readonly to avoid the problem.

When using XFS, the kernel errors look like this:
[588571.823845] XFS (nbd1): Mounting V5 Filesystem
[588571.881098] XFS (nbd1): recovery required on read-only device.
[588571.881100] XFS (nbd1): write access unavailable, cannot proceed.
[588571.881101] XFS (nbd1): log mount/recovery failed: error -30
[588571.881123] XFS (nbd1): log mount failed

Some searching turns up the -o norecovery options for XFS which might also work.

In summary, I like this change but I'd urge that CSI driver authors consider special handling for mounting of volumes published read-only at the controller level. And if users do get stuck with unmountable volumes, be ready to suggest the workaround of using a writable volume.

humblec: @bswartz @pohly thanks for the inputs.. First of all good discussion about the concerned area. 

>
Create an writable volume, with an ext4 (or other journaling) filesystem
Pod attaches to the volume, and begins writing data
Node containing that pod crashes, loses power, or otherwise dies suddenly
Take a snapshot of the volume before it is attached to another pod
....
>

Let me some thoughts here:  In general,  a storage vendor  does  seizing of `in flight i/o`s  and also put write `barriers` in place to make sure the data is persisted correctly while snapshot is taken. Thats one of the main requirement of consistent snapshots. Thus, in above scenario, it would be ideally achieved by the SP. From CSI layer, I assume it  should not be of much concern.  iow, a snapshot  taken on an `in used volume` or  snapshot taken after a crash or node/pod loss comes pretty much to the same situation and SP or any layer responsible at storage low level should take care of it.  While it comes to controller publish/node stage, again, decision can be made by SP to node stage or controller publish either by respecting the pv spec or not. 
Regardless, the mention about the possibilities of a 'unclean journal'  situation and a workaround/solution is indeed a good point :+1: . I am not sure about the different mount options of ext4 or xfs mentioned, have to look into the details of these. 
About the direction of this PR and its completion, I feel its good wait some more input from storage sig.

bswartz: Probably the best thing to do before merging this fix would be to add documentation to the CSI spec around the PUBLISH_READONLY controller capability. Drivers that assert that capability will need to be prepared to mount unmodifiable volumes on the nodes, which means understanding how to deal with unclean journals of each of the fstypes that your driver supports.

humblec: While we wait for others to share their thoughts, let me rebase the PR and keep it updated with latest code.

pkg/controller/controller.go
```diff
@@ -509,6 +514,12 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 		if strings.ToLower(k) == "fstype" {
 			klog.Warningf(deprecationWarning("fstype", prefixedFsTypeKey, ""))
 		}
+		if strings.ToLower(k) == controllerPublishVolumeReadOnly {
```
jsafrane, line 517: While I am in favor of this PR, I don't like a new parameter in StorageClasses. Do we expect that there will be different classes with different setting for the same driver? What is the use case?

I'd prefer a new cmdline option that turns it on and where CSI driver vendors opts in (or opt out?) the whole driver into this new behavior. Alternatively, a new CSIDriver field would allow cluster admins to opt in / opt out.

bswartz: +1 for cmdline option. That puts the power in the hands of the driver author rather than the admin, and avoids polluting the Kubernetes API.

Opt-out is likely to cause more breakage than opt-in, but opt-out will get us to the desired outcome more quickly.

humblec:
Summary of the great discussions so far :)

- [ ] Unlike the initial solution we planned in the referred issue, ie `a new flag in SC`, the consensus here is to place `this new readonly option as a 'driver' arg/param` and give control to SP driver/vendor to set it.

- [ ]  There is a concern about how to be give the  platform  for `unclean` journal correction scenario in certain cases if SP publish the volume as `RO`, a couple of thoughts here is that, regardless of what is marked in PV source, SP driver can do 'RW' or 'RO' at publish/stage.  But its good to have some documentation added  to the CSI spec around the PUBLISH_READONLY controller capability. Drivers that assert that capability will need to be prepared to mount unmodifiable volumes on the nodes, which means understanding how to deal with unclean journals of each of the fstypes that your driver support.

Additional thoughts:

CSI spec so far have below 'READ_ONLY' specific capabilities available, however it is not yet used in provisioner sidecar, so the decision to mark `readonly` can skip these capabilities check for now, and just rely on the `arg/param` value of the driver discussed above. Please let me know if I failed to conclude here.
 
```
      // Can only be published once as readonly on a single node, at
      // any given time.
      SINGLE_NODE_READER_ONLY = 2;

      // Can be published as readonly at multiple nodes simultaneously.
      MULTI_NODE_READER_ONLY = 3;
```

 @msau42 @xing-yang @jsafrane @bswartz @pohly @gnufied please let me know if I missed any other suggestion so far.

humblec:
@verult as you are busy with other stuffs, with your permission, I am taking this PR  back :+1:

humblec:
[Changes in the last refresh compared to previous version]
- [x] new flag has been introduced in the provisioner sidecar
- [x] Removed SC param setting which was the initial POC based on the reference issue
- [x] Test cases have been added for provisioner sidecar flag to be `true` and `false` against MULTI_NODE_MULTI_READER capability

PTAL.. Thanks :+1:

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -103,17 +103,13 @@ var (
 	nodeDeploymentImmediateBinding = flag.Bool("node-deployment-immediate-binding", true, "Determines whether immediate binding is supported when deployed on each node.")
 	nodeDeploymentBaseDelay        = flag.Duration("node-deployment-base-delay", 20*time.Second, "Determines how long the external-provisioner sleeps initially before trying to own a PVC with immediate binding.")
 	nodeDeploymentMaxDelay         = flag.Duration("node-deployment-max-delay", 60*time.Second, "Determines how long the external-provisioner sleeps at most before trying to own a PVC with immediate binding.")
+	publishROXVol                  = flag.Bool("publish-rox-vol", false, "This option enables PV to be marked as readonly at nodepublish call if PVC accessmode has been set to ROX.")
```
msau42, line 106: From the analysis at https://github.com/kubernetes/kubernetes/issues/70505, this impacts ControllerPublish, not NodePublish

humblec: sure.. updated , thanks @msau42 . :+1:

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -103,17 +103,13 @@ var (
 	nodeDeploymentImmediateBinding = flag.Bool("node-deployment-immediate-binding", true, "Determines whether immediate binding is supported when deployed on each node.")
 	nodeDeploymentBaseDelay        = flag.Duration("node-deployment-base-delay", 20*time.Second, "Determines how long the external-provisioner sleeps initially before trying to own a PVC with immediate binding.")
 	nodeDeploymentMaxDelay         = flag.Duration("node-deployment-max-delay", 60*time.Second, "Determines how long the external-provisioner sleeps at most before trying to own a PVC with immediate binding.")
+	publishROXVol                  = flag.Bool("publish-rox-vol", false, "This option enables PV to be marked as readonly at controllerpublish call if PVC accessmode has been set to ROX.")
```
msau42, line 106: What do you think of calling the flag something like "controller-publish-readonly".

Also in the help text, use the full csi call name "ControllerPublishVolume"

pkg/controller/controller_test.go
```diff
@@ -417,7 +417,7 @@ func TestCreateDriverReturnsInvalidCapacityDuringProvision(t *testing.T) {
 
 	pluginCaps, controllerCaps := provisionCapabilities()
 	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test",
-		5, csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, false, defaultfsType, nil)
+		5, csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, false, defaultfsType, nil, true)
```
msau42, line 420: We should investigate this separately, but I think the number of function arguments is getting to the point that we should probably have an options struct.

cmd/csi-provisioner/csi-provisioner.go
msau42: Can you also update the README?

humblec: @msau42 yeah, thats a good suggestion, considering we are on controller publish, `controller-publish-readonly` suite better. Let me update the flag and README.

pkg/controller/controller_test.go
humblec: completely agree here, the arg has grown a lot and even the test case organization is difficult in general with current state. I will revisit this part in a followup PR. thanks.

jsafrane:
@humblec, please update `release-note` section of the first comment of this PR to reflect the latest changes.

humblec:
> @humblec, please update `release-note` section of the first comment of this PR to reflect the latest changes.

Sure @jsafrane

humblec:
> > @humblec, please update `release-note` section of the first comment of this PR to reflect the latest changes.
> 
> Sure @jsafrane .

Now the release note has been updated, ptal.. Thanks @jsafrane

cmd/csi-provisioner/csi-provisioner.go
humblec: @msau42 done.. thanks

chrishenzie:
> Adding a hold to give others an opportunity to review.

Removing now that PR is LGTM'd and approved.

/hold cancel

jingxu97:
Can we cherrypick this in 1.21?

PR #475, opened 2020-08-31
capacity: fix handling of topology changes with immediate binding SCs

pohly:
**What type of PR is this?**
/kind bug

A copy-and-pasted "return" instead of "continue" caused processing to
stop after encountering a storage class with immediate binding, which
had the effect that further CSIStorageCapacity objects for other
storage classes never got created.

This only happened when topology changed at runtime. When topology
was already known when the controller started, all expected objects
got created.

```release-note
CSIStorageCapacity objects were potentially incomplete when at least one storage class used "immediate binding" and topology changed while the controller was already running.
```

### DIFF
--- a/pkg/capacity/capacity.go
+++ b/pkg/capacity/capacity.go
@@ -299,7 +299,7 @@ func (c *Controller) onTopologyChanges(added []*topology.Segment, removed []*top
 			continue
 		}
 		if !c.immediateBinding && sc.VolumeBindingMode != nil && *sc.VolumeBindingMode == storagev1.VolumeBindingImmediate {
-			return
+			continue
 		}
 		for _, segment := range added {
 			c.addWorkItem(segment, sc)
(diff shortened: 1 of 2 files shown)


pohly:
/assign @msau42 

I noticed this bug when trying out the steps in the upcoming blog post https://github.com/kubernetes/website/pull/23102

I'll work around the bug in the example, therefore we don't need to hurry with getting a bug fix release out.

pkg/capacity/capacity.go
```diff
@@ -299,7 +299,7 @@ func (c *Controller) onTopologyChanges(added []*topology.Segment, removed []*top
 			continue
 		}
 		if !c.immediateBinding && sc.VolumeBindingMode != nil && *sc.VolumeBindingMode == storagev1.VolumeBindingImmediate {
-			return
+			continue
```
msau42, line 302: Can we add a unit test for this?

pohly: We can, but it's not perfect: we need a specific ordering of storage classes (first the one with immediate binding, then the one with late binding), and there's no guarantee that the informer will list them in any particular order. But in practice, the informer seems to use "first in, first out", so I was able to trigger the problem.

But worse, `onTopologyChanges` wasn't tested at all. Fixed...

msau42:
/lgtm
/approve

Can you cherry-pick this to release-2.0 branch?

pohly:
> Can you cherry-pick this to release-2.0 branch?

See PR #476

PR #476, opened 2020-09-03
capacity: fix handling of topology changes with immediate binding SCs

pohly:
**What type of PR is this?**
/kind bug

A copy-and-pasted "return" instead of "continue" caused processing to
stop after encountering a storage class with immediate binding, which
had the effect that further CSIStorageCapacity objects for other
storage classes never got created.

This only happened when topology changed at runtime. When topology
was already known when the controller started, all expected objects
got created.

(cherry picked from commit 9352d216b40265196facd79f2e040fbdffee32d4, https://github.com/kubernetes-csi/external-provisioner/pull/475)

```release-note
CSIStorageCapacity objects were potentially incomplete when at least one storage class used "immediate binding" and topology changed while the controller was already running.
```

### DIFF
--- a/pkg/capacity/capacity.go
+++ b/pkg/capacity/capacity.go
@@ -299,7 +299,7 @@ func (c *Controller) onTopologyChanges(added []*topology.Segment, removed []*top
 			continue
 		}
 		if !c.immediateBinding && sc.VolumeBindingMode != nil && *sc.VolumeBindingMode == storagev1.VolumeBindingImmediate {
-			return
+			continue
 		}
 		for _, segment := range added {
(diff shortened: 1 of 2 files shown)


msau42:
The target branch looks like it's still the master branch? (I'm unclear why github isn't showing this as a no-op)

pohly:
> The target branch looks like it's still the master branch? 

Sorry, I did indeed pick the wrong target. 

> (I'm unclear why github isn't showing this as a no-op)

That would have been nicer. I suspect it just shows the diff against the common ancestor, not the diff against current master.

Replaced with https://github.com/kubernetes-csi/external-provisioner/pull/477

PR #479, opened 2020-09-16
optional VolumeAttachments check

pohly:
**What type of PR is this?**

/kind feature

Watching all VolumeAttachments in the cluster even when the CSI driver
itself doesn't support attach/detach causes unnecessary overhead. It
is now automatically enabled if and only if the driver reports the
PUBLISH_UNPUBLISH_VOLUME controller capability.

Checking the capability is better than checking the "skip attach" flag
in the CSIDriver object because the capability is readily available
and constant. It also has the advantage that the watch is disabled for
drivers like csi-host-path-driver which do not support attach/detach
but don't use "skip attach" (for whatever reason).

```release-note
the watch of VolumeAttachements is avoided for drivers which don't support attach/detach (aka controller PUBLISH_UNPUBLISH_VOLUME)
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -202,7 +202,13 @@ func main() {
 	claimLister := factory.Core().V1().PersistentVolumeClaims().Lister()
 
 	var csiNodeLister storagelistersv1.CSINodeLister
-	vaLister := factory.Storage().V1().VolumeAttachments().Lister()
+	var vaLister storagelistersv1.VolumeAttachmentLister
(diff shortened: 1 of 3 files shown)


pohly:
/assign @jsafrane 

Note that I have only tested this on a real cluster with drivers that don't support attach/detach. It may be useful to try it once with a driver that does.

Besides the obvious "this avoids useless work", I'm also hoping that it avoids the issue that I am seeing in PMEM-CSI CI testing where volumes don't get deleted because they are "still attached". I've not been able to root-cause that one, though. My theory is that sometimes the "skip attach" flag is missed and thus a useless VolumeAttachment object gets created, but I should have logging of those objects now and don't see anything. :shrug:

msau42:
/lgtm
/approve

I suspect the issue you're hitting is https://github.com/kubernetes/kubernetes/issues/89173

PR #493, opened 2020-10-06
WIP: prow.sh: update to kind 0.9, support Kubernetes 1.19

pohly:
That release adds support for recent Kubernetes releases like 1.19 and
simplifies configuration of feature gates.

With Kubernetes 1.19, new feature gates were introduced which might
become relevant for the alpha Prow jobs.

/kind cleanup

Testing of the prow.sh change - do not merge.

```release-note
NONE
```

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -33,9 +33,9 @@
 # The expected environment is:
 # - $GOPATH/src/<import path> for the repository that is to be tested,
 #   with PR branch merged (when testing a PR)
+# - optional: bazel installed (when testing against Kubernetes master),
+#   must be recent enough for Kubernetes master
 # - running on linux-amd64
-# - bazel installed (when testing against Kubernetes master), must be recent
-#   enough for Kubernetes master
 # - kind (https://github.com/kubernetes-sigs/kind) installed
 # - optional: Go already installed
 
@@ -100,6 +100,9 @@ kindest/node:v1.16.15@sha256:a89c771f7de234e6547d43695c7ab047809ffc71a0c3b65aa54
 kindest/node:v1.15.12@sha256:d9b939055c1e852fe3d86955ee24976cab46cba518abcb8b13ba70917e6547a6
 kindest/node:v1.14.10@sha256:ce4355398a704fca68006f8a29f37aafb49f8fc2f64ede3ccd0d9198da910146
 kindest/node:v1.13.12@sha256:1c1a48c2bfcbae4d5f4fa4310b5ed10756facad0b7a2ca93c7a4b5bae5db29f5" "kind images"
+
+# Use kind node-image --type=bazel by default, but allow to disable that.
+configvar CSI_PROW_USE_BAZEL true "use Bazel during 'kind node-image' invocation"
 
 # ginkgo test runner version to use. If the pre-installed version is
 # different, the desired version is built from source.
@@ -534,6 +537,8 @@ start_cluster () {
 
     # Try to find a pre-built kind image if asked to use a specific version.
     if ! [[ "${CSI_PROW_KUBERNETES_VERSION}" =~ ^release-|^latest$ ]]; then
+        # Ignore: See if you can use ${variable//search/replace} instead.
+        # shellcheck disable=SC2001
         major_minor=$(echo "${CSI_PROW_KUBERNETES_VERSION}" | sed -e 's/^\([0-9]*\)\.\([0-9]*\).*/\1.\2/')
         for i in ${CSI_PROW_KIND_IMAGES}; do
             if echo "$i" | grep -q "kindest/node:v${major_minor}"; then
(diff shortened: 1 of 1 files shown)


release-tools/prow.sh
```diff
@@ -137,14 +137,23 @@ configvar CSI_PROW_BUILD_JOB true "building code in repo enabled"
...
 # If the version is prefixed with "release-", then nothing
 # is overridden.
-override_k8s_version "1.15.3"
+override_k8s_version "1.13.12"
+override_k8s_version "1.14.10"
+override_k8s_version "1.15.12"
+override_k8s_version "1.16.15"
+override_k8s_version "1.17.11"
+override_k8s_version "1.18.8"
+override_k8s_version "1.19.1"
```
pohly, line 156: Overriding the K8S version was added in https://github.com/kubernetes-csi/external-provisioner/pull/313 because KinD only has images for certain versions.

But the recommendations in the KinD release notes go even further than that: it can happen that `kindest/node:v1.19.1` becomes incompatible with kind v0.9 if the image needs to be rebuilt for, e.g. kind 1.0.

@msau42: should we switch to using images with hash, i.e. `kindest/node:v1.19.1@sha256:98cf5288864662e37115e362b23e4369c8c4a408f99cbc06e58ac30ddc721600`?

msau42: Sure let's use images with hash.

But I also want to move away from the model of specifying versions in prow.sh, so that we don't need to update prow.sh every time we need to add a new K8s version (although maybe unavoidable until kind apis stabilze more). I prefer to put the burden on the caller to make sure the kind version + k8s version specified are compatible.

release-tools/prow.sh
```diff
@@ -137,14 +137,23 @@ configvar CSI_PROW_BUILD_JOB true "building code in repo enabled"
...
+#
+# We need to override CSI_PROW_KUBERNETES_VERSION
 # passed in by our CI/pull jobs to the versions that
-# kind v0.5.0 supports.
+# kind v0.9.0 supports (https://github.com/kubernetes-sigs/kind/releases/tag/v0.9.0).
 #
 # If the version is prefixed with "release-", then nothing
 # is overridden.
-override_k8s_version "1.15.3"
+override_k8s_version "1.13.12"
```
msau42, line 150: We don't support versions < 1.17 anymore

release-tools/prow.sh
```diff
@@ -137,14 +137,23 @@ configvar CSI_PROW_BUILD_JOB true "building code in repo enabled"
 configvar CSI_PROW_KUBERNETES_VERSION 1.17.0 "Kubernetes"
 
 # This is a hack to workaround the issue that each version
-# of kind currently only supports specific patch versions of
-# Kubernetes. We need to override CSI_PROW_KUBERNETES_VERSION
+# of kind currently only has images for specific patch versions of
+# Kubernetes, as documented in the release notes of each
+# kind release.
+#
+# We need to override CSI_PROW_KUBERNETES_VERSION
```
msau42, line 144: I already changed our CI jobs to explicitly specify the K8s version so we should update them there too: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-csi/gen-jobs.sh#L25

release-tools/prow.sh
```diff
@@ -337,9 +346,10 @@ configvar CSI_PROW_E2E_ALPHA "$(get_versioned_variable CSI_PROW_E2E_ALPHA "${csi
 # it anymore for older releases.
 configvar CSI_PROW_E2E_ALPHA_GATES_1_15 'VolumeSnapshotDataSource=true,ExpandCSIVolumes=true' "alpha feature gates for Kubernetes 1.15"
 configvar CSI_PROW_E2E_ALPHA_GATES_1_16 'VolumeSnapshotDataSource=true' "alpha feature gates for Kubernetes 1.16"
-# TODO: add new CSI_PROW_ALPHA_GATES_xxx entry for future Kubernetes releases and
-# add new gates to CSI_PROW_E2E_ALPHA_GATES_LATEST.
-configvar CSI_PROW_E2E_ALPHA_GATES_LATEST '' "alpha feature gates for latest Kubernetes"
+configvar CSI_PROW_E2E_ALPHA_GATES_1_17 '' "alpha feature gates for Kubernetes 1.17"
```
msau42, line 349: We don't run alpha tests on versions < latest

pohly:
@msau42 I've pushed the updated commit, please take another look.

release-tools/prow.sh
msau42: I disagree. This repo is not intended to be used for anything other than CI for these repos. We don't test or support these older configurations and it's a maintenance burden for us if there are problems.

Downstream repos that consume this do so at their own risk and need to do their own maintenance if they fall out of the Kubernetes support window.

release-tools/prow.sh
msau42: I would like to work towards a goal where adding support for a new K8s version doesn't require updating the release framework to understand the new version. I'm looking at k/k and kubetest as an example. But maybe that is not achievable at the moment because of the kind image dependencies.

release-tools/prow.sh
pohly: I've removed them in https://github.com/kubernetes-csi/csi-release-tools/pull/107

release-tools/prow.sh
pohly: Yes, kind is the problem here. There's just no clean "API" for installing kind and configuring it for different Kubernetes versions.

We either have to have all of that in the component or in the job definition. In the latter case, the prow.sh would have to be given some helper script that complete replaces the current `start_kubernetes`. The downside then will be that it becomes much harder to try out changes to prow.sh locally because one has to provide that helper script.

pohly:
/close

Merged via https://github.com/kubernetes-csi/csi-release-tools/pull/107

PR #501, opened 2020-10-08
topology: also control for immediate binding

pohly:
**What type of PR is this?**
/kind feature

A CSI driver for local volumes doesn't care about the (long!) list of nodes where the driver is currently running. It needs to track that itself anyway if it supports volume provisioning with immediate binding. The `CreateVolumeRequest` can be made a lot shorter by omitting the topology requirements in that case.

```release-note
TBD
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ See the [storage capacity section](#capacity-support) below for details.
 
 * `--strict-topology`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of delayed binding. See [the table below](#topology-support) for an explanation how this option changes the result. This option has no effect if either `Topology` feature is disabled or `Immediate` volume binding mode is used.
 
+* `--immediate-topology`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of immediate binding. See [the table below](#topology-support) for an explanation how this option changes the result. This option has no effect if either `Topology` feature is disabled or `WaitForFirstConsumer` (= delayed) volume binding mode is used. The default is true, so use `--immediate-topology=false` to disable it. It should not be disabled if the CSI driver might create volumes in a topology segment that is not accessible in the cluster. Such a driver should use the topology information to create new volumes where they can be accessed.
+
 * `--kubeconfig <path>`: Path to Kubernetes client configuration that the external-provisioner uses to connect to Kubernetes API server. When omitted, default token provided by Kubernetes will be used. This option is useful only when the external-provisioner does not run as a Kubernetes pod, e.g. for debugging. Either this or `--master` needs to be set if the external-provisioner is being run out of cluster.
 
 * `--master <url>`: Master URL to build a client config from. When omitted, default token provided by Kubernetes will be used. This option is useful only when the external-provisioner does not run as a Kubernetes pod, e.g. for debugging. Either this or `--kubeconfig` needs to be set if the external-provisioner is being run out of cluster.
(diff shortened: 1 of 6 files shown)


pkg/controller/topology.go
```diff
@@ -181,6 +181,12 @@ func GenerateAccessibilityRequirements(
 			// Distribute out one of the OR layers in allowedTopologies
 			requisiteTerms = flatten(allowedTopologies)
 		} else {
+			if strictTopology {
+				// Don't specify any topology requirements because neither the PVC nor
+				// the storage class have limitations.
+				return nil, nil
+			}
```
pohly, line 188: @msau42: the change itself makes sense for local volume CSI drivers like PMEM-CSI.

Where I am undecided is whether this should be controlled via `strictTopology` or some new flag. The current definition of "strict topology" in the top-level README explicitly describes it as "ignored for immediate binding", so checking it here changes the semantic.

As user of the flag I am fine with that change; it actually would be simpler.

msau42: hmmmm thinking about it closer, this can break drivers that support topology, so it's probably better to have a new flag. For example, for a typical cloud volume that supports zones, you may have a cluster that is deployed in only a subset of the zones. Drivers would depend on this aggregation to get the list of all valid zones.

Going back to the local volume case, how would a driver that has a central controller pick which node should be provisioned? Wouldn't that driver already need to keep some list of all nodes in the cluster to choose from?

pohly: > For example, for a typical cloud volume that supports zones, you may have a cluster that is deployed in only a subset of the zones. Drivers would depend on this aggregation to get the list of all valid zones.

Do those use `--strict-topology`? I'm not sure. But you are right, as we can't know whether they do and changing the meaning of the `--strict-topology` would break them, we have to add another flag.

> Going back to the local volume case, how would a driver that has a central controller pick which node should be provisioned? Wouldn't that driver already need to keep some list of all nodes in the cluster to choose from?

A central controller needs a way to communicate with its nodes when it wants to provision a volume there. Just knowing what its nodes are is not enough.

In PMEM-CSI, we let nodes register themselves with the central controller, which also stores information about a port on which the node service can be reached to provision volumes. The list of nodes from external-provisioner in the CreateVolumeRequest topology alone is not enough and redundant.

pohly: I decided to call this new flag `--topology-hints` because it felt like in this particular case, external-provisioner was providing additional information that wasn't strictly required, but may be useful. As always, naming is up for debate.. :stuck_out_tongue_closed_eyes:

msau42: I don't think topology hints really says anything useful about what it does. What about something like "no-immediate-topology" to mean that if it's immediate binding, we don't pass any topology down and let the driver decide.

README.md
```diff
@@ -89,6 +89,8 @@ See the [storage capacity section](#capacity-support) below for details.
 
 * `--strict-topology`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of delayed binding. See [the table below](#topology-support) for an explanation how this option changes the result. This option has no effect if either `Topology` feature is disabled or `Immediate` volume binding mode is used.
 
+* `--topology-hints`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of immediate binding. See [the table below](#topology-support) for an explanation how this option changes the result. This option has no effect if either `Topology` feature is disabled or `WaitForFirstConsumer` (= delayed) volume binding mode is used.
```
msau42, line 92: It should be noted that this option should not be used for drivers that can see more of a storage system's topology than what's actually available in the cluster.

pohly: Added this:
> It should not be used if the CSI driver might create volumes in a topology segment that is not accessible in the cluster. Such a driver should use the topology information to create new volumes where they can be accessed.

pkg/controller/topology.go
pohly: That's better. I've updated the commit accordingly, please take another look.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -68,7 +68,8 @@ var (
 	enableLeaderElection = flag.Bool("leader-election", false, "Enables leader election. If leader election is enabled, additional RBAC rules are required. Please refer to the Kubernetes CSI documentation for instructions on setting up these RBAC rules.")
 
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "Namespace where the leader election resource lives. Defaults to the pod namespace if not set.")
-	strictTopology          = flag.Bool("strict-topology", false, "Passes only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.")
+	strictTopology          = flag.Bool("strict-topology", false, "Late binding: pass only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.")
+	immediateTopology       = flag.Bool("no-immediate-topology", true, "Immediate binding: pass aggregated cluster topologies for all nodes where the CSI driver is available (enabled, the default) or no topology requirements (if disabled).")
```
msau42, line 72: Can we name the variable noImmediateTopology to match?

README.md
```diff
@@ -89,6 +89,8 @@ See the [storage capacity section](#capacity-support) below for details.
 
 * `--strict-topology`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of delayed binding. See [the table below](#topology-support) for an explanation how this option changes the result. This option has no effect if either `Topology` feature is disabled or `Immediate` volume binding mode is used.
 
+* `--no-immediate-topology`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of immediate binding. See [the table below](#topology-support) for an explanation how this option changes the result. This option has no effect if either `Topology` feature is disabled or `WaitForFirstConsumer` (= delayed) volume binding mode is used. It should not be used if the CSI driver might create volumes in a topology segment that is not accessible in the cluster. Such a driver should use the topology information to create new volumes where they can be accessed.
```
msau42, line 92: Can you document the default

msau42 (commented):
/approve

Just some nits

cmd/csi-provisioner/csi-provisioner.go
pohly: I got stuck halfway here removing a double negation: `--no-immediate-topology=false` is what I wanted to avoid. I just missed updating the flag name, variable name and default are correct already. Will fix.

pohly: Fixed. Only the flag name and documentation changed, the implementation is the same as before.

README.md
pohly: Done.

jsafrane:
I prefer `immediate-topology=true` over `no-immediate-topology=false` as the default.

/lgtm

PR #524, opened 2020-11-03
distributed provisioning

### LINKED ISSUE #487: deployment without central controller
For CSI drivers which manage local storage it is difficult to provide dynamic provisioning support because the central driver instance needs some way of communicating with the driver instance on each node. It would be nice if:
- external-provisioner could be deployed on each node together with the CSI driver
- These external-provisioner instances collaborated with each other on provisioning of PVCs, with support for late binding *and* immediate binding.

pohly:
**What type of PR is this?**
/kind feature

A CSI driver which manages local resources typically has no control plane and thus (currently) can only provision volumes on a single node. With this PR, external-provisioner gets extended to support deployment alongside a local CSI driver on each node.

The csi-driver-host-path deployment could be extended to use this.

Fixes #487

This is based on https://github.com/kubernetes-csi/external-provisioner/pull/367. Out of courtesy to @jsanda, the original commit is the one from that PR although it gets changed substantially later on. I could also squash more aggressively, if that is desired.

TODO:
- [x] make the UML diagram nicer (text formatting potentially layout)
- [ ] scale testing: done on 100 nodes, maybe try something higher
- [x] automated tests for the new code paths
- [x] storage capacity tracking support also for distributed provisioning

```release-note
external-provisioner can be deployed alongside a CSI driver on each node to manage local volumes.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ See the [storage capacity section](#capacity-support) below for details.
 
 * `--node-deployment-max-delay`: Determines how long the external-provisioner sleeps at most before trying to own a PVC with immediate binding. Defaults to 60 seconds.
 
+* `--local-topology`: Instead of watching Node and CSINode objects, use only the topology provided by the CSI driver. Only valid in combination with `--node-deployment`. Disabled by default, but recommended for drivers which have a single topology key with different values for each node (i.e. local volumes).
+
 #### Other recognized arguments
 * `--feature-gates <gates>`: A set of comma separated `<feature-name>=<true|false>` pairs that describe feature gates for alpha/experimental features. See [list of features](#feature-status) or `--help` output for list of recognized features. Example: `--feature-gates Topology=true` to enable Topology feature that's disabled by default.
 
@@ -271,6 +273,8 @@ each CSI driver on different nodes. The CSI driver deployment must:
   to match the expected cluster size and desired response times
   (only relevant when there are storage classes with immediate binding,
   see below for details)
+- use `--local-topology` if volumes are only accessible inside the node
+  where they get provisioned
 - set the `NODE_NAME` environment variable to the name of the Kubernetes node
 - implement `GetCapacity`
 
@@ -303,11 +307,10 @@ can own it.
 
 The `--node-deployment-base-delay` parameter determines the initial
 wait period. It also sets the jitter, so in practice the initial wait period will be
-in the range from zero to the base delay. After a collision, the delay
-increases exponentially. If the value is high, volumes with immediate
-binding get created more slowly. If it is low, then the risk of
-conflicts while setting the "selected node" annotation increases and
-the apiserver load will be higher.
+in the range from zero to the base delay. If the value is high,
+volumes with immediate binding get created more slowly. If it is low,
+then the risk of conflicts while setting the "selected node"
+annotation increases and the apiserver load will be higher.
 
 There is an exponential backoff per PVC which is used for unexpected
 problems. Normally, an owner for a PVC is chosen during the first
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ See the [storage capacity section](#capacity-support) below for details.
 
 * `--node-deployment`: Enables deploying the external-provisioner together with a CSI driver on nodes to manage node-local volumes. Off by default.
 
-* `--node-deployment-immediate-binding`: Determines whether immediate binding is supported when deployed on each node. Enabled by default, use `--node-deployment-immediate-binding=false` if not desired.
+* `--node-deployment-immediate-binding`: Determines whether immediate binding is supported when deployed on each node. Enabled by default, use `--node-deployment-immediate-binding=false` if not desired. Disabling it may be useful for example when a custom controller will select nodes for PVCs.
 
 * `--node-deployment-base-delay`: Determines how long the external-provisioner sleeps initially before trying to own a PVC with immediate binding. Defaults to 20 seconds.
 
@@ -335,7 +335,10 @@ volume was probably higher, but that wasn't measured.
 
 Note that the QPS settings of kube-controller-manager and
 external-provisioner have to be increased at the moment (Kubernetes
-1.19) to provision volumes faster than around 4 volumes/second.
+1.19) to provision volumes faster than around 4 volumes/second. Those
+settings will eventually get replaced with [flow control in the API
+server
+itself](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/).
 
 Beware that if *no* node has sufficient storage available, then also
 no `CreateVolume` call is attempted and thus no events are generated
@@ -364,6 +367,9 @@ When an administrator is sure that the node is never going to come
 back, then the local volumes can be removed manually:
 - force-delete objects: `kubectl delete pv <pv> --wait=false --grace-period=0 --force`
 - remove all finalizers: `kubectl patch pv <pv> -p '{"metadata":{"finalizers":null}}'`
+
+It may also be necessary to scrub disks before reusing them because
+the CSI driver had no chance to do that.
 
 If there still was a PVC which was bound to that PV, it then will be
 moved to phase "Lost". It has to be deleted and re-created if still
--- a/README.md
+++ b/README.md
@@ -342,6 +342,27 @@ instead implement a custom policy in a separate controller which sets
 the "selected node" annotation to trigger local provisioning on the
 desired node.
 
+### Deleting local volumes after a node failure or removal
+
+When a node with local volumes gets removed from a cluster before
+deleting those volumes, the PV and PVC objects may still exist. It may
+be possible to remove the PVC normally if the volume was not in use by
+any pod on the node, but normal deletion of the volume and thus
+deletion of the PV is not possible anymore because the CSI driver
+instance on the node is not available or reachable anymore and therefore
+Kubernetes cannot be sure that it is okay to remove the PV.
+
+When an administrator is sure that the node is never going to come
+back, then the local volumes can be removed manually:
+- force-delete objects: `kubectl delete pv <pv> --wait=false --grace-period=0 --force`
+- remove all finalizers: `kubectl patch pv <pv> -p '{"metadata":{"finalizers":null}}'`
+
+If there still was a PVC which was bound to that PV, it then will be
+moved to phase "Lost". It has to be deleted and re-created if still
+needed because no new volume will be created for it. Editing the PVC
+to revert it to phase "Unbound" is not allowed by the Kubernetes
+API server.
+
 ## Community, discussion, contribution, and support
 
 Learn how to engage with the Kubernetes community on the [community page](http://kubernetes.io/community/).
--- a/README.md
+++ b/README.md
@@ -84,6 +84,16 @@ See the [storage capacity section](#capacity-support) below for details.
 
 * `--capacity-for-immediate-binding <bool>`: Enables producing capacity information for storage classes with immediate binding. Not needed for the Kubernetes scheduler, maybe useful for other consumers or for debugging. Defaults to `false`.
 
+##### Distributed provisioning
+
+* `--node-deployment`: Enables deploying the external-provisioner together with a CSI driver on nodes to manage node-local volumes. Off by default.
+
+* `--node-deployment-immediate-binding`: Determines whether immediate binding is supported when deployed on each node. Enabled by default, use `--node-deployment-immediate-binding=false` if not desired.
+
+* `--node-deployment-base-delay`: Determines how long the external-provisioner sleeps initially before trying to own a PVC with immediate binding. Defaults to 20 seconds.
+
+* `--node-deployment-max-delay`: Determines how long the external-provisioner sleeps at most before trying to own a PVC with immediate binding. Defaults to 60 seconds.
+
 #### Other recognized arguments
 * `--feature-gates <gates>`: A set of comma separated `<feature-name>=<true|false>` pairs that describe feature gates for alpha/experimental features. See [list of features](#feature-status) or `--help` output for list of recognized features. Example: `--feature-gates Topology=true` to enable Topology feature that's disabled by default.
 
@@ -241,6 +251,93 @@ The external-provisioner optionally exposes an HTTP endpoint at address:port spe
 
 * Metrics path, as set by `--metrics-path` argument (default is `/metrics`).
 * Leader election health check at `/healthz/leader-election`. It is recommended to run a liveness probe against this endpoint when leader election is used to kill external-provisioner leader that fails to connect to the API server to renew its leadership. See https://github.com/kubernetes-csi/csi-lib-utils/issues/66 for details.
+
+### Deployment on each node
+
+Normally, external-provisioner is deployed once in a cluster and
+communicates with a control instance of the CSI driver which then
+provisions volumes via some kind of storage backend API. CSI drivers
+which manage local storage on a node don't have such an API that a
+central controller could use.
+
+To support this case, external-provisioner can be deployed alongside
+each CSI driver on different nodes. The CSI driver deployment must:
+- support topology, usually with one topology key
+  ("csi.example.org/node") and the Kubernetes node name as value
+- use a service account that has the same RBAC rules as for a normal
+  deployment
+- invoke external-provisioner with `--node-deployment`
+- tweak `--node-deployment-base-delay` and `--node-deployment-max-delay`
+  to match the expected cluster size and desired response times
+  (only relevant when there are storage classes with immediate binding,
+  see below for details)
+- set the `NODE_NAME` environment variable to the name of the Kubernetes node
+- implement `GetCapacity`
+
+Usage of `--strict-topology` and `--immediate-topology=false` is
+recommended because it makes the `CreateVolume` invocations simpler.
+
+Volume provisioning with late binding works as before, except that
+each external-provisioner instance checks the "selected node"
+annotation and only creates volumes if that node is the one it runs
+on. It also only deletes volumes on its own node.
+
+Immediate binding is also supported, but not recommended. It is
+implemented by letting the external-provisioner instances assign a PVC
+to one of them: when they see a new PVC with immediate binding, they
+all attempt to set the "selected node" annotation with their own node
+name as value. Only one update request can succeed, all others get a
+"conflict" error and then know that some other instance was faster. To
+avoid the thundering herd problem, each instance waits for a random
+period before issuing an update request.
+
+When `CreateVolume` call fails with `ResourcesExhausted`, the normal
+recovery mechanism is used, i.e. the external-provisioner instance
+removes the "selected node" annotation and the process repeats. But
+this triggers events for the PVC and delays volume creation, in
+particular when storage is exhausted on most nodes. Therefore
+external-provisioner checks with `GetCapacity` *before* attempting to
+own a PVC whether the currently available capacity is sufficient for
+the volume. When it isn't, the PVC is ignored and some other instance
+can own it.
+
+The `--node-deployment-base-delay` parameter determines the initial
+wait period. It also sets the jitter, so in practice the initial wait period will be
+in the range from zero to the base delay. After a collision, the delay
+increases exponentially. If the value is high, volumes with immediate
+binding get created more slowly. If it is low, then the risk of
+conflicts while setting the "selected node" annotation increases and
+the apiserver load will be higher.
+
+There is an exponential backoff per PVC which is used for unexpected
+problems. Normally, an owner for a PVC is chosen during the first
+attempt, so most PVCs will use the base delays. A maximum can be set
+with `--node-deployment-max-delay` anyway, to avoid very long delays
+when something went wrong repeatedly.
+
+During scale testing with 100 external-provisioner instances, a base
+delay of 20 seconds worked well. When provisioning 3000 volumes, there
+were only 500 conflicts which the apiserver handled without getting
+overwhelmed. The average provisioning rate of around 40 volumes/second
+was the same as with a delay of 10 seconds. The worst-case latency per
+volume was probably higher, but that wasn't measured.
+
+Note that the QPS settings of kube-controller-manager and
+external-provisioner have to be increased at the moment (Kubernetes
+1.19) to provision volumes faster than around 4 volumes/second.
+
+Beware that if *no* node has sufficient storage available, then also
+no `CreateVolume` call is attempted and thus no events are generated
+for the PVC, i.e. some other means of tracking remaining storage
+capacity must be used to detect when the cluster runs out of storage.
+
+Because PVCs with immediate binding get distributed randomly among
+nodes, they get spread evenly. If that is not desirable, then it is
+possible to disable support for immediate binding in distributed
+provisioning with `--node-deployment-immediate-binding=false` and
+instead implement a custom policy in a separate controller which sets
+the "selected node" annotation to trigger local provisioning on the
+desired node.
 
 ## Community, discussion, contribution, and support
 
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 See the [storage capacity section](#capacity-support) below for details.
 
-* `--capacity-controller-deployment-mode=central`: Setting this enables producing CSIStorageCapacity objects with capacity information from the driver's GetCapacity call. 'central' is currently the only supported mode. Use it when there is just one active provisioner in the cluster. The default is to not produce CSIStorageCapacity objects.
+* `--capacity-controller-deployment-mode=central|local`: Setting this enables producing CSIStorageCapacity objects with capacity information from the driver's GetCapacity call. Use `central` when there is just one active external-provisioner in the cluster. Use `local` when deploying external-provisioner on each node with distributed provisioning. The default is to not produce CSIStorageCapacity objects.
 
 * `--capacity-ownerref-level <levels>`: The level indicates the number of objects that need to be traversed starting from the pod identified by the POD_NAME and POD_NAMESPACE environment variables to reach the owning object for CSIStorageCapacity objects: 0 for the pod itself, 1 for a StatefulSet, 2 for a Deployment, etc. Defaults to `1` (= StatefulSet).
 
@@ -151,7 +151,7 @@ determine with the `POD_NAME/POD_NAMESPACE` environment variables and
 the `--capacity-ownerref-level` parameter. Other solutions will be
 added in the future.
 
-To enable this feature in a driver deployment (see also the
+To enable this feature in a driver deployment with a central controller (see also the
 [`deploy/kubernetes/storage-capacity.yaml`](deploy/kubernetes/storage-capacity.yaml)
 example):
 
@@ -167,7 +167,7 @@ example):
         fieldRef:
         fieldPath: metadata.name
 ```
-- Add `--enable-capacity=central` to the command line flags.
+- Add `--capacity-controller-deployment-mode=central` to the command line flags.
 - Add `StorageCapacity: true` to the CSIDriver information object.
   Without it, external-provisioner will publish information, but the
   Kubernetes scheduler will ignore it. This can be used to first
@@ -182,7 +182,7 @@ example):
   with `--capacity-threads`.
 - Optional: enable producing information also for storage classes that
   use immediate volume binding with
-  `--enable-capacity=immediate-binding`. This is usually not needed
+  `--capacity-for-immediate-binding`. This is usually not needed
   because such volumes are created by the driver without involving the
   Kubernetes scheduler and thus the published information would just
   be ignored.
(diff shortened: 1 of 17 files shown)


pohly:
Before I proceed with this feature, I'd like to get feedback whether the approach of letting the "selected node" annotation be set by the external-provisioner is acceptable for volumes with immediate binding. I have tested this on a 50 node cluster and it worked nicely.

My argument why it is compatible with the current design is that the state of the PVC is indistinguishable from a PVC that was assigned to a node while the storage class used late binding. This can happen and if the current code cannot handle this, then it needs to be fixed. However, I don't see why it should fail (famous last words...).

@msau42, @jsafrane: what do you think about the UML diagram? That commit itself is unrelated to this PR, I could also submit it separately. I added it here because it makes it easier to understand how the node-local deployment fits into the overall architecture.

pohly:
Hmm, looks like my recent rebasing broke the unit tests. Will fix that...

jingxu97:
cc @jingxu97

pkg/controller/ratelimiter.go
```diff
@@ -0,0 +1,55 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
jsafrane, line 2: 2020

pkg/controller/controller.go
```diff
@@ -1108,6 +1251,196 @@ func (p *csiProvisioner) volumeHandleToId(handle string) string {
...
+// becomeOwner updates the PVC with the current node as selected node.
+// Returns an error if something unexpectedly failed, otherwise an updated PVC with
+// the current node selected or nil if not the owner.
+func (nc *internalNodeDeployment) becomeOwner(ctx context.Context, client kubernetes.Interface, claim *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaim, error) {
+	attempts := nc.rateLimiter.NumRequeues(claim.UID)
+	delay := nc.rateLimiter.When(claim.UID)
+	klog.V(5).Infof("will try to become owner of PVC %s/%s in %s, attempt #%d", claim.Namespace, claim.Name, delay, attempts)
+	sleep, cancel := context.WithTimeout(ctx, delay)
+	defer cancel()
+	ticker := time.NewTicker(10 * time.Millisecond)
```
jsafrane, line 1377: Did you consider using claimInformer to get the events?

README.md
```diff
@@ -87,8 +87,12 @@ See the [storage capacity section](#capacity-support) below for details.
...
 
+* `--node-deployment`: Enables deploying the external-provisioner together with a CSI driver on nodes to manage node-local volumes.
+
 * `--strict-topology`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of delayed binding. See [the table below](#topology-support) for an explanation how this option changes the result. This option has no effect if either `Topology` feature is disabled or `Immediate` volume binding mode is used.
 
+* `--no-immediate-topology`: This controls what topology information is passed to `CreateVolumeRequest.AccessibilityRequirements` in case of immediate binding. See [the table below](#topology-support) for an explanation how this option changes the result. This option has no effect if either `Topology` feature is disabled or `WaitForFirstConsumer` (= delayed) volume binding mode is used. It should not be used if the CSI driver might create volumes in a topology segment that is not accessible in the cluster. Such a driver should use the topology information to create new volumes where they can be accessed.
```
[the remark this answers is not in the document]
pohly: Yes, it's an older version of the commit. Let's merge #501 and I'll rebase.

pkg/controller/controller.go
pohly: Instead of polling? Yes, I had been thinking about that, but couldn't come up with a good way to hook up the informer callback with this function here.

Callbacks can only be added, but not removed. So I would have to install some generic callback that gets told at runtime who is interested in which PVC - felt rather complicated for a rather small improvement in latency.

doc/provisioning.puml
```diff
@@ -0,0 +1,31 @@
+@startuml
+(*) -->[user creates PVC] "PVC exists"
+-->[PV controller checks storage class and existing PVs, sets the\nvolume.beta.kubernetes.io/storage-provisioner annotation] "PVC ready for provisioning" as PVC_READY
```
jsafrane, line 3: I think that PV controller does not add `volume.beta.kubernetes.io/storage-provisioner` right away. It's:

```
PVC exists:
  if  late binding:
      wait for pod + annotate the PVC
  else (immediate binding): 
      annotate the PVC right away
```

pkg/controller/controller.go
```diff
@@ -326,6 +356,48 @@ func NewCSIProvisioner(client kubernetes.Interface,
...
+		// initial delay will be somewhere between 0 and
+		// baseDelay.
+		//
+		// It must be large enough that one instance has the chance to
+		// become owner of a PVC and get that change propagated to the
+		// other instances before those wake up to try that
+		// themselves.
+		//
+		// With a value of 10 seconds, when creating 5000
+		// volumes on a cluster with 50 instances only ~300
```
jsafrane, line 373: Just to be sure, nr. of conflicts does not grow linearly with nr. of nodes, right? 5000 nodes != 30 000 conflicts, but much more. How much?

pkg/controller/controller.go
```diff
@@ -326,6 +356,48 @@ func NewCSIProvisioner(client kubernetes.Interface,
...
+		// over 6000 conflicts occurred. The rate of volume
+		// provisioning was the same in both cases.
+		//
+		// The latency per volume was probably higher with the
+		// higher delay, but that was not measured.
+		//
+		// It might make sense to make this value configurable so
+		// that CSI driver deployments can tweak it depending
+		// on their needs.
+		baseDelay := 10 * time.Second
```
jsafrane, line 384: This IMO needs to be configurable or grow with the cluster size (as a separate PR)

jsafrane (commented):
I think there will be some changes necessary for scalability, but I'm fine with merging this and fix them later after we get some numbers.

One thing troubles me more: **PV deletion**.
What should happen when a node is removed from the cluster? All its PVs can't be deleted - there is no provisioner that would do that. Is it just a documentation issue? Or shall there be a cluster-level provisioner that deletes PVs from missing nodes? What if the node rejoins the cluster?

pohly:
> I think there will be some changes necessary for scalability, but I'm fine with merging this and fix them later after we get some numbers.

I just had a call today regarding that: I will get access to a 100 node cluster from SIG-Scalability that I can run scalability tests on. We don't need to merge for that, I'll simply use a custom image. If it works on 100 nodes, I'll try to run with more.

> One thing troubles me more: PV deletion.
> What should happen when a node is removed from the cluster? All its PVs can't be deleted - there is no provisioner that would do that. Is it just a documentation issue?

I think we need to document how to deal with this: if the admin is sure that the node is gone, they will have to force-delete orphan PVs from the node, including removal of finalizers.

The same problem occurs when the CSI driver handles communication with nodes internally - if it handles the issue. In PMEM-CSI we have an open bug around that, just that we erred towards allowing PV removal when uncertain, with the result that we sometimes don't delete a volume on a node.

> Or shall there be a cluster-level provisioner that deletes PVs from missing nodes? What if the node rejoins the cluster?

That's exactly the problem. Without further information, Kubernetes cannot know whether the node is truly gone. I'm hoping that the work on better node shutdown handling may lead to a better solution, but that'll take time.

pkg/controller/controller.go
pohly: I'll try to test that.

doc/provisioning.puml
pohly: You are right, I checked by watching PVC updates also for late binding. I've updated the diagram and also made it easier to read by styling the text.

When following the "unset selected node" arrow, it is obvious that the PV controller may see a PVC where it has already set the "provisioner" annotation earlier. That annotation is never reset, is it?

I changed the text to "if necessary, PV controller sets provisioner annotation" to reflect that it is possible and valid that the annotation already exists and then gets preserved.

I did that on *both* branches because expect that the code is the same.

pkg/controller/controller.go
pohly: This got me thinking. The current implementation does exponential backoff *per PVC*. This is almost never going to make a difference in practice because after the initial conflict, the PVC has a selected node and there's no need to do another update attempt for it. The only exception is when provisioning on that selected node fails and there has to be a global retry - that should be rare.

In this approach, we can't make this grow with cluster size either: we cannot assume that the CSI driver runs on all nodes. It might manage specialized resources that are only available on some nodes. With deployment handled by Kubernetes, the individual instances have no information which other instances exist.

I had considered making this independent of the PVC. The issue with that is the question of when to reduce the delay and how much. When the "winning" instance resets it to zero, then it is also likely to win for the next PVC until its space is exhausted and it stops trying to own a PVC (because of the capacity check). I preferred even spreading across the cluster. But perhaps filling up one node is also fine: if users want a certain policy for volume scheduling, they should use late binding and influence volume placement indirectly through that.

 I think I will give that a try...

pohly: It ended up working worse. The main problem was that the instance with the smallest backoff delay ends up grabbing more volumes for provisioning than what it has space for, so once it is the owner it fails for at least some of them. Those then are made available to other instances, but the same pattern just repeats.

It could be made to work if the provisioner could model remaining capacity considering how many volumes already were assigned, but that is driver specific and complex.

This is more of a problem under load (= many pending PVCs) and a full cluster (= capacity exhausted on some nodes), but still, when volumes are spread evenly purely based on statistics, it worked better.

So I'll stick with the current approach and perhaps tweak it a bit: when tracking the number of collisions per number of PVCs, it should be possible to scale the base delay.

pohly: > So I'll stick with the current approach and perhaps tweak it a bit: when tracking the number of collisions per number of PVCs, it should be possible to scale the base delay.

That has the same problem. When using a running average of the failure rate as indicator for "delay is too low", I saw increasing numbers of failed provisioning attempts because of full nodes.

TLDR: cluster admins will have to adjust the delay for the expected cluster size and number of volumes.

doc/provisioning.puml
pohly: I moved the diagram out of this PR into PR #532

pohly:
I've pushed updated commits that reflect the current status.

I included a temporary commit for using https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/100 but there is some issue with the replace statement.

@verult IMHO it makes sense to review anyway. Let's do a new sig-storage-lib-external-provisioner release with that pending PR and then I can use that cleanly here.

pohly:
> I think we need to document how to deal with this: if the admin is sure that the node is gone, they will have to force-delete orphan PVs from the node, including removal of finalizers.

I've added one commit with documentation for that.

> I included a temporary commit for using kubernetes-sigs/sig-storage-lib-external-provisioner#100 but there is some issue with the replace statement.

I took out that commit to make this PR ready for merging now.

The only remaining gap is support for storage capacity tracking in distributed provisioning mode. I'll work on that next, but the PR is already useful without it.

pohly:
Rebased because of conflicting changes in README.md.

pohly:
Provisioning of raw block volumes is failing on 1.18:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-provisioner/524/pull-kubernetes-csi-external-provisioner-1-18-on-kubernetes-1-18/1336730847977410560/artifacts/default/csi-hostpath-provisioner-0/csi-provisioner.log
```
I1209 18:10:15.140377       1 connection.go:182] GRPC call: /csi.v1.Controller/CreateVolume
I1209 18:10:15.140403       1 connection.go:183] GRPC request: {"accessibility_requirements":{"preferred":[{"segments":{"topology.hostpath.csi/node":"csi-prow-worker2"}}],"requisite":[{"segments":{"topology.hostpath.csi/node":"csi-prow-worker2"}}]},"capacity_range":{"required_bytes":1048576},"name":"pvc-c6dbfade-e0ed-45d1-a088-c62b26632f6e","volume_capabilities":[{"AccessType":{"Block":{}},"access_mode":{"mode":1}}]}
I1209 18:10:15.140369       1 event.go:282] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"volume-expand-3589", Name:"hostpath.csi.k8s.iojlfsx", UID:"c6dbfade-e0ed-45d1-a088-c62b26632f6e", APIVersion:"v1", ResourceVersion:"2112", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "volume-expand-3589-e2e-scmhh4w": rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

This seems unrelated to this PR.

pohly:
This might be one of those cases where testing under kind may leave behind bound loop device on the host where the kind cluster was created (https://github.com/kubernetes-sigs/kind/issues/1333).

verult:
Building off of @jsafrane 's idea of a cluster-level provisioner - for immediate binding, would it make sense to have a cluster-level provisioner set `selected-node`?

Pros of an additional cluster-level provisioner:
* PV deletion handling
* No race from all nodes
* Node selection is more predictable and can be deterministic.

Cons:
* An extra cluster-level deployment is required (although it may not need an accompanying driver?)

pohly:
> Building off of @jsafrane 's idea of a cluster-level provisioner - for immediate binding, would it make sense to have a cluster-level provisioner set selected-node?

There are various issues with that which make such an approach difficult as a generic component and (IMHO) not useful at the moment. If someone has a good idea how to implement such a component, then they can disable the immediate binding support - I intentionally added a flag for that case. 

> Pros of an additional cluster-level provisioner:
>
>    PV deletion handling

How would the cluster-level provisioner decide that it is okay to delete the PV? The only case where it seems safe is if the node object got removed from the cluster. Was that what you had in mind?

It does not help in situations where the node is temporarily unreachable, because then the cluster-level provisioner cannot be sure that the volume is gone.

>   No race from all nodes

True, but ...

>  Node selection is more predictable and can be deterministic.

... based on which information can the cluster-level provisioner choose a node? Without storage capacity tracking (currently in alpha) it has no information which node may be able to create the volume. It basically has to choose randomly (neither predictable nor deterministic) and when storage capacity gets exhausted on nodes, it becomes less and less likely that it will hit a node that still can create the volume.

> Cons:
> 
>    An extra cluster-level deployment is required (although it may not need an accompanying driver?)

It would need such an accompanying driver to retrieve capacity information, and then we are back at the point where CSI driver developers need to do something special and somehow establish a communication path between the node and cluster component of their driver.

msau42:
To handle the node deletion case, can we set PV.OwnerRef to the Node object?

pohly:
> To handle the node deletion case, can we set PV.OwnerRef to the Node object?

So we agree that "node deleted" = "all resources on it gone forever"? I thought so, but wasn't sure.

I suspect that this won't be enough, though. The finalizer will still be set and therefore garbage collection won't remove the PV. To handle the deletion case in the "deleted node" scenario, a central (de-)provisioner makes sense: it can remove the finalizer and delete the PV.

Let's merge this PR with the caveat documented and then I'll work on that deletion case next, okay?

pohly:
> scale testing: done on 100 nodes, maybe try something higher

I tried with 1000 nodes on Thursday, but pulling the custom external-provisioner image from my Docker Hub (my temporary location because I don't have an account for gcr) turned out be too slow and ran into a timeout. I have another change on Tuesday. If the PR is in the canary image by then, then we don't need further workarounds for slow pulling from Docker Hub - but let's not rush it just because of that.

I don't think we should hold the PR to complete that scale testing. I see immediate binding in this mode as something that anyone should even try to use at scale (large numbers of nodes, high number of PVCs) - just stick to late binding.

msau42:
> I suspect that this won't be enough, though. The finalizer will still be set and therefore garbage collection won't remove the PV. To handle the deletion case in the "deleted node" scenario, a central (de-)provisioner makes sense: it can remove the finalizer and delete the PV.

I don't think we need a central controller to remove the finalizer. We can let the normal workflow fully clean up the object. When someone is using local PVs and the node is deleted, they have to manually go in and delete and recreate the PVC object to rebind to a new disk. The purpose of the Node ownerref is so that the PV object will get cleaned up once the PVC is gone.

pohly:
> The purpose of the Node ownerref is so that the PV object will get cleaned up once the PVC is gone.

But who removes the finalizer on the PV object (the one that prevents removing a PV which still has a volume)? As long as that finalizer is set, the garbage collector is not going to remove the PV object.

msau42:
> who removes the finalizer on the PV object (the one that prevents removing a PV which still has a volume)?

The storage protection controller will remove the finalizer once the user deletes the PVC object

pohly:
I've tried this out by manually setting an ownerref on a PV that was created with distributed provisioning. Here the node and PVC still exist:
```
apiVersion: v1
items:
- apiVersion: v1
  kind: PersistentVolume
  metadata:
    annotations:
      pv.kubernetes.io/provisioned-by: pmem-csi.intel.com
    creationTimestamp: "2020-12-14T19:55:31Z"
    finalizers:
    - kubernetes.io/pv-protection
    ...
    name: pvc-3dd3edd2-b097-46c6-b586-eb664e5638c0
    ownerReferences:
    - apiVersion: v1
      kind: node
      name: pmem-csi-pmem-govm-worker2
      uid: 9ce9cd5b-79b6-41dd-9d33-4357f55a5d0e
    resourceVersion: "2300"
    selfLink: /api/v1/persistentvolumes/pvc-3dd3edd2-b097-46c6-b586-eb664e5638c0
    uid: 6314a6c8-9b14-464b-9a5a-1a7e94de3f5f
  spec:
    accessModes:
    - ReadWriteOnce
    capacity:
      storage: 2Mi
    claimRef:
      apiVersion: v1
      kind: PersistentVolumeClaim
      name: pmem-csi-pvc-ext4
      namespace: default
      resourceVersion: "1435"
      uid: 3dd3edd2-b097-46c6-b586-eb664e5638c0
    csi:
      driver: pmem-csi.intel.com
      fsType: ext4
      volumeAttributes:
        storage.kubernetes.io/csiProvisionerIdentity: 1607975727541-8081-pmem-csi.intel.com-pmem-csi-pmem-govm-worker2
      volumeHandle: pvc-3d-02f5d6560bed55abf8bb280db99ee462cf5f7629fe8bf472419e44ee
    nodeAffinity:
      required:
        nodeSelectorTerms:
        - matchExpressions:
          - key: pmem-csi.intel.com/node
            operator: In
            values:
            - pmem-csi-pmem-govm-worker2
    persistentVolumeReclaimPolicy: Delete
    storageClassName: pmem-csi-sc-ext4
    volumeMode: Filesystem
  status:
    phase: Bound
```

I then deleted the node. This only marked the PV for deletion:
```
8a9,10
>     deletionGracePeriodSeconds: 0
>     deletionTimestamp: "2020-12-14T20:02:47Z"
80c82
<     resourceVersion: "2300"
---
>     resourceVersion: "3057"
```

The finalizer still existed. Next I deleted the PVC.

Much to my surprise, the PV then did get removed. Why is that? Did the PV controller remove the `kubernetes.io/pv-protection` finalizer although the pv was still in `phase: Bound`?

`DeleteVolume` was never called for that volume. If someone manually deletes a PV while a CSI driver is down, will that then also leak the volume?

pohly:
I take back what I said about finalizers. That part apparently isn't the problem. I also need to check again how `kubernetes.io/pv-protection` works.

However, even if we set the ownerref cleanup is not automatic:

> When someone is using local PVs and the node is deleted, they have to manually go in and delete and recreate the PVC object to rebind to a new disk.

That means some manual cleanup is needed. Can we solve this better?

Suppose a central controller force-deletes the PV in this case while the PVC still exists. Will this trigger provisioning elsewhere?

Worth a try... unfortunately not. The PVC enters `phase: Lost` and stays there.

I'd still prefer to get this merged with the caveat about having to manual delete PVs (as documented). If this is not acceptable, then I can add the ownerref tomorrow.

pohly:
I rebased to resolve the `vendor/modules.txt`conflict.

pohly:
The rebase is a bit more complex, I'll finish it tomorrow.

pohly:
Regardless of the rebase it would be good to know whether I shall add the ownerref or defer that aspect.

msau42:
I think manual cleanup will always be in the picture when it comes to local volumes because the PVC needs to be deleted to get a new disk on a new node. There are some thoughts on a controller that workloads can opt into automatic PVC deletion behavior. https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/202

I think adding an ownerref could be done separately. But I don't think we should write a central controller to remove the finalizer out of band of the storage protection controller.

travisghansen:
I'm not sure how this PR is intended to fit into the bigger picture, but having a solution where once the workload has landed on a node all controller calls get invoked to the controller deployed on that node would be great (diagram 3 here: https://github.com/container-storage-interface/spec/blob/master/spec.md#architecture)

msau42:
> I'm not sure how this PR is intended to fit into the bigger picture, but having a solution where once the workload has landed on a node all controller calls get invoked to the controller deployed on that node would be great (diagram 3 here: https://github.com/container-storage-interface/spec/blob/master/spec.md#architecture)

This PR is only dealing wit the provisioner piece. No one is looking at the other controller pieces atm.

pkg/controller/controller.go
```diff
@@ -1121,6 +1228,230 @@ func (p *csiProvisioner) volumeHandleToId(handle string) string {
...
+		}
+
+		if sc == nil {
+			var err error
+			sc, err = p.scLister.Get(*claim.Spec.StorageClassName)
+			if err != nil {
+				return false, err
+			}
+		}
+		if sc.VolumeBindingMode == nil ||
```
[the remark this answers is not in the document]
pohly: It's the default, so I think yes. In practice this should never occur because the apiserver will set the default.

README.md
```diff
@@ -242,6 +252,93 @@ The external-provisioner optionally exposes an HTTP endpoint at address:port spe
...
+removes the "selected node" annotation and the process repeats. But
+this triggers events for the PVC and delays volume creation, in
+particular when storage is exhausted on most nodes. Therefore
+external-provisioner checks with `GetCapacity` *before* attempting to
+own a PVC whether the currently available capacity is sufficient for
+the volume. When it isn't, the PVC is ignored and some other instance
+can own it.
+
+The `--node-deployment-base-delay` parameter determines the initial
+wait period. It also sets the jitter, so in practice the delay will be
```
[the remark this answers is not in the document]
pohly: Changed.

pkg/controller/controller.go
```diff
@@ -1121,6 +1228,230 @@ func (p *csiProvisioner) volumeHandleToId(handle string) string {
...
+	if err != nil {
+		return false, err
+	}
+
+	// In practice, we expect exactly one entry here once a node
+	// has been selected. But we have to be prepared for more than
+	// one (=> check all, success if there is at least one) and
+	// none (no node selected => check once without topology).
+	topologies := []*csi.Topology{nil}
+	if result.req.AccessibilityRequirements != nil && len(result.req.AccessibilityRequirements.Requisite) > 0 {
```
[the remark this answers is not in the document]
pohly: Can you explain this improvement with an example? I'm not sure I understand.

pohly:
I've pushed an updated version. The newer Kubernetes helper packages moved the node selector match function. I updated the "Deleting local volumes after a node failure or removal" section with information about dealing with PVCs.

I did not add the ownerref because I'd like to get something merged before making further refinements.

pkg/controller/controller.go
```diff
@@ -201,7 +206,22 @@ type requiredCapabilities struct {
 	clone    bool
 }
 
-// CSIProvisioner struct
+// NodeDeployment contains additional parameters for running external-provisioner alongside a
+// CSI driver on one or more nodes in the cluster.
+type NodeDeployment struct {
+	NodeName         string
```
msau42, line 212: Can you add some comments on what each of these fields do?

pkg/controller/controller.go
```diff
@@ -1112,6 +1234,230 @@ func (p *csiProvisioner) volumeHandleToId(handle string) string {
...
+	}
+
+	var selectedNode string
+	if claim.Annotations != nil {
+		selectedNode = claim.Annotations[annSelectedNode]
+	}
+	switch selectedNode {
+	case "":
+		logger := klog.V(5)
+		if logger.Enabled() {
```
msau42, line 1255: Any reason why we do this method of checking log level?

pohly:
Rebased to resolve the conflict in pkg/controller/controller.go.

pkg/controller/controller.go
msau42: Do we need the polling period to be so frequent considering that the delays are in seconds?

pohly: Probably not. I've replaced it with 'base delay / 100', so now it scales together with the base delay and (for the default) is at 300ms.

I don't want to make it too large either because then one go routine would be blocked for potentially quite a while before it becomes available for some other PVC work item. With many provisioners running in parallel the actual smallest delay can be smaller than the maximum of 30 seconds.

pkg/controller/controller.go
pohly: Done.

pohly:
@msau42 : please take another look.

pkg/controller/controller.go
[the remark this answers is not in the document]
verult: So should it be sc.VolumeBindingMode != nil && and let becomeOwner() run if it's nil? It's probably not a big deal though if apiserver always sets it.

README.md
```diff
@@ -84,6 +84,18 @@ See the [storage capacity section](#capacity-support) below for details.
...
+
+* `--node-deployment`: Enables deploying the external-provisioner together with a CSI driver on nodes to manage node-local volumes. Off by default.
+
+* `--node-deployment-immediate-binding`: Determines whether immediate binding is supported when deployed on each node. Enabled by default, use `--node-deployment-immediate-binding=false` if not desired.
+
+* `--node-deployment-base-delay`: Determines how long the external-provisioner sleeps initially before trying to own a PVC with immediate binding. Defaults to 20 seconds.
+
+* `--node-deployment-max-delay`: Determines how long the external-provisioner sleeps at most before trying to own a PVC with immediate binding. Defaults to 60 seconds.
+
+* `--local-topology`: Instead of watching Node and CSINode objects, use only the topology provided by the CSI driver. Only valid in combination with `--node-deployment`. Disabled by default, but recommended for drivers which have a single topology key with different values for each node (i.e. local volumes).
```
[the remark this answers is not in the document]
msau42: Yeah I think that would simplify things initially.

README.md
```diff
@@ -84,6 +84,18 @@ See the [storage capacity section](#capacity-support) below for details.
 
 * `--capacity-for-immediate-binding <bool>`: Enables producing capacity information for storage classes with immediate binding. Not needed for the Kubernetes scheduler, maybe useful for other consumers or for debugging. Defaults to `false`.
 
+##### Distributed provisioning
+
+* `--node-deployment`: Enables deploying the external-provisioner together with a CSI driver on nodes to manage node-local volumes. Off by default.
```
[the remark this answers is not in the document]
msau42: I'm wondering if we can avoid flag skew issues like `--node-deployment=true` and `--capacity-controller-deployment-mode=central`.

Can change the capacity flag to be a boolean, and then determine which mode to use based on node deployment?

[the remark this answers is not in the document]
pohly: > Can change the capacity flag to be a boolean, and then determine which mode to use based on node deployment?

In retrospect that would be nicer, but it is a bit late to change the semantic of the `--capacity-controller-deployment-mode` that way - we would have to do a major update of external-provisioner for that because it is a breaking change.

pkg/controller/controller.go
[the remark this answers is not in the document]
pohly: I re-read this and remembered why I wrote it this way: `sc.VolumeBindingMode == nil` is unusual and therefore I didn't want the code to proceed with doing something that is potentially wrong.

README.md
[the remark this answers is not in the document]
msau42: It's an alpha feature, I think we can make breaking changes to it.

README.md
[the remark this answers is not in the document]
pohly: Done.

I kept the removal in a separate commit, in case that we want to bring it back.

pkg/controller/controller.go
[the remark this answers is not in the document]
verult: I don't think my suggest works actually. Originally I was thinking to only take the element in `AccessbilityRequirements.Requisite` that equals `NodeGetInfoResponse.AccessibleTopology`, but strictly speaking the `AccessibleTopology` of a volume may be different from the node.

verult:
/lgtm

LGTM on my end, leaving final approval to @msau42 / @jsafrane

README.md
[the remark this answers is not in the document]
pohly: So `--enable-capacity=true/false` instead of `--capacity-controller-deployment-mode=central` and the actual mode of operation (central vs. local) determined by `--node-deployment`?

Works for me, I just won't get to it today.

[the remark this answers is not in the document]
msau42: That sounds good to me! Do you think enable-capacity will always be a permanent flag, or would it fit a feature gate better?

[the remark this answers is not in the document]
pohly: I think a permanent flag is better. The feature causes additional work and might not be that relevant for some drivers (like those where all volumes are available everywhere), so being able to keep it turned off will remain useful.

README.md
```diff
@@ -242,6 +262,115 @@ The external-provisioner optionally exposes an HTTP endpoint at address:port spe
...
+then the risk of conflicts while setting the "selected node"
+annotation increases and the apiserver load will be higher.
+
+There is an exponential backoff per PVC which is used for unexpected
+problems. Normally, an owner for a PVC is chosen during the first
+attempt, so most PVCs will use the base delays. A maximum can be set
+with `--node-deployment-max-delay` anyway, to avoid very long delays
+when something went wrong repeatedly.
+
+During scale testing with 100 external-provisioner instances, a base
```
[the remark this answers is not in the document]
pohly: I got a chance to run on 1000 nodes. Distributed provisioning was almost as fast as central provisioning in terms of volumes/second for immediate binding with no overload of the apiserver - see report in https://github.com/kubernetes/perf-tests/pull/1676

PR #529, opened 2020-11-17
Update to snapshot v3client

xing-yang:
**What type of PR is this?**
/kind cleanup

This PR updates snapshot dependency to v3/client.

Fixes #

```release-note
Update to use snapshot v3/client
```


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/golang/mock v1.4.3
 	github.com/kubernetes-csi/csi-lib-utils v0.8.1
 	github.com/kubernetes-csi/csi-test/v3 v3.1.1
-	github.com/kubernetes-csi/external-snapshotter/client/v2 v2.2.0-rc3
+	github.com/kubernetes-csi/external-snapshotter/client/v3 v3.0.1-0.20201116082704-1e5a7ef7a73d
```
msau42, line 10: do you know why there's a hash afterwards instead of just "v3.0.1"?

xing-yang: Yes.  I do a go get with the snapshot commit `go get github.com/kubernetes-csi/external-snapshotter/client/v3@1e5a7ef7a73d691da0e6d350ba9613b05244d78f` and that will generate this hash automatically when it picks that commit.

xing-yang: We didn't really cut a 3.0.1 release for the client so it will not be just 3.0.1.  The latest official release is 3.0.0 for the client.

msau42: Can we import v3.0.0 instead?

xing-yang: That's probably fine too.

PR #535, opened 2020-12-03
Allow some PVC annotations to be passed to CreateVolume()

bswartz:
Add an option to the external-provisioner sidecar to pass a subset of the
PVC annotations to CreateVolume(). Using a vendor or administrator-specified
prefix, PVC annotations matching that prefix will be included alongside
storage class parameters.

This avoids conflicts between storage class parameters and PVC annotations
by only allowing annotations with a speficic prefix to be passed through.
Vendors must use vendor-specific prefixes to avoid conflicts between CSI
plugins.

The options defaults to disabled, and requires vendor support in the CSI
plugin to enable. Administrators may override this and disable it by
modifying the sidecar container to not pass the option.

**What type of PR is this?**

/kind feature

This allows per-PVC options to be passed through in a way that minimizes the problematic reasons we've rejected similar proposals in the past.

```release-note
TBD
```


xing-yang:
/hold

We did get lots of requests for this feature, so I think this will be very useful.  However, we discussed at the SIG-Storage meeting on 12/3 and decided that this will still break portability.  Putting this on hold for now.  Need further discussions to come up with a better solution.

PR #560, opened 2021-02-05
feat: Add migrated label to metrics

nearora-msft:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR is an extension to https://github.com/kubernetes-csi/csi-lib-utils/pull/77 to send "migrated" label in the metrics

Partially Addressed # kubernetes/kubernetes#98279

Fixes #

```release-note
Set the value of "migrated" field in the metrics to true or false to indicate if the call is a migration enabled feature or not
```


pkg/controller/controller.go
```diff
@@ -726,7 +726,7 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 	pvName := req.Name
 	provisionerCredentials := req.Secrets
 
-	createCtx, cancel := context.WithTimeout(ctx, p.timeout)
+	createCtx, cancel := context.WithTimeout(markAsMigrated(ctx), p.timeout)
```
Jiawei0227, line 729: I think one more thing you want to do is to redefine the metrics 
https://github.com/kubernetes-csi/external-provisioner/blob/aef465af4d2649eda1b8e63c9f8db9be257b27b5/cmd/csi-provisioner/csi-provisioner.go#L183

Here change the definition to 
```
metricsManager := metrics.NewCSIMetricsManagerWithOptions(
    "" /* driverName */,
    metrics.WithMigration(),
)
```

pkg/controller/controller.go
```diff
@@ -1714,3 +1714,7 @@ func checkFinalizer(obj metav1.Object, finalizer string) bool {
 	}
 	return false
 }
+
+func markAsMigrated(parent context.Context) context.Context {
+	return context.WithValue(parent, connection.AdditionalInfoKey, connection.AdditionalInfo{Migrated: metrics.LabelMigrated})
```
Jiawei0227, line 1719: Here we should record the migrated status for this operation. So I think it should be
```
connection.AdditionalInfo{Migrated: <migrated-val>}
```
<migrated-val> should be pass from the caller.

For example, in CreateVolume, it should get the <migrated-val> from `result.migratedVolume`. `result` is a response we get from prepareProvisionResult.

The purpose is to record whether this operation is generated based on CSI migration.

Jiawei0227:
Hey Neha, thanks a lot for your contribution!
/cc @msau42 
/ok-to-test

I had some comments. I am wondering can you also help to verify the change locally? You can 
```
make container
```
to build a local container and push it to dockerhub or wherever you are comfortable with. And then deploy it with a CSI driver into your cluster. Then if you trigger some operation like create volume, it should be able to generate some metrics. You can login to the node that deployed this node and do a `curl localhost:22021/metrics` to check the results. Let me know if you encounter any issues. If you dont know how to do that I am happy to test it as well once you addressed the comments.

Jiawei0227:
And can you also add a release-note for this?

pkg/controller/controller.go
```diff
@@ -507,7 +508,7 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		// the in-tree provisioner name in case of CSI migration scenarios. The annotation will be
 		// set to the CSI provisioner name by PV controller for migration scenarios
 		// so that external provisioner can correctly pick up the PVC pointing to an in-tree plugin
-		if sc.Provisioner == p.supportsMigrationFromInTreePluginName {
+		if hasMigrated(p, sc) {
```
Jiawei0227, line 511: Is there anywhere else that we call this function? If not, let's keep it as it is. I think from a readability perspective, it can be more straightforward

pkg/controller/controller.go
```diff
@@ -726,7 +727,7 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 	pvName := req.Name
 	provisionerCredentials := req.Secrets
 
-	createCtx, cancel := context.WithTimeout(ctx, p.timeout)
+	createCtx, cancel := context.WithTimeout(markAsMigrated(ctx, result.migratedVolume), p.timeout)
```
Jiawei0227, line 730: Can we separate this into two lines?
```
createCtx := markAsMigrated(ctx, result.migratedVolume)
createCtx, cancel = context.WithTimeout(createCtx, p.timeout)
```
Same as below.

pkg/controller/controller.go
```diff
@@ -1162,7 +1165,7 @@ func (p *csiProvisioner) Delete(ctx context.Context, volume *v1.PersistentVolume
 			klog.Warningf("failed to get storageclass: %s, proceeding to delete without secrets. %v", storageClassName, err)
 		}
 	}
-	deleteCtx, cancel := context.WithTimeout(ctx, p.timeout)
+	deleteCtx, cancel := context.WithTimeout(markAsMigrated(ctx, hasMigrated(p, storageClass)), p.timeout)
```
Jiawei0227, line 1168: Should we add a boolean here to record if the PV is actually an in-tree PV
https://github.com/kubernetes-csi/external-provisioner/blob/aef465af4d2649eda1b8e63c9f8db9be257b27b5/pkg/controller/controller.go#L1101

And we can just pass the bool here?

nearora-msft: Do you mean that we could set a bool value here:https://github.com/kubernetes-csi/external-provisioner/blob/aef465af4d2649eda1b8e63c9f8db9be257b27b5/pkg/controller/controller.go#L1101, something like 
https://github.com/kubernetes-csi/external-provisioner/blob/aef465af4d2649eda1b8e63c9f8db9be257b27b5/pkg/controller/controller.go#L517 and then there's no need to have the hasMigrated method?

Jiawei0227: Yep

nearora-msft:
> curl localhost:22021/metrics

Tested it with the azuredisk csi driver and I see these metrics in the node. The metrics are published with the migrated label indeed.

csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="5"} 24
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="10"} 24
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="15"} 24
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="25"} 24
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="50"} 24
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="120"} 24
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="300"} 24
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="600"} 24
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="+Inf"} 24
csi_sidecar_operations_seconds_sum{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false"} 1.9117699509999997
csi_sidecar_operations_seconds_count{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false"} 24
csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="0.1"} 1
csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="0.25"} 1
csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="0.5"} 1
csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="1"} 1
csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="2.5"} 1
csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="5"} 1
csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="10"} 1
csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="15"} 1
csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="25"} 1

nearora-msft:
> > curl localhost:22021/metrics
> 
> Tested it with the azuredisk csi driver and I see these metrics in the node. The metrics are published with the migrated label indeed.
> 
> csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="5"} 24
> csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="10"} 24
> csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="15"} 24
> csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="25"} 24
> csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="50"} 24
> csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="120"} 24
> csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="300"} 24
> csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="600"} 24
> csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false",le="+Inf"} 24
> csi_sidecar_operations_seconds_sum{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false"} 1.9117699509999997
> csi_sidecar_operations_seconds_count{driver_name="disk.csi.azure.com",grpc_status_code="Unknown",method_name="/csi.v1.Controller/DeleteVolume",migrated="false"} 24
> csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="0.1"} 1
> csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="0.25"} 1
> csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="0.5"} 1
> csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="1"} 1
> csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="2.5"} 1
> csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="5"} 1
> csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="10"} 1
> csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="15"} 1
> csi_sidecar_operations_seconds_bucket{driver_name="unknown-driver",grpc_status_code="OK",method_name="/csi.v1.Identity/GetPluginInfo",migrated="false",le="25"} 1

Adding metrics for CreateVolume too:
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="0.1"} 0
100csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="0.25"} 0
 16511   csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="0.5"} 0
 0 1651csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="1"} 0
1 csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="2.5"} 0
   0  csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="5"} 1
   0  2995k   csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="10"} 1
   0 --:csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="15"} 1
--:-- --:--:csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="25"} 1
-- --:--:-csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="50"} 1
- 3224kcsi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="120"} 1

csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="300"} 1
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="600"} 1
csi_sidecar_operations_seconds_bucket{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false",le="+Inf"} 1
csi_sidecar_operations_seconds_sum{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false"} 2.690683139
csi_sidecar_operations_seconds_count{driver_name="disk.csi.azure.com",grpc_status_code="OK",method_name="/csi.v1.Controller/CreateVolume",migrated="false"} 1

Jiawei0227:
/lgtm
Thanks for your contribution! I tested it locally and it works well for migration cases. 

/assign @msau42 @pohly 
I will let Patrick and Michelle to take a second look and approve.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -180,7 +180,7 @@ func main() {
 		klog.Fatalf("Error getting server version: %v", err)
 	}
 
-	metricsManager := metrics.NewCSIMetricsManager("" /* driverName */)
+	metricsManager := metrics.NewCSIMetricsManagerWithOptions("" /* driverName */, metrics.WithMigration())
```
msau42, line 183: Do we want to add the migration field for all drivers, or only those that are part of the migratable plugins? I believe this logic determines that: https://github.com/kubernetes-csi/external-provisioner/blob/37b54f06131df131cdcbe1084917075b5ea2a5a1/cmd/csi-provisioner/csi-provisioner.go#L332

pkg/controller/controller.go
```diff
@@ -726,7 +727,8 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 	pvName := req.Name
 	provisionerCredentials := req.Secrets
 
-	createCtx, cancel := context.WithTimeout(ctx, p.timeout)
+	createCtx := markAsMigrated(ctx, result.migratedVolume)
```
msau42, line 730: is it possible to add/extend a unit test to check that we set the ctx correctly?

cmd/csi-provisioner/csi-provisioner.go
Jiawei0227: +1, yeah
I think we should only add to those that are part of migratable.

nearora-msft: I looked into it. Not sure if it is possible to set the migration field only for the migratable drivers here. 
https://github.com/kubernetes-csi/external-provisioner/blob/37b54f06131df131cdcbe1084917075b5ea2a5a1/cmd/csi-provisioner/csi-provisioner.go#L332
This logic requires a provisionername and we are getting the provisionerName after creating the gRPC client here : https://github.com/kubernetes-csi/external-provisioner/blob/37b54f06131df131cdcbe1084917075b5ea2a5a1/cmd/csi-provisioner/csi-provisioner.go#L185

And to create the gRPC client we need  to create the metricsManager.

Please let me know if you think there's some other way to do it.

pkg/controller/controller.go
nearora-msft: I tried to capture the Context argument that is being passed to CreateVolume by using this logic:

controllerServerMock.EXPECT().CreateVolume(/*arguments*/).Do(func(ctx context.Context, req *csi.CreateVolumeRequest) {
						capturedContext = ctx
				})

and then validating the captured context using below logic:

			if (capturedContext == nil || capturedContext.Value(connection.AdditionalInfoKey) == nil) {
				t.Errorf("Got no migrated label in context, expected migrated label")
			}

But the context that is being captured by 'Do' doesn't contain the values for the AdditionalInfoKey or any other key that I tried to set and validate.

Please let me know if you can know of any other way to do it.

cmd/csi-provisioner/csi-provisioner.go
Jiawei0227: Hmm, I think there might be one possible way we can do.

We can create a new grpcClient with a new metricsManager after we get the provisionerName here:
https://github.com/kubernetes-csi/external-provisioner/blob/f937d0d4a1d88d866d95fad51148b686d9357cfc/cmd/csi-provisioner/csi-provisioner.go#L198

And move https://github.com/kubernetes-csi/external-provisioner/blob/f937d0d4a1d88d866d95fad51148b686d9357cfc/cmd/csi-provisioner/csi-provisioner.go#L329-L339 above so we know if this provisioner supports migration.

And dont forget to call grpcClient.Close() on the old connection to release resources.

nearora-msft: Addressed

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -180,27 +180,45 @@ func main() {
 		klog.Fatalf("Error getting server version: %v", err)
 	}
 
-	metricsManager := metrics.NewCSIMetricsManager("" /* driverName */)
+	metricsManager := metrics.NewCSIMetricsManagerWithOptions("" /* driverName */)
```
Jiawei0227, line 184: nit: change back to `metrics.NewCSIMetricsManager("" /* driverName */)`

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -326,15 +344,7 @@ func main() {
...
-	translator := csitrans.New()
-
-	supportsMigrationFromInTreePluginName := ""
-	if translator.IsMigratedCSIDriverByName(provisionerName) {
-		supportsMigrationFromInTreePluginName, err = translator.GetInTreeNameFromCSIName(provisionerName)
-		if err != nil {
-			klog.Fatalf("Failed to get InTree plugin name for migrated CSI plugin %s: %v", provisionerName, err)
-		}
-		klog.V(2).Infof("Supports migration from in-tree plugin: %s", supportsMigrationFromInTreePluginName)
+	if supportsMigrationFromInTreePluginName != "" {
```
Jiawei0227, line 347: nit: maybe move this above?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -180,27 +180,45 @@ func main() {
...
-	metricsManager := metrics.NewCSIMetricsManager("" /* driverName */)
+	metricsManager := metrics.NewCSIMetricsManagerWithOptions("" /* driverName */)
 
 	grpcClient, err := ctrl.Connect(*csiEndpoint, metricsManager)
 	if err != nil {
 		klog.Error(err.Error())
 		os.Exit(1)
 	}
 
-	err = ctrl.Probe(grpcClient, *operationTimeout)
```
Jiawei0227, line 191: We should probably keep it to make sure there is a connection? And add another this check inside the new grpc connection creation block

pkg/controller/controller.go
Jiawei0227: Can we add a unit test for the markAsMigrated function?

cmd/csi-provisioner/csi-provisioner.go
nearora-msft: Hmm we won't be able to move it above without moving a lot of code. This is because it used provisioner options which is declared above and provisioner option uses some of the variables that are being initialized above.

Jiawei0227: I see. It's cool to leave it here.

cmd/csi-provisioner/csi-provisioner.go
nearora-msft: I haven't removed it. Just moved it below so that it used the updated grpc connection. Do you mean that we should have it for both the old and the new connection?

Jiawei0227: Yeah, I think we should have both

nearora-msft: Ok. I am not sure(couldn't find much in the documentation) how the probe will be handled after this connection is closed. Do you think it will stop making the probe calls to that connection on its own once the connection is closed?

Jiawei0227: Oh, sorry my bad.

https://github.com/kubernetes-csi/external-provisioner/blob/f937d0d4a1d88d866d95fad51148b686d9357cfc/pkg/controller/controller.go#L276

It is probing forever instead of once. While I guess we can just have the new one

pkg/controller/controller.go
nearora-msft: Done

pkg/controller/controller.go
```diff
@@ -1714,3 +1719,7 @@ func checkFinalizer(obj metav1.Object, finalizer string) bool {
 	}
 	return false
 }
+
+func markAsMigrated(parent context.Context, hasMigrated bool) context.Context {
```
nearora-msft, line 1723: @Jiawei0227 It would probably be better to move this to the csi-lib-utils code because it wil be used across all the sidecars. What do you say?

Jiawei0227: Well. Yeah, I agree. But this function seems too small and I think it is okay to move around. So let's just keep it here maybe

Jiawei0227 (commented):
/lgtm
Thanks a lot! This looks good. I will let @msau42 to do a final approval.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -188,19 +188,37 @@ func main() {
...
+		supportsMigrationFromInTreePluginName, err = translator.GetInTreeNameFromCSIName(provisionerName)
+		if err != nil {
+			klog.Fatalf("Failed to get InTree plugin name for migrated CSI plugin %s: %v", provisionerName, err)
+		}
+		klog.V(2).Infof("Supports migration from in-tree plugin: %s", supportsMigrationFromInTreePluginName)
+
+		// Create a new connection with the metrics manager with migrated label
+		metricsManager = metrics.NewCSIMetricsManagerWithOptions("" /* driverName */, metrics.WithMigration())
+		migratedGrpcClient, err := ctrl.Connect(*csiEndpoint, metricsManager)
+		if err == nil {
```
msau42, line 210: should we exit if the new connect failed?

nearora-msft: Oh I felt it would be better to continue. I thought we shouldn't break even if are not able to create a connection with the added labels in the metrics manager and should continue with the previous connection. Do you think it would be better to exit instead?

msau42: Yes I think it's important that we get accurate metrics. I would not expect the second connect to fail in normal situations.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -188,19 +188,37 @@ func main() {
...
+	supportsMigrationFromInTreePluginName := ""
+	if translator.IsMigratedCSIDriverByName(provisionerName) {
+		supportsMigrationFromInTreePluginName, err = translator.GetInTreeNameFromCSIName(provisionerName)
+		if err != nil {
+			klog.Fatalf("Failed to get InTree plugin name for migrated CSI plugin %s: %v", provisionerName, err)
+		}
+		klog.V(2).Infof("Supports migration from in-tree plugin: %s", supportsMigrationFromInTreePluginName)
+
+		// Create a new connection with the metrics manager with migrated label
+		metricsManager = metrics.NewCSIMetricsManagerWithOptions("" /* driverName */, metrics.WithMigration())
```
Jiawei0227, line 208: nit: since we have got the driver name, we might just set it here directly: provisionerName
This can prevent lot of unknown-driver metrics

cmd/csi-provisioner/csi-provisioner.go
Jiawei0227: I took another look at the probeforever implementation. And find that this is a blocking function that will wait for the connection to be established. So I think we should have both the old probe and new probe. Otherwise `provisionerName, err := ctrl.GetDriverName(grpcClient, *operationTimeout)` could fail I guess? 

https://github.com/kubernetes-csi/external-provisioner/blob/f937d0d4a1d88d866d95fad51148b686d9357cfc/vendor/github.com/kubernetes-csi/csi-lib-utils/rpc/common.go#L126

Please also check the code and see if my understanding is correct

cmd/csi-provisioner/csi-provisioner.go
nearora-msft: Done

Jiawei0227 (commented):
/lgtm

msau42:
/approve

Thanks!

Jiawei0227:
/lgtm
Thanks!

PR #562, opened 2021-02-08
Check for volume in use before proceeding with delete volume

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

This PR checks for VolumesInUse node status before proceeding with deletion for volumes provisioned by csi drivers.

Fixes #

Couple of observation while making this fix:
1. In CanDelete call, VA object seems to be read from the informer. This needs some investigation as a separate issue if we should also check the API server, if the VA object is not found in the cache
2. Some of the tests cases of TestDelete are buggy. e.g. `fail - delete when volumeattachment exists but not attached to node` - The tests fail prematurely in trying to locate secretreference, instead of the expected VA object found errors. The test does not differentiate the errors and passes. I have added a TestCanDelete function which also includes the VA tests cases.
3. There is a dependent k/k patch which add support to report staged volumes in VolumesInUse in node status field.

```release-note
Check for volume in use before proceeding with delete volume
```


pkg/controller/controller.go
```diff
@@ -1175,20 +1183,26 @@ func (p *csiProvisioner) Delete(ctx context.Context, volume *v1.PersistentVolume
...
-	vaList, err := p.vaLister.List(labels.Everything())
-	if err != nil {
-		return fmt.Errorf("failed to list volumeattachments: %v", err)
-	}
+		for _, va := range vaList {
+			if va.Spec.Source.PersistentVolumeName != nil && *va.Spec.Source.PersistentVolumeName == volume.Name {
+				return fmt.Errorf("persistentvolume %s is still attached to node %s", volume.Name, va.Spec.NodeName)
+			}
+		}
+	} else if p.volumesInUseProtection {
```
jingxu97, line 1198: We only chceck if vaLister == nil?
if vaLister != nil and no VA matches volume name, we also need to check volumeInUse?

pkg/controller/controller.go
```diff
@@ -1362,6 +1376,38 @@ func (p *csiProvisioner) checkCapacity(ctx context.Context, claim *v1.Persistent
...
+
+	nodes, err := p.nodeLister.List(labels.Everything())
+	if err != nil {
+		return false, "", fmt.Errorf("unexpected error listing nodes from node informer: %v", err)
+	}
+
+	found, nodeName := isVolumeInUse(uniqueVolHandle, nodes)
+	if !found {
+		// Check nodes list from API server.
+		nodes, err := p.client.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
```
jingxu97, line 1397: I am trying to reason whether get list from API server is very necessary. 

The only case it is needed is volume was added into VolumeInUse list, but lister 's node has not yet updated? If lister has a delay, I agree checking api server will make it safer. But it does add some cost especially large scale cluster..

pkg/controller/controller.go
Jiawei0227: > if vaLister != nil and no VA matches volume name, we also need to check volumeInUse?

+1, for driver that is non-attachable, do we create VA? If not, then we need to check even if there is no VA, is the volume still in use

pkg/controller/controller.go
Jiawei0227: What's the frequency of node.volumeInUse get updated? I am trying to understand if it is possible that the node.volumeInUse get updated during the volume deletion.

Jiawei0227 (commented):
I think this change does not affect CSI migration.

pkg/controller/controller.go
saikat-royc: > I am trying to reason whether get list from API server is very necessary.
> 
> The only case it is needed is volume was added into VolumeInUse list, but lister 's node has not yet updated? If lister has a delay, I agree checking api server will make it safer. But it does add some cost especially large scale cluster..

@jingxu97 yes I added the API server check because the lister could potentially be out of date. And since its a delete call we should be conservative in making the decision to delete. However O(nodes) is still relatively moderate in the order of thousands for very large clusters. Thoughts?

saikat-royc: > What's the frequency of node.volumeInUse get updated? I am trying to understand if it is possible that the node.volumeInUse get updated during the volume deletion.

The node status is updated in a separate go thread. And the default periodicity is 10s (kubeCfg.NodeStatusUpdateFrequency.Duration). See [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L516)

Yes there are still very edge case scenario where delete can still sneak in.
Say volume is created and deleted very fast, before node status could be updated or say the kubelet is very slow in updating the node status/connectivity issues between kubelet and the API server. This fix acknowledges such known edge cases are not handled.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -93,10 +93,10 @@ var (
...
 	nodeDeploymentBaseDelay        = flag.Duration("node-deployment-base-delay", 20*time.Second, "Determines how long the external-provisioner sleeps initially before trying to own a PVC with immediate binding.")
 	nodeDeploymentMaxDelay         = flag.Duration("node-deployment-max-delay", 60*time.Second, "Determines how long the external-provisioner sleeps at most before trying to own a PVC with immediate binding.")
-
-	featureGates        map[string]bool
-	provisionController *controller.ProvisionController
-	version             = "unknown"
+	volumesInUseProtection         = flag.Bool("volumes-in-use-by-nodes-protection", false, "Enabling this flag will check for VolumesInUse node status, to identify volumes which are still attached/staged on a node. If the volume plugin supports Controller Publish/Unpublis calls, the flag will be ignored. The flag is honored for volume plugins which exclusively support Node Stage/Unstage calls. The flag should be enabled when deployed with kubelet that reports node staged volumes in node status.")
```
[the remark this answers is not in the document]
saikat-royc: Fixed

saikat-royc:
w.r.t https://github.com/kubernetes-csi/external-provisioner/pull/562#discussion_r575559182
"+1, for driver that is non-attachable, do we create VA? If not, then we need to check even if there is no VA, is the volume still in use"

@Jiawei0227  Can you elaborate your thoughts?
For non-attachable drivers we dont have a VA, hence the vaLister == nil. If driver is attachable (vaLister != nil), delete volume call would be blocked by checking VA objects.

Hence for vaLister == nil, we would proceed to volumesInUse check. However before proceeding with the check of node status, if you see I have created a flag to control this feature. i.e proceed with the check only if volumesInUseProtection == true. The flag needs to be enabled, when run on a k8s version with the changes of https://github.com/kubernetes/kubernetes/pull/98890 (to avoid unnecessary node list calls from API server).

Jiawei0227:
> Hence for vaLister == nil, we would proceed to volumesInUse check. However before proceeding with the check of node status, if you see I have created a flag to control this feature. i.e proceed with the check only if volumesInUseProtection == true. The flag needs to be enabled, when run on a k8s version with the changes of kubernetes/kubernetes#98890 (to avoid unnecessary node list calls from API server).

Yeah you are right. I forget if the driver is non-attachable the vaLister will be nil

pkg/controller/controller.go
mattcary: I think we should make it more explicit that no vaLister <=> no controller publish/unpublish. Right now it only seems documented at csi_provisioner.go:270 and it would be easy to break that assumption no knowing the implication here.

So maybe adding a flag in csiProvisioner would make that more clear?

mattcary: For the case when vaLister !=nil, ie we expect to have a VA, in https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_attacher.go#L359 it seems pretty clear if the VA is deleted then the volume is not in use... at least the csi controller detach has finished successfully, isn't that enough?

pkg/controller/controller.go
mattcary: It will be hard to get a definitive answer on this without actual scale testing.

We know the stale device problem is pretty bad and leads to degraded kubelet performance which is hard to diagnose. 

Since we know the extra apiserver calls only happen when volumes are being deleted, I think it will be not too hard to discover any scale problems that arise.

So I'd lean towards keeping the api server check in.

Similarly, for the case when the kubelet doesn't have #98890 we will have extra api server calls that aren't necessary. But it seems hard to avoid that without a lot more version plumbing? So I think what we have here is ok.

Jiawei0227: @bswartz this is also related to the scaling problem you mentioned. ;)

pkg/controller/controller.go
saikat-royc: @mattcary  yes if vaLister != nil (which implies volume supports VA creation), and the absence of a VA is a definite signal that controller unpublish was successful (and that indirectly implies unpublish/unstage succeeded)

mattcary: From other comments it's clear that vaLister indicating if the driver is attachable is a little subtle. Can you document that a bit more in the code to avoid future confusion?

saikat-royc: Ack

pkg/controller/controller.go
msau42: With generic ephemeral volumes coming in soon, it's very possible we'll have pvc create/delete as frequent as pods.

pkg/controller/controller.go
```diff
@@ -188,6 +188,11 @@ var (
 	}
 )
 
+const (
+	volumeHandleCSIPrefix = "kubernetes.io/csi/"
+	volumeHandleSep       = "^"
+)
```
gnufied, line 194: Why not define this with other constants at the top of the file?

pkg/controller/controller.go
gnufied: I think there is another issue here. The fix in - https://github.com/kubernetes/kubernetes/pull/98890 causes CSI volumes that don't have attach/detach to be reported in VolumeInUse status, but the kubelet can start mount operation without waiting for volume to be reported InUse in node. 

This has important side effects for this PR I think. This means that, if user deletes the pv/pvc before it is reported in node's status (i.e before 10s) then we are back to problematic case and then listing all node objects also will not help. 

I think what we might have to do is - https://github.com/kubernetes/kubernetes/blob/4f9317596ccb9fcb37bf183babf3e8640afcca9b/pkg/volume/util/operationexecutor/operation_generator.go#L1398 code we may to special case CSI volumes that support stage operation, so as mount operation can not start until volume is reported in-use by the kubelet. 

This will perhaps avoid about 95% chance of volume not being reported in node.Status

saikat-royc:
w.r.t [comment](https://github.com/kubernetes-csi/external-provisioner/pull/562#discussion_r587062806), yes this case that you pointed is a known issue, and acknowledged that this fix does not completely handle corner scenarios like this, where node status is stale and delete was initiated.

gnufied:
I think node status being stale because we are reading the informer and node is stale because we never updated node object are two different issues.

gnufied:
Another thing I was thinking is - could we watch for pod events in external-provisioner and keep a record of where a volume is being used and once pod is deleted, we set expiry on that record (so as it can be removed later) and when volume is deleted we could check this cache and see where volume was previously attached and only fetch those node objects directly from apiserver?

PR #567, opened 2021-02-19
Add TranslateInTreeStorageClassToCSI in Delete

### LINKED ISSUE #552: Forget TranslateInTreeStorageClassToCSI in Delete.
https://github.com/kubernetes-csi/external-provisioner/blob/637608cab7d37ddef0575fa2191ba22c57e43cc2/pkg/controller/controller.go#L1094

After writing the rbd-csi migration code, I tested it on my own cluster.
Provision successed, but Delete failed.
Then I realized that there is no TranslateInTreeStorageClassToCSI in Delete.
I will create a PR for this.

huchengze:
**What type of PR is this?**
> /kind bug

Fixes #552

```release-note
If the in-tree providers are depending on StorageClass for DeleteVolume, volumes could not get deleted after migrating from them. So translate in-tree StorageClass to CSI to get SecretReference in DeleteVolume call for CSI Migration scenario.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1145,6 +1145,14 @@ func (p *csiProvisioner) Delete(ctx context.Context, volume *v1.PersistentVolume
 	storageClassName := util.GetPersistentVolumeClass(volume)
 	if len(storageClassName) != 0 {
 		if storageClass, err := p.scLister.Get(storageClassName); err == nil {
+			if migratedVolume && storageClass.Provisioner == p.supportsMigrationFromInTreePluginName {
+				klog.V(2).Infof("translating storage class for in-tree plugin %s to CSI", storageClass.Provisioner)
+				storageClass, err = p.translator.TranslateInTreeStorageClassToCSI(p.supportsMigrationFromInTreePluginName, storageClass)
+				if err != nil {
+					return err
+				}
+			}
+
 			// Resolve provision secret credentials.
 			provisionerSecretRef, err := getSecretReference(provisionerSecretParams, storageClass.Parameters, volume.Name, &v1.PersistentVolumeClaim{
 				ObjectMeta: metav1.ObjectMeta{
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -62,9 +62,11 @@ func init() {
 }
 
 const (
-	timeout           = 10 * time.Second
-	driverName        = "test-driver"
-	driverTopologyKey = "test-driver-node"
+	timeout                = 10 * time.Second
+	driverName             = "test-driver"
+	driverTopologyKey      = "test-driver-node"
+	inTreePluginName       = "test-in-tree-plugin"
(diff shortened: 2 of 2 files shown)


huchengze:
@Jiawei0227 
I faile to rebase the last pr, so I create a new one to continue...

Jiawei0227:
Can you also squash the commit into one?

huchengze:
> Can you also squash the commit into one?

Yes, I have finished it.

pohly:
If this fixes a bug, then there should be a changelog entry for it, i.e. there has to be a release note comment in the PR description, not the `NONE` that's currently there.

huchengze:
@pohly PTAL

pohly:
As a user, that technical description doesn't tell me what the bug was and why an update is necessary. But it's better than nothing, so I'm okay with it.

Jiawei0227:
Maybe something like:
```
Bug fix: Translate in-tree StorageClass to CSI to get SecretReference in DeleteVolume call for CSI Migration scenario.
```

pohly:
Still too technical. What's the user-visible effect?

I'm looking for something like "volumes could not get deleted after migrating from in-tree providers" - if that's what the problem was. I really only get a very vague idea from the description of the implementation change.

You also don't need the "Bug fix:" prefix. The `/kind bug` takes care of that - well, would have if it hadn't been quoted :sweat_smile:  Let's fix that:

/kind bug

huchengze:
@pohly All information about this bug:
https://github.com/kubernetes-csi/external-provisioner/issues/552#issuecomment-783834139

Jiawei0227:
/lgtm

Thanks Patrick, I think it makes a lot of sense. Let's change the release-note as Patrick suggested. And I think we are good to merge this.

Jiawei0227:
/assign @msau42
For approval.

pkg/controller/controller.go
```diff
@@ -1145,6 +1145,14 @@ func (p *csiProvisioner) Delete(ctx context.Context, volume *v1.PersistentVolume
 	storageClassName := util.GetPersistentVolumeClass(volume)
 	if len(storageClassName) != 0 {
 		if storageClass, err := p.scLister.Get(storageClassName); err == nil {
+			if migratedVolume && storageClass.Provisioner == p.supportsMigrationFromInTreePluginName {
```
msau42, line 1148: Is it possible to add a unit test to https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller_test.go#L4310 to cover this scenario?

huchengze: I'm not familiar with gomock new, but I will try to do it.

Jiawei0227:
seems like there is a format issue, can you help to fix that?

huchengze:
> seems like there is a format issue, can you help to fix that?

Already done that. Seem like my IDE's version is a little low.

Jiawei0227:
/lgtm
@msau42 for approval

PR #583, opened 2021-03-11
storage capacity: object ownership

### LINKED ISSUE #465: storage capacity: make ownership of CSIStorageCapacity objects optional
Apparently some CSI drivers run outside of the cluster.

https://github.com/kubernetes-csi/external-provisioner/pull/450/files#r470300935

pohly:
**What type of PR is this?**
/kind api-change

This promotes "storage capacity tracking" to beta and changes its usages as agreed upon during KEP review for beta (optional ownership, labels used to identify objects managed by each instance).

Fixes #465 

Based on https://github.com/kubernetes-csi/external-provisioner/pull/579

```release-note
Storage capacity tracking becomes a beta feature. The parameters are different (`POD_NAMESPACE` -> `NAMESPACE`) and `CSIStorageCapacity` objects created via the alpha feature in previous releases must be deleted before rolling out the new release, because the way how each instance identifies the objects to manage has changed (from owner reference to labels).
```


pohly:
/hold

Let's merge this one last once all other changes for storage capacity tracking are in.

verult:
Just curious: how does a label selector increase API server CPU load?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -460,10 +461,17 @@ func main() {
 		}
 
 		// We only need objects from our own namespace. The normal factory would give
-		// us an informer for the entire cluster.
+		// us an informer for the entire cluster. We can further restrict the
+		// watch to just those objects with the right labels.
 		factoryForNamespace = informers.NewSharedInformerFactoryWithOptions(clientset,
 			ctrl.ResyncPeriodOfCsiNodeInformer,
 			informers.WithNamespace(namespace),
+			informers.WithTweakListOptions(func(lo *metav1.ListOptions) {
```
verult, line 469: I'm wondering how easy it is to detect if the informer doesn't fetch objects we expect, and whether we need to unit test this.

If informer logic is wrong, two things could happen:

* Missing items that are supposed to be included
  * May try to create an existing CSIStorageCapacity object. In the logic today this is fine, because nothing is done after object creation (besides logging / error reporting), and the creation error would give us a hint.
* Includes more things than it should
  * No failures, just more network traffic.

The latter case is hard to detect. So maybe a unit test is worthwhile (if it's not too much effort)?

pohly:
> Just curious: how does a label selector increase API server CPU load?

I had exactly the same concern and asked on #sig-scalability. The response was that spending more CPU time on server-side filtering can be assumed to be a net win when it decreases the number of objects that need to be transferred. That costs both network bandwidth *and* CPU time for conversion.

chrishenzie:
When you have a moment can you please address this merge conflict? We are planning to release this tomorrow to align with the Kubernetes 1.21 release.

cmd/csi-provisioner/csi-provisioner.go
verult: Actually since this is just an options configuration, the logic is simple enough, so it's probably low risk. Totally up to you if you want to unit test this.

verult:
> I had exactly the same concern and asked on #sig-scalability. The response was that spending more CPU time on server-side filtering can be assumed to be a net win when it decreases the number of objects that need to be transferred. That costs both network bandwidth and CPU time for conversion.

Since I'm not familiar with API server code, I was actually wondering exactly how computationally heavy server-side filtering is. What does it do under the hood to do this filtering?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -92,7 +92,7 @@ var (
...
 	capacityImmediateBinding = flag.Bool("capacity-for-immediate-binding", false, "Enables producing capacity information for storage classes with immediate binding. Not needed for the Kubernetes scheduler, maybe useful for other consumers or for debugging.")
 	capacityPollInterval     = flag.Duration("capacity-poll-interval", time.Minute, "How long the external-provisioner waits before checking for storage capacity changes.")
-	capacityOwnerrefLevel    = flag.Int("capacity-ownerref-level", 1, "The level indicates the number of objects that need to be traversed starting from the pod identified by the POD_NAME and POD_NAMESPACE environment variables to reach the owning object for CSIStorageCapacity objects: 0 for the pod itself, 1 for a StatefulSet, 2 for a Deployment, etc.")
+	capacityOwnerrefLevel    = flag.Int("capacity-ownerref-level", 1, "The level indicates the number of objects that need to be traversed starting from the pod identified by the POD_NAME and POD_NAMESPACE environment variables to reach the owning object for CSIStorageCapacity objects: -1 for no owner, 0 for the pod itself, 1 for a StatefulSet, 2 for a Deployment, etc.")
```
verult, line 96: 1 for a StatefulSet *or DaemonSet*

pkg/capacity/capacity.go
```diff
@@ -551,13 +561,17 @@ func (c *Controller) syncCapacity(ctx context.Context, item workItem) error {
 		// would race with receiving that object through the event handler. In the unlikely
 		// scenario that we end up creating two objects for the same work item, the second
 		// one will be recognized as duplicate and get deleted again once we receive it.
-	} else if capacity.Capacity.Value() == quantity.Value() {
+	} else if capacity.Capacity.Value() == quantity.Value() &&
+		(c.owner == nil || c.isOwnedByUs(capacity)) {
```
verult, line 565: Should this be `isManaged()`? If a driver running in binary mode manages an object which is unchanged, it should still skip right?

pkg/capacity/capacity.go
```diff
@@ -720,13 +733,20 @@ func (c *Controller) isObsolete(capacity *storagev1alpha1.CSIStorageCapacity) bo
...
+	for _, owner := range capacity.OwnerReferences {
 		if owner.Controller != nil && *owner.Controller && owner.UID == c.owner.UID {
 			return true
 		}
 	}
 	return false
 }
+
+// isManaged checks the labels to determine whether this capacity object is managed by
+// the controller instance.
```
verult, line 748: Since the informer is already filtered by labels in the next commit, and all callers of `isManaged()` fetch from the informer, is this mostly here as a safeguard? If so, nit: could you add a comment to clarify?

pkg/capacity/capacity.go
```diff
@@ -587,9 +601,8 @@ func (c *Controller) deleteCapacity(ctx context.Context, capacity *storagev1alph
 // ensures that it gets deleted if no longer needed. Foreign objects
```
verult, line 601: nit: update function name in comment

pkg/capacity/capacity_test.go
```diff
@@ -1217,7 +1264,8 @@ func fakeController(ctx context.Context, client *fakeclientset.Clientset, storag
 		driverName,
 		client,
 		queue,
-		defaultOwner,
+		&defaultOwner,
```
verult, line 1267: Should there be a test case for "managed by us but not owned by us"?

verult (changes requested):

cmd/csi-provisioner/csi-provisioner.go
pohly: I'd prefer to not test this. I suspect that it wouldn't be easy to test comprehensively and in the end is more about correct implementation of the informer and apiserver filtering than it is about correct configuration.

README.md
```diff
@@ -129,34 +129,56 @@ No | Irrelevant | No  | No  | `Requisite` and `Preferred` both nil
...
+They get created in the namespace identified with the `NAMESPACE`
+environment variable.
+
+Each external-provisioner instance manages exactly those objects with
+the labels that correspond to the instance.
+
+Optionally, all CSIStorageCapacity objects created by an instance of
+the external-provisioner can have an
+[owner](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents).
+This ensures that the objects get removed automatically when uninstalling
```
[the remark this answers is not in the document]
pohly: Added.

As you said, orphaned and incorrect objects cause no permanent damage for pods with normal PVCs (which includes generic ephemeral volumes), so it's not too bad. It's worse for CSI ephemeral volumes because then a pod might get scheduled onto a node where the driver then cannot provide the volume and the pod is stuck there (no rescheduling). I think that was what we chatted about, not the scheduler itself hanging.

README.md
```diff
@@ -129,34 +129,56 @@ No | Irrelevant | No  | No  | `Requisite` and `Preferred` both nil
...
+the CSI driver.  The owner is
+determine with the `POD_NAME/NAMESPACE` environment variables and
+the `--capacity-ownerref-level` parameter.
+
+If ownership is disabled, then the following command can be used to clean
```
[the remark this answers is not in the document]
pohly: Added.

cmd/csi-provisioner/csi-provisioner.go
pohly: added

pkg/capacity/capacity.go
pohly: This really is an ownerref check. It covers the scenario where a driver was deployed without ownerref and then gets redeployed with ownerref. The existing objects then need to be updated even when the capacity is the same, which is achieved here by falling through to the final else clause.

I added "same capacity %v *and correct owner*" to the log message to make this more clear.

pkg/capacity/capacity.go
pohly: Correct. I added a comment.

pkg/capacity/capacity.go
pohly: Fixed.

pkg/capacity/capacity_test.go
pohly: There is "reuse one capacity object, add owner" for that.

pohly: The inverse "reuse one capacity object, remove owner" doesn't work because the driver instance without owner configured doesn't know which owner references it is expected to remove. This shouldn't be needed in practice and the documentation mentions it.

pohly:
> Since I'm not familiar with API server code, I was actually wondering exactly how computationally heavy server-side filtering is. What does it do under the hood to do this filtering?

It has an in-memory representation of the object and then simply checks labels in the meta data. In that sense it's simply brute force. But the in-memory representation is needed anyway and then the check is still faster than the rest of the processing that would need to happen when sending the object. At least that is my understanding...

pohly:
I pushed an updated version. As before it is based on PR #586

pohly:
> When you have a moment can you please address this merge conflict? We are planning to release this tomorrow to align with the Kubernetes 1.21 release.

What about updating client-go in external-provisioner to 1.21.0? Kubernetes 1.21 must be released first, then I can update client-go and only then can we release external-provisioner.

Or we release external-provisioner with the current code. None of the differences seem relevant. I created the update for rc0 here: https://github.com/kubernetes-csi/external-provisioner/pull/599

chrishenzie:
@pohly I had planned on updating to 1.21 when it was released. Thank you for taking care of this.

verult:
> It has an in-memory representation of the object and then simply checks labels in the meta data. In that sense it's simply brute force. But the in-memory representation is needed anyway and then the check is still faster than the rest of the processing that would need to happen when sending the object. At least that is my understanding...

Yeah that makes sense. Some sort of index could maybe speed up processing but maybe that's overkill 🤷

README.md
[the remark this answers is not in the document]
verult: Ah yeah. Could you maybe mention this in the design doc somewhere (if it's not there already)?

pkg/capacity/capacity.go
```diff
@@ -562,14 +636,18 @@ func (c *Controller) syncCapacity(ctx context.Context, item workItem) error {
 		// would race with receiving that object through the event handler. In the unlikely
 		// scenario that we end up creating two objects for the same work item, the second
 		// one will be recognized as duplicate and get deleted again once we receive it.
-	} else if capacity.Capacity.Value() == quantity.Value() {
-		klog.V(5).Infof("Capacity Controller: no need to update %s for %+v, same capacity %v", capacity.Name, item, quantity)
+	} else if capacity.Capacity.Value() == quantity.Value() &&
+		(c.owner == nil || c.isOwnedByUs(capacity)) {
```
verult, line 640: I see, and `c.owner == nil` covers the case that a driver stays owner-less. Sgtm!

pkg/capacity/capacity_test.go
verult: I was actually referring to a test that captures the container-less driver case, where ownerref stays nil always. IIUC in all the tests so far assumes that capacity objects are eventually managed by a driver in containers (i.e. owner is always set to `defaultOwner`).

If you agree this test is necessary I would prefer that the test is added in this PR rather than delaying it after the release, because it is a use case we plan to support. Hopefully it's not a lot of work.

verult (changes requested):
One more potential change, everything else lgtm

README.md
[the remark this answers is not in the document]
pohly: It is covered by the generic ephemeral volume KEP.

pkg/capacity/capacity_test.go
pohly: You are right, that is a supported use case and wasn't covered. I've added two test cases for it which cover the relevant difference (create objects without owner reference, keep existing references).

verult:
/lgtm

/assign @msau42 
For approval

README.md
```diff
@@ -76,7 +76,7 @@ See the [storage capacity section](#capacity-support) below for details.
 
 * `--enable-capacity`: This enables producing CSIStorageCapacity objects with capacity information from the driver's GetCapacity call. The default is to not produce CSIStorageCapacity objects.
 
-* `--capacity-ownerref-level <levels>`: The level indicates the number of objects that need to be traversed starting from the pod identified by the POD_NAME and POD_NAMESPACE environment variables to reach the owning object for CSIStorageCapacity objects: 0 for the pod itself, 1 for a StatefulSet, 2 for a Deployment, etc. Defaults to `1` (= StatefulSet).
+* `--capacity-ownerref-level <levels>`: The level indicates the number of objects that need to be traversed starting from the pod identified by the POD_NAME and NAMESPACE environment variables to reach the owning object for CSIStorageCapacity objects: 0 for the pod itself, 1 for a StatefulSet, 2 for a Deployment, etc. Defaults to `1` (= StatefulSet). Ownership is optional and can be disabled with -1.
```
msau42, line 79: Update this to also include Daemonset

msau42 (commented):
/approve

just some nits

pohly:
> just some nits

Both fixed, please lgtm again.

/hold cancel

PR #584, opened 2021-03-11
storage capacity: maximum volume size

pohly:
**What type of PR is this?**
/kind api-change

The `MaximumVolumeSize` is useful to make the volume size check in kube-scheduler more accurate.

Depends on the new client-go bindings from Kubernetes 1.21.

```release-note
If the CSI driver reports `MaximumVolumeSize`, that value is also published in the `CSIStorageCapacity` objects. The v1beta1 storage API from Kubernetes >= 1.21 is required.
```


pohly:
This makes the storage capacity feature incompatible with Kubernetes < 1.21. I'm including that breaking change in the release note for PR #583.

We could also merge the two PRs.

We might want to hold this one until we can import the final code from Kubernetes v1.21.0?

/hold

pohly:
We could continue to support Kubernetes 1.19 and 1.20 by continuing to use the v1alpha1 API. `MaximumVolumeSize` was also added to that one. With my CSI driver developer hat on, I would prefer that because it makes it simpler for our users to try out the feature. But as upstream developer it is easier to drop usage of the v1alpha1 API as soon as possible

 I'm undecided...

verult:
Implementation LGTM.

I wonder if there were similar considerations when other in-tree CSI objects (VolumeAttachment/CSINode/CSIDriver) moved from v1alpha1 to v1beta1.

verult:
Synced with @msau42 offline. One option is to hold off moving to v1beta1 until most core Storage Capacity functionality is complete, cut a minor release with v1alpha1, and then immediately cut another release with a single PR changing it to v1beta1. This way, there's at least 1 release with nearly complete Storage Capacity implementation if users want to test on 1.19/1.20.

However the main downside is the CSI implementation team has to support more minor versions. The current policy is that we'll support minor versions (backporting bug fixes etc) for a year, so every minor release increases cherry-pick toil. This also sets a precedent which presumably will be the same for v1beta1->v1 and also other potential features relying on in-tree objects, multiplying the toil.

If we kept v1alpha1 this release and delayed moving to v1beta1 further than the option above, we'll end up with a bit less soak time for the v1beta1 machinery.

So personally I'm leaning towards moving to v1beta1 now. If testing the newer features on 1.19 & 1.20 is needed, one option is to implement all features and cut an RC release before changing to v1beta1, in which case the minor support policy doesn't apply.

pohly:
> If testing the newer features on 1.19 & 1.20 is needed, one option is to implement all features and cut an RC release before changing to v1beta1, in which case the minor support policy doesn't apply.

RC images don't get promoted, so using them for future testing on 1.19 or 1.20 might not work anymore once the image gets deleted from the staging registry.

The pros and cons of using v1beta1 match what I was considering, too. Let's move to v1beta1 right away.

pohly:
@verult : I rebased, please take another look.

Another open question (and the reason for the hold) was whether we should depend on unreleased k/k 1.21.0-beta.1 code. Any thoughts on that?

verult:
My 2 cents is depending on 1.21.0-beta.1 is OK so long as there's a tracking issue for updating it to 1.21.0 prior to the next provisioner release cut.

/lgtm
/assign @msau42 
for approval

pohly:
/hold cancel

https://github.com/orgs/kubernetes-csi/projects/32#card-57578868 tracks the update.

msau42:
/assign @Jiawei0227
 to also take a look at the metrics parts

pkg/capacity/capacity.go
```diff
@@ -536,16 +536,22 @@ func (c *Controller) syncCapacity(ctx context.Context, item workItem) error {
...
 				GenerateName:    "csisc-",
 				OwnerReferences: []metav1.OwnerReference{c.owner},
 			},
-			StorageClassName: item.storageClassName,
-			NodeTopology:     item.segment.GetLabelSelector(),
-			Capacity:         quantity,
+			StorageClassName:  item.storageClassName,
+			NodeTopology:      item.segment.GetLabelSelector(),
+			Capacity:          quantity,
+			MaximumVolumeSize: maximumVolumeSize,
```
msau42, line 554: Do we have a unit test for this?

verult:
Note the metric parts in this PR are the same as #579

pkg/capacity/capacity.go
pohly: We do now... :sweat_smile:

verult:
Test LGTM

/lgtm

PR #585, opened 2021-03-12
Refactor Controller Test

### LINKED ISSUE #376: Refactor controller_tests to use builder pattern
Provision and snapshot tests have a lot of logic built into the test case itself to construct input objects to the functions. For examples:
https://github.com/kubernetes-csi/external-provisioner/blob/16a2da32e052db2e2eaa31d51b70b202a9b39319/pkg/controller/controller_test.go#L1487
https://github.com/kubernetes-csi/external-provisioner/blob/16a2da32e052db2e2eaa31d51b70b202a9b39319/pkg/controller/controller_test.go#L2020

I think it would be cleaner to refactor these tests to use a builder pattern and construct the input object when defining the test cases.

Kartik494:
**What type of PR is this?**

/kind feature

Fixes #376

```release-note
NONE
```


pkg/controller/controller_test.go
```diff
@@ -3114,6 +3100,35 @@ func TestProvisionFromSnapshot(t *testing.T) {
 	}
 }
 
+func (s *testcase) notNilSnapshotStatusTest(notNilSnapshotStatusTest bool) {
```
Kartik494, line 3103: Hi @msau42 , i think added a constructor could be useful while i changing removing if-else statement and using a method instead. i am facing error when fetching `expectedpvSpec` property from  testcase struct.

pkg/controller/controller_test.go
```diff
@@ -1014,9 +1014,9 @@ func TestFSTypeProvision(t *testing.T) {
 		},
 	}
 
-	for k, tc := range testcases {
+	for k := range testcases {
 		t.Run(k, func(t *testing.T) {
-			runFSTypeProvisionTest(t, k, tc, requestedBytes, driverName, "" /* no migration */)
+			runFSTypeProvisionTest(t)
```
mauriciopoppe, line 1019: for every `k, testcase := range testcase` we're trying to test `testcase` however I don't see any reference to it here

pkg/controller/controller_test.go
```diff
@@ -2190,7 +2194,12 @@ func runFSTypeProvisionTest(t *testing.T, k string, tc provisioningFSTypeTestcas
 	}
 }
 
-func runProvisionTest(t *testing.T, tc provisioningTestcase, requestedBytes int64, provisionDriverName, supportsMigrationFromInTreePluginName string, testProvision bool) {
+func runProvisionTest(t *testing.T) *provisioningTestcase {
+	var k string
```
mauriciopoppe, line 2198: all of these local variables aren't set, if set they should be coming from the current `testcase`

pohly:
@Kartik494: this needs a rebase. @mauriciopoppe also had some review comments for which I don't see a response.

Kartik494:
@pohly Sorry for  the delay i was busy with some other work. I will update this PR and addressed the comments ASAP.

PR #589, opened 2021-03-19
remove duplicate informer.run

bai3shuo4:
**What type of PR is this?**
>
> /kind bug

When csi-provisoner start, it will call factory.Start(ctx.Done()). StorageClass Informer and topology Informer is created from factory. It will automatically run when calling factory.Start. Duplication of calling informer.run may cause some problems

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/capacity/capacity.go
+++ b/pkg/capacity/capacity.go
@@ -187,8 +187,6 @@ func NewCentralCapacityController(
 func (c *Controller) Run(ctx context.Context, threadiness int) {
 	klog.Info("Starting Capacity Controller")
 	defer c.queue.ShutDown()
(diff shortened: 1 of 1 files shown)


pohly:
Why was this closed?

At first glace, it looks like a reasonable enhancement. I'd like to know more about the "may cause some problems" part though - none of those seem to have occurred in the past, or were they simply not noticed?

bai3shuo4:
I signed it

bai3shuo4:
I recreate another PR, because of some author problems. We can discuss in that PR: https://github.com/kubernetes-csi/external-provisioner/pull/590

PR #590, opened 2021-03-19
remove duplicate informer.run()

bai3shuo4:
**What type of PR is this?**
>
> /kind api-change
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind bug

When csi-provisoner start, it will call factory.Start(ctx.Done()). StorageClass Informer and topology Informer is created from factory. It will automatically run when calling factory.Start. Duplication of calling informer.run may cause some problems

Fixes #

```release-note
Redundant CSIStorageCapacity objects might have been created because informers were not handled correctly.
```

### DIFF
--- a/pkg/capacity/capacity.go
+++ b/pkg/capacity/capacity.go
@@ -187,8 +187,6 @@ func NewCentralCapacityController(
 func (c *Controller) Run(ctx context.Context, threadiness int) {
 	klog.Info("Starting Capacity Controller")
 	defer c.queue.ShutDown()
-	go c.scInformer.Informer().Run(ctx.Done())
-	go c.topologyInformer.Run(ctx)
 
 	c.prepare(ctx)
 	for i := 0; i < threadiness; i++ {


bai3shuo4:
@pohly "May cause some problems": When I test the CSIStorageCapacity feature. After provisioner restarts for some times, It will create wrong number CSIStorageCapacity resources. For example, I use host ip as topology key and I have four node. It should have four CSIStorageCapacities in expect, but it shows 6 or more. After I check log. The problem happens at topology informer init. 

I0317 22:14:25.775357       1 capacity.go:302] Capacity Controller: topology changed: added [0xc000394ee0 = io.kubernetes.storage.mock/node: minikube-m03 0xc000394f40 = io.kubernetes.storage.mock/node: minikube-m04 0xc000394fa0 = io.kubernetes.storage.mock/node: minikube 0xc000395040 = io.kubernetes.storage.mock/node: minikube-m02], removed []
I0317 22:14:25.775245       1 capacity.go:302] Capacity Controller: topology changed: added [0xc000448980 = io.kubernetes.storage.mock/node: minikube-m03 0xc000448a00 = io.kubernetes.storage.mock/node: minikube-m04 0xc000448a60 = io.kubernetes.storage.mock/node: minikube 0xc000448b00 = io.kubernetes.storage.mock/node: minikube-m02], removed []

you can check log, the wrong happens because the topology sync add twice of per node with different address as hashmap key. I think it may cause from duplication informer.run because it may cause cache crash

pkg/capacity/capacity.go
```diff
@@ -187,8 +187,6 @@ func NewCentralCapacityController(
 func (c *Controller) Run(ctx context.Context, threadiness int) {
 	klog.Info("Starting Capacity Controller")
 	defer c.queue.ShutDown()
-	go c.scInformer.Informer().Run(ctx.Done())
-	go c.topologyInformer.Run(ctx)
```
pohly, line 191: There is at least one other redundant Run:
https://github.com/kubernetes-csi/external-provisioner/blob/bcf1a405b5f6a0fa2aadb80d08c9c2569c4b6912/pkg/controller/clone_controller.go#L82

On the other hand, this code does avoid Run if the informer is passed in:
https://github.com/kubernetes-csi/external-provisioner/blob/bcf1a405b5f6a0fa2aadb80d08c9c2569c4b6912/vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v6/controller/controller.go#L863-L873

What happens when `Run` is called too often? I.e., what goes wrong?

bai3shuo4: I think all these places should be fixed. I'm not sure how often it will cause problem. And In my case, the redundant CSIStorageCapacities is not showed every time. But I check the source code. In informer.run(), it will call controller.run(), and the comment said it 's an error to call Run more than once

// Run begins processing items, and will continue until a value is sent down stopCh.
// It's an error to call Run more than once.
// Run blocks; call via go.
func (c *controller) Run(stopCh <-chan struct{}) {
	defer utilruntime.HandleCrash()
	go func() {
		<-stopCh
		c.config.Queue.Close()
	}()
	r := NewReflector(
		c.config.ListerWatcher,
		c.config.ObjectType,
		c.config.Queue,
		c.config.FullResyncPeriod,
	)
	r.ShouldResync = c.config.ShouldResync
	r.clock = c.clock

	c.reflectorMutex.Lock()
	c.reflector = r
	c.reflectorMutex.Unlock()

	var wg wait.Group
	defer wg.Wait()

	wg.StartWithChannel(stopCh, r.Run)

	wait.Until(c.processLoop, time.Second, stopCh)
}

bai3shuo4: BTW, after I remove these two lines, the redundant CSIStorageCapacities case is not happened again. Now I try to run this test case for one or two days to check whether the case happens again

pohly: So the negative effect was that the capacity controller created redundant objects, probably because its informer callbacks were called too often. I wonder whether the duplicates should have been avoided in this case. This might be worth investigating, just to get full closure on the issue.

The fix itself here makes sense.

/kind bug

Can you add a change note that explains the bug? Instead of "NONE" in the description, put something like "Redundant CSIStorageCapacity objects might have been created because informers were not handled correctly."

pohly:
/hold

Let me first set up a pull test job which actually runs this code. Right now we don't have that.

pkg/capacity/capacity.go
bai3shuo4: OK, no problem

bai3shuo4:
> /hold
> 
> Let me first set up a pull test job which actually runs this code. Right now we don't have that.

I remove those two lines and run the controller for three days, the redundant never happens. I think the fix works

pohly:
The new "distributed-on-kubernetes-master" test job is probably not going to be very interesting yet.

The underlying problem is that this functionality needs to be covered by an E2E test, but external-provisioner has none of those. E2E testing only happens indirectly as part of driver testing, but the driver tests then have to be generic instead of focusing on external-provisioner.

The mock testing would be a good place - if we ran it for sidecar pre-merge tests, which we don't....

pohly:
/test pull-kubernetes-csi-external-provisioner-distributed-on-kubernetes-master

Now with an updated test job that uses csi-driver-host-path v1.6.0...

bai3shuo4:
The failed test cases look strange. Because this commit only changes some file with CSIStorageCapacity. But the failed cases are related to PV and SubPath something....

pohly:
Don't worry about those test failures. They also occur without this PR.

I'm currently working on setting up these Prow jobs to increase coverage of the new feature.

bai3shuo4:
> Don't worry about those test failures. They also occur without this PR.
> 
> I'm currently working on setting up these Prow jobs to increase coverage of the new feature.

Thanks!

pohly:
Let's merge. I filed an issue for the other controller: https://github.com/kubernetes-csi/external-provisioner/issues/595

/lgtm

pohly:
/assign @xing-yang 

For approval.

pohly:
@xing-yang can you approve?

Would it make sense to add an OWNERS file in pkg/capacity and make me an approver there? Or perhaps even in the top-level OWNERS?

PR #612, opened 2021-04-19
distributed provisioning: allowed topologies + immediate binding

### LINKED ISSUE #611: distributed provisioning: random failures for immediate binding + AllowedTopologies
Enabling topology tests in csi-driver-host-path led to random test failures for `External Storage [Driver: hostpath.csi.k8s.io] [Testpattern: Dynamic PV (immediate binding)] topology should provision a volume and schedule a pod with AllowedTopologies`.

The smoking gun is this event:
```
Apr 15 23:17:54.837: INFO: At 2021-04-15 23:13:00 +0000 UTC - event for pvc-plnph: {hostpath.csi.k8s.io_csi-hostpathplugin-dqmvv_a6875f50-3875-4081-bdbe-dc75bc843d42 } ProvisioningFailed: failed to provision volume with StorageClass "topology-7135-e2e-sc54m8h": error generating accessibility requirements: selected node '"csi-prow-worker2"' topology 'map[topology.hostpath.csi/node:csi-prow-worker2]' is not in allowed topologies: [map[topology.hostpath.csi/node:csi-prow-worker]]
```

What happened is that the test randomly picked one of the two worker nodes, `csi-prow-worker`, for AllowedTopologies in the test storage class, but then that field was ignored when the different provisioners determined who should provision the volume. The provisioner on `csi-prow-worker2` became the owner, but it then ran into the internal error about "generating accessibility requirements" quoted above. That causes the PVC to become stuck.

It's debatable whether such an internal error should cause the affected provisioner to retry or given up ownership of the PVC. I think such errors should not occur and therefore I wouldn't introduce special behavior for them, i.e. keep retrying.

The fix will be to check AllowedTopologies and ignore PVCs with incompatible topology.

pohly:
**What type of PR is this?**
/kind bug

When using immediate binding with allowed topologies set in the storage class together with distributed provisioning, node instances must ignore PVCs that do not fit their node.

Fixes #611 

Includes some drive-by test enhancements.

```release-note
When using immediate binding, allowed topologies and distributed provisioning, PVCs may have been assigned to nodes which did not fit the allowed topologies, leading to permanent `error generating accessibility requirements` failures.
```

/cc @chrishenzie

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1293,6 +1293,32 @@ func (p *csiProvisioner) checkNode(ctx context.Context, claim *v1.PersistentVolu
 			return false, nil
 		}
 
+		// If the storage class has AllowedTopologies set, then
+		// it must match our own. We can find out by trying to
+		// create accessibility requirements.  If that fails,
+		// we should not become the owner.
+		if len(sc.AllowedTopologies) > 0 {
+			node, err := p.nodeLister.Get(p.nodeDeployment.NodeName)
+			if err != nil {
+				return false, err
+			}
+			if _, err := GenerateAccessibilityRequirements(
+				p.client,
+				p.driverName,
+				claim.Name,
+				sc.AllowedTopologies,
+				node,
+				p.strictTopology,
+				p.immediateTopology,
+				p.csiNodeLister,
+				p.nodeLister); err != nil {
+				if logger.Enabled() {
+					logger.Infof("%s: ignoring PVC %s/%s, allowed topologies is not compatible: %v", caller, claim.Namespace, claim.Name, err)
+				}
+				return false, nil
+			}
+		}
+
 		// Try to select the current node if there is a chance of it
 		// being created there, i.e. there is currently enough free space (checked in becomeOwner).
 		//
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -817,12 +817,13 @@ type provisioningTestcase struct {
 	createVolumeError  error
 	expectErr          bool
 	expectState        controller.ProvisioningState
-	expectCreateVolDo  interface{}
+	expectCreateVolDo  func(t *testing.T, ctx context.Context, req *csi.CreateVolumeRequest)
 	withExtraMetadata  bool
 	skipCreateVolume   bool
 	deploymentNode     string // fake distributed provisioning with this node as host
 	immediateBinding   bool   // enable immediate binding support for distributed provisioning
 	expectSelectedNode string // a specific selected-node of the PVC in the apiserver after the test, same as before if empty
+	expectNoProvision  bool   // if true, then ShouldProvision should return false
 }
 
 type provisioningFSTypeTestcase struct {
@@ -833,7 +834,6 @@ type provisioningFSTypeTestcase struct {
 	createVolumeError error
 	expectErr         bool
 	expectState       controller.ProvisioningState
-	expectCreateVolDo interface{}
 
 	skipDefaultFSType bool
 }
@@ -1020,7 +1020,7 @@ func TestFSTypeProvision(t *testing.T) {
 	}
 }
 
-func TestProvision(t *testing.T) {
+func provisionTestcases() (int64, map[string]provisioningTestcase) {
 	var requestedBytes int64 = 100
 	deletePolicy := v1.PersistentVolumeReclaimDelete
 	immediateBinding := storagev1.VolumeBindingImmediate
@@ -1035,7 +1035,7 @@ func TestProvision(t *testing.T) {
 			Name: "bar",
 		},
 	}
-	testcases := map[string]provisioningTestcase{
+	return requestedBytes, map[string]provisioningTestcase{
 		"normal provision": {
 			volOpts: controller.ProvisionOptions{
 				StorageClass: &storagev1.StorageClass{
@@ -1091,7 +1091,7 @@ func TestProvision(t *testing.T) {
 					},
 				},
 			},
-			expectCreateVolDo: func(ctx context.Context, req *csi.CreateVolumeRequest) {
+			expectCreateVolDo: func(t *testing.T, ctx context.Context, req *csi.CreateVolumeRequest) {
 				pvc := createFakePVC(requestedBytes)
 				expectedParams := map[string]string{
 					pvcNameKey:      pvc.GetName(),
@@ -1146,7 +1146,7 @@ func TestProvision(t *testing.T) {
 					},
 				},
 			},
-			expectCreateVolDo: func(ctx context.Context, req *csi.CreateVolumeRequest) {
+			expectCreateVolDo: func(t *testing.T, ctx context.Context, req *csi.CreateVolumeRequest) {
 				if len(req.Parameters) != 0 {
 					t.Errorf("Parameters should have been stripped")
 				}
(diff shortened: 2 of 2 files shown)


pohly:
The whole `pull-kubernetes-csi-external-provisioner-distributed-on-kubernetes-master ` job can be ignored. It still uses the v1.6.0 release of the hostpath driver. I just wasn't sure anymore and thought I give it a try...

pkg/controller/controller.go
```diff
@@ -1293,6 +1293,32 @@ func (p *csiProvisioner) checkNode(ctx context.Context, claim *v1.PersistentVolu
 			return false, nil
 		}
 
+		// If the storage class as AllowedTopologies set, then
```
chrishenzie, line 1296: typo: has

pohly: Fixed.

pkg/controller/controller.go
```diff
@@ -1293,6 +1293,32 @@ func (p *csiProvisioner) checkNode(ctx context.Context, claim *v1.PersistentVolu
 			return false, nil
 		}
 
+		// If the storage class has AllowedTopologies set, then
+		// it must match our own. We can find out by trying to
+		// create accessibility requirements.  If that fails,
+		// we should become the owner.
```
verult, line 1299: we should *not* become the owner

pkg/controller/controller.go
```diff
@@ -1293,6 +1293,32 @@ func (p *csiProvisioner) checkNode(ctx context.Context, claim *v1.PersistentVolu
...
+			if err != nil {
+				return false, err
+			}
+			if _, err := GenerateAccessibilityRequirements(
+				p.client,
+				p.driverName,
+				claim.Name,
+				sc.AllowedTopologies,
+				node,
+				p.strictTopology,
```
verult, line 1311: If `strictTopology` is false, does `GenerateAccessibilityRequirements()` return an error if `node` isn't part of `AllowedTopologies`?

pohly: Yes, it should. The "strict topology" case is just a simpler version of that existing, older check.

pkg/controller/controller_test.go
```diff
@@ -2149,64 +2268,91 @@ func runProvisionTest(t *testing.T, k string, tc provisioningTestcase, requested
...
 		}
+	} else {
+		provision := csiProvisioner.(controller.Qualifier).ShouldProvision(context.Background(), tc.volOpts.PVC)
+		if provision != !tc.expectNoProvision {
+			t.Fatalf("expect ShouldProvision result %v, got %v", !tc.expectNoProvision, provision)
+		}
 	}
 
-	if expectSelectedNode != "" {
+	if tc.volOpts.PVC != nil {
```
verult, line 2367: Could you quickly explain why this check was added? (just for my understanding, not necessarily as a code comment)

pkg/controller/controller_test.go
```diff
@@ -1960,13 +1962,91 @@ func TestProvision(t *testing.T) {
...
+						},
+					},
+				},
+				PVName: "test-name",
+				PVC:    createFakePVC(requestedBytes),
+			},
+			expectErr:          true,
+			expectState:        controller.ProvisioningNoChange,
+			skipCreateVolume:   true,
+			expectNoProvision:  true, // not owner and will not change that either
```
verult, line 2054: Prior to the fix, the test will fail at `expectNoProvision`, while `expectErr`, `expectState`, and `skipCreateVolume` will pass, right?

pkg/controller/controller_test.go
pohly: Just for the sake of completeness. All other pointers were also checked for non-nil.

pkg/controller/controller_test.go
pohly: It would have passed `expectNoProvision`, but for the wrong reason (not owner yet). Without the fix, it fails `expectSelectedNode` because the provisioner would set the current node to itself. The correct behavior is to not provision and not set the node.

verult: Gotcha. The behavior of selected node is a little non-obvious to me, and I had trouble making sense of what behavior should be expected from the test case parameters. Given enough time I should be able to figure it out but I couldn't quite grasp it just by scanning the test case.

Just a nit - not sure what could improve it but maybe a comment explaining the scenario could help?

pkg/controller/controller.go
verult: Sg. I haven't been able to find the old logic but since we have test coverage this should be great. I'll find the logic eventually 😛

verult:
/lgtm
/hold
(in case you want to address any nits)

pkg/controller/controller_test.go
verult: Maybe consider adding a comment that selecting a node == becoming owner

pohly: Added a comment.

pohly:
> /lgtm
> /hold
> (in case you want to address any nits)

I do, and did :grin: 

Please re-approve.

verult:
/assign @msau42 
for approval

pkg/controller/controller.go
```diff
@@ -1293,6 +1293,32 @@ func (p *csiProvisioner) checkNode(ctx context.Context, claim *v1.PersistentVolu
...
+				p.immediateTopology,
+				p.csiNodeLister,
+				p.nodeLister); err != nil {
+				if logger.Enabled() {
+					logger.Infof("%s: ignoring PVC %s/%s, allowed topologies is not compatible: %v", caller, claim.Namespace, claim.Name, err)
```
msau42, line 1316: This call can error for other reasons besides incompatible allowedTopologies. Will this correctly recover in that case? Do we need a more specific error code to look for?

pkg/controller/controller.go
verult: I actually still couldn't find the logic for this when `strictTopology` is false. If you happen to know off-hand could you point me to it?

verult: The logic must exist somewhere because if it doesn't, `GenerateAccessibilityRequirements` would return a topology and no error. `checkNode()` then proceeds and tries to `becomeOwner()`, in which case either it succeeds and selected node is updated, or it fails and returns an error, and the caller, `ShouldProvision()` proceeds to return `true`. In both cases the `"distributed immediate, allowed topologies not okay` test case should've failed

pkg/controller/controller.go
pohly: Which failures? When deployed on a node, fake CSINode and Node listers are used (https://github.com/kubernetes-csi/external-provisioner/blob/213cd3d4e56fb439b06922ecf85d230a99d4e70d/cmd/csi-provisioner/csi-provisioner.go#L304-L340), so all errors related to access to those can be ruled out. Errors related to incorrect or incomplete content of those objects also cannot occur.

I think all that remains is an unsuitable selected node, and therefore we don't need a more detailed error.

After having described it like this, the code could also be replaced with a direct comparison of the selected node. I'm undecided whether that is better: if `GenerateAccessibilityRequirements` then fails after the node became owner (against all expectations), then the volume is stuck. More code would be needed to handle that, like detecting this particular problem earlier and/or recovering earlier.

Given that immediate binding isn't recommended when using distributed provisioning, I think I prefer to merge this simpler solution even if it doesn't cover some edge cases.

msau42: Thanks for explaining. It wasn't obvious to me from a quick review of the code that this is the case. I think from a general readability and maintenance standpoint, it's better to code more defensively in case the function changes in the future and we can no longer make this assumption.

pohly: I can add a specific error and check for it here.

msau42: A separate error code, or a new condition, or perhaps a more targeted function like you mentioned would all work. Out of all those, I like the separate method approach the most because there's no confusion about what parts of the function are/are not relevant depending on the codepath.

Anyway, I think we can leave this to a future refactoring. For now, this works with the current assumptions.

/approve

PR #615, opened 2021-04-21
CreateVolume Metadata:  Provide StorageClass name

theduke:
**What type of PR is this?**

/kind feature

Injects the `StorageClass` name into the metadata provided to `CreateVolume` if metadata passing is enabled with `--extra-create-metadata`.

This allows a single CSI driver to handle multiple storage classes.

```release-note

`CreateVolume` requests to the CSI driver now contain an additional parameter specifying the name of the storage class.
The parameter key is `csi.storage.k8s.io/sc/name`.

Only applicable if the metadata feature is enabled with `--extra-create-metadata`.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -113,6 +113,7 @@ const (
 	pvcNameKey      = "csi.storage.k8s.io/pvc/name"
 	pvcNamespaceKey = "csi.storage.k8s.io/pvc/namespace"
 	pvNameKey       = "csi.storage.k8s.io/pv/name"
+	scNameKey       = "csi.storage.k8s.io/sc/name"
 
 	// Defines parameters for ExponentialBackoff used for executing
 	// CSI CreateVolume API call, it gives approx 4 minutes for the CSI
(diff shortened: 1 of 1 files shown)


theduke:
I signed it.

theduke:
@davidz627 @lpabon could I get a review on this?

It's a tiny patch, but I need it for one of my CSI drivers.

davidz627:
I no longer work on this project. @msau42 might be able to find the right person to look at this

RaunakShah:
/ok-to-test

I think this change would also need some modifications to the unit test

humblec:
@theduke can you please explain the use case in bit more detail here? 

```
This allows a single CSI driver to handle multiple storage classes.
```

Because the abstraction of storage class and its parameters in the PV request were discussed heavily in the past and it was avoided deliberately.  The entire reason for the sidecar to deal with SC and its parameters and then form the createVolume request was based on the same. It would be appreciated if you can share some more details about the use case.

theduke:
The concrete use case is a LVM driver that handles multiple storage classes , up to 10 per node in my setup.

All additional information that the driver needs is the name of the storage class.

Running 10 `external-provisioner` sidecars just to handle the different storage classes is really quite bothersome and needless overhead.

msau42:
Can you explain in more detail what you need from the StorageClass name? Most CSI drivers today are able to support multiple storageclasses without knowing the name.

PR #617, opened 2021-04-25
node topology worker should run

bai3shuo4:
**What type of PR is this?**

/kind bug

bugfix for commit before, if we directly remove topology informer run, the topology worker will not run and do not sync. fix this PR https://github.com/kubernetes-csi/external-provisioner/pull/590
Fixes #

```release-note
Fix capacity information updates when topology changes. Only affected central deployment and network attached storage, not deployment on each node. This broke in v2.2.0 as part of a bug fix for capacity informer handling.
```

### DIFF
--- a/pkg/capacity/topology/mock.go
+++ b/pkg/capacity/topology/mock.go
@@ -51,7 +51,7 @@ func (mt *Mock) List() []*Segment {
 	return mt.segments
 }
 
-func (mt *Mock) Run(ctx context.Context) {
+func (mt *Mock) RunWorker(ctx context.Context) {
 }
 
 func (mt *Mock) HasSynced() bool {
--- a/pkg/capacity/topology/nodes.go
+++ b/pkg/capacity/topology/nodes.go
@@ -201,14 +201,12 @@ func (nt *nodeTopology) List() []*Segment {
 	return segments
 }
 
-func (nt *nodeTopology) Run(ctx context.Context) {
-	go nt.nodeInformer.Informer().Run(ctx.Done())
-	go nt.csiNodeInformer.Informer().Run(ctx.Done())
+func (nt *nodeTopology) RunWorker(ctx context.Context) {
 	go nt.runWorker(ctx)
 
-	klog.Info("Started node topology informer")
+	klog.Info("Started node topology worker")
 	<-ctx.Done()
-	klog.Info("Shutting node topology informer")
+	klog.Info("Shutting node topology worker")
 }
 
 func (nt *nodeTopology) HasSynced() bool {
--- a/pkg/capacity/topology/topology.go
+++ b/pkg/capacity/topology/topology.go
@@ -127,11 +127,11 @@ type Informer interface {
 	// List returns all known segments, in no particular order.
 	List() []*Segment
 
-	// Run starts watching for changes.
-	Run(ctx context.Context)
-
 	// HasSynced returns true once all segments have been found.
 	HasSynced() bool
+
+	// RunWorker starts a worker to process queue.
+	RunWorker(ctx context.Context)
 }
 
 type Callback func(added []*Segment, removed []*Segment)
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -438,6 +438,7 @@ func main() {
 			klog.Infof("producing CSIStorageCapacity objects with fixed topology segment %s", segment)
 			topologyInformer = topology.NewFixedNodeTopology(&segment)
 		}
+		go topologyInformer.RunWorker(context.Background())
 
 		managedByID := "external-provisioner"
 		if *enableNodeDeployment {


bai3shuo4:
@pohly sorry, this PR should be fix further. Do you remember this PR: https://github.com/kubernetes-csi/external-provisioner/pull/590. It will cause problem when topology change

pohly (changes requested):
Please squash your commits and reference the commit where this broke.

pkg/capacity/capacity.go
```diff
@@ -231,6 +231,7 @@ func (c *Controller) Run(ctx context.Context, threadiness int) {
 	klog.Info("Starting Capacity Controller")
 	defer c.queue.ShutDown()
 
+	go c.topologyInformer.Run(ctx)
```
pohly, line 234: This looks like the wrong place to put this. To be consistent with how the other informers are handled, this should be in the location where the informer gets created (i.e. in csi-provisioner.go), not where it is used.

bai3shuo4: This is the place it was placed: https://github.com/kubernetes-csi/external-provisioner/pull/590/commits/ff439eeba754c8dcbe4a4a1b9ed59ace51291133

bai3shuo4: According to this commit, if this line was removed directly, the node topology worker will not run. If you change the node label or csinode, nothing will happen. It will cause problem

pohly:
We have unit tests for "topology changes", but that doesn't help here because this is an issue about combining the various pieces into the final binary. An E2E test would be needed which:
- deploys a CSI driver on multiple nodes, with capacity tracking
- changes the node topology while the test runs

Right now we don't have a driver for testing this.

@bai3shuo4: You are using this with network attached storage, i.e. you have a central external-provisioner instance, right? Distributed provisioning uses a fixed topology, so it shouldn't matter there whether the informers run.

pkg/capacity/capacity.go
pohly: I agree that removal of `go c.topologyInformer.Run(ctx)` in that commit introduced a bug. The question just is where that `Run()` call should go instead. I propose `cmd/csi-provisioner/csi-provisioner.go`.

bai3shuo4:
> We have unit tests for "topology changes", but that doesn't help here because this is an issue about combining the various pieces into the final binary. An E2E test would be needed which:
> 
> * deploys a CSI driver on multiple nodes, with capacity tracking
> * changes the node topology while the test runs
> 
> Right now we don't have a driver for testing this.
> 
> @bai3shuo4: You are using this with network attached storage, i.e. you have a central external-provisioner instance, right? Distributed provisioning uses a fixed topology, so it shouldn't matter there whether the informers run.

Yes, I use central one and network storage

pkg/capacity/capacity.go
bai3shuo4: In fact, `c.topologyInformer.Run(ctx)` is not an informer. It is only a topology queue worker. the node informer and csinode informer are started by `factory.Start()` in `cmd/csi-provisioner/csi-provisioner.go`. Moreover, only capacity feature uses this worker and it should belong to capacity controller right?

pohly:
> Yes, I use central one and network storage

How important is storage capacity tracking in such a setup? Are there some limitations which nodes have access to which storage?

pkg/capacity/capacity.go
bai3shuo4: Maybe I can change this method name to avoid ambiguity like changing `Run` to `Run worker`

bai3shuo4:
> > Yes, I use central one and network storage
> 
> How important is storage capacity tracking in such a setup? Are there some limitations which nodes have access to which storage?

We have different storage cluster with different capacity for node which is divided by topology. Your CSIStorageCapacities really help a lot

bai3shuo4:
> > Yes, I use central one and network storage
> 
> How important is storage capacity tracking in such a setup? Are there some limitations which nodes have access to which storage?

One more thing, I think CSIStorageCapacity should refresh when we expand volume or something. Now I'm working with csi-resizer to help this happening

pohly:
> One more thing, I think CSIStorageCapacity should refresh when we expand volume or something. Now I'm working with csi-resizer to help this happening

The periodic refresh will cover that, but I agree, something that is actually responding to changes would be better. If you have some idea then that would be great.

The modular design of the CSI sidecars is working against us here. If all functionality was in a single "controller sidecar", it would be easier.

pkg/capacity/capacity.go
pohly: Yes, `c.topologyInformer` it is not a `SharedInformer` instance from an informer factory. But conceptually it is the same.

We can rename `Run -> RunWorker` to make that more obvious.

bai3shuo4: @pohly I have revised to this way. Can you review again? Thanks!

pkg/capacity/topology/mock.go
```diff
@@ -54,6 +54,9 @@ func (mt *Mock) List() []*Segment {
 func (mt *Mock) Run(ctx context.Context) {
 }
 
+func (mt *Mock) RunWorker(ctx context.Context) {
+}
+
```
pohly, line 59: This replaces the `Run` method, doesn't it?

pkg/capacity/topology/nodes.go
```diff
@@ -204,13 +204,20 @@ func (nt *nodeTopology) List() []*Segment {
...
 	klog.Info("Shutting node topology informer")
 }
 
+func (nt *nodeTopology) RunWorker(ctx context.Context) {
+	go nt.runWorker(ctx)
+
+	klog.Infof("Started node topology worker")
+	<-ctx.Done()
+	klog.Infof("Shutting node topology worker")
+}
```
pohly, line 219: Same here.

pkg/capacity/topology/topology.go
```diff
@@ -132,6 +132,9 @@ type Informer interface {
 
 	// HasSynced returns true once all segments have been found.
 	HasSynced() bool
+
+	// RunWorker starts a worker to process queue.
+	RunWorker(ctx context.Context)
```
pohly, line 137: And here.

pkg/capacity/capacity.go
```diff
@@ -231,6 +231,7 @@ func (c *Controller) Run(ctx context.Context, threadiness int) {
 	klog.Info("Starting Capacity Controller")
 	defer c.queue.ShutDown()
 
+	go c.topologyInformer.RunWorker(ctx)
```
pohly, line 234: I still think this belongs into csi-provisioner.go. We should add a comment to `NewCentralCapacityController` that "all informers are expected to be started by the caller".

pohly (changes requested):

pkg/capacity/topology/mock.go
bai3shuo4: Yes, you mean I can remove topology informer `Run` method directly?

pkg/capacity/capacity.go
bai3shuo4: Node topology worker only works for Capacity controller. If something goes wrong before CapacityController Run, I think we don't even need to start this worker. Therefore, why not put it into controller run function :)

pohly: Because it is consistent conceptually ("*all* informers are expected to be started by the caller"- we can even add "started *and stopped*").

If something goes wrong in csi-provisioner.go, the binary will exit. It'll make no difference how many goroutines get killed by that and the overhead is also irrelevant.

pkg/capacity/topology/mock.go
pohly: Yes. I thought that was the proposal: clarify that this is not a `SharedInformer` by naming the "Run" method not "Run". Nothing else calls it, so we are free to choose an arbitrary name.

If that makes the patch very large, then perhaps we should simply stick with `Run`.

pohly (changes requested):
Getting close... now we need to describe the change better in the release note.

Can you change that into:
```
Fix capacity information updates when topology changes. Only affected central deployment and network attached storage, not deployment on each node. This broke in v2.2.0 as part of a bug fix for capacity informer handling.
```

pkg/capacity/topology/nodes.go
pohly: Now that we have simplified this function like this, one other enhancement becomes possible: we no longer need the separate `runWorker` function and can save one goroutine:
```
func (nt *nodeTopology) RunWorker(ctx context.Context) {
	klog.Info("Started node topology informer")
        for nt.processNextWorkItem(ctx) {
	}
	klog.Info("Stopped node topology worker")
}
```

pohly (changes requested):
Sorry, one more idea...

pohly:
> Sorry, one more idea...

That can be added later. Let me merge it as-is.

/lgtm
/approve

bai3shuo4:
> Sorry, one more idea...

@pohly Sorry, I just check my github at this moment....yeap, I think we can simplify this func, maybe I can add this commit in further bugfix

PR #618, opened 2021-05-03
Remove dependency to k8s.io/kubernetes

### LINKED ISSUE #603: Remove k8s.io/kubernetes dependency
k8s.io/kubernetes is not intended to be imported, and pulling it in requires us to override many other packages as well. It would be nice to replace the dependency to clean up our vendoring.

/help
/kind cleanup

ialidzhikov:
/kind cleanup 

Fixes #603

```release-note
NONE
```


go.mod
```diff
@@ -39,62 +39,7 @@ require (
...
-
-replace k8s.io/component-base => k8s.io/component-base v0.21.0
-
-replace k8s.io/controller-manager => k8s.io/controller-manager v0.21.0
-
-replace k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.21.0
-
-replace k8s.io/component-helpers => k8s.io/component-helpers v0.21.0
-
-replace k8s.io/mount-utils => k8s.io/mount-utils v0.21.0
```
pohly, line 100: Please don't remove these replace statements.

`go get -u ./...` breaks without them.

pohly: You can prune them by running `release-tools/go-get-kubernetes.sh 1.21.0`.

ialidzhikov: Well, to be honest I am not super convinced that we should keep the replace statements. I now executed `release-tools/go-get-kubernetes.sh 1.21.0` and the end result was pretty much the same - it added the same replace statements.
I see replace statements for pkgs such as  `k8s.io/sample-apiserver`, `k8s.io/sample-cli-plugin`, `k8s.io/sample-controller` which are obviously pkgs that the external-provisioner does not vendor or import in any way. I don't see reason why we should have such replace statements.
To my understanding the main goal of #603 is actually to get rid of these replace statements.
So I don't think `release-tools/go-get-kubernetes.sh` is needed/useful anymore as external-provisioner no longer vendors `k8s.io/kubernetes`. If you want I can put replace statements only for the k8s.io pkgs that external-provisioner vendors and imports.

pohly: > Well, to be honest I am not super convinced that we should keep the replace statements.

Then try running `go get -u ./...` in your branch without them...

> I now executed release-tools/go-get-kubernetes.sh 1.21.0 and the end result was pretty much the same 

Ah, sorry. I thought we had updated `release-tools/go-get-kubernetes.sh` already, but it is still behind the one in `csi-release-tools`. Try the script from https://github.com/kubernetes-csi/csi-release-tools/blob/master/go-get-kubernetes.sh - that one will prune unused replace statements.

> To my understanding the main goal of #603 is actually to get rid of these replace statements.

No, the main goal is to replace code from k/k with code that is meant to be used by out-of-tree components.

> If you want I can put replace statements only for the k8s.io pkgs that external-provisioner vendors and imports.

That's what the updated `go-get-kubernetes.sh` does.

pohly: Hmm, it actually works better than expected. But elsewhere I've seen it update client-go to a 1.x version that was older than the intended 0.21.0, so I stand by my assessment that these replace statements are a useful safeguard against using inconsistent and potentially wrong versions of the Kubernetes packages.

ialidzhikov: > Ah, sorry. I thought we had updated release-tools/go-get-kubernetes.sh already, but it is still behind the one in csi-release-tools. Try the script from https://github.com/kubernetes-csi/csi-release-tools/blob/master/go-get-kubernetes.sh - that one will prune unused replace statements.

Thanks for the hint! The run with the latest `release-tools/go-get-kubernetes.sh` script was much better. I updated the PR now with https://github.com/kubernetes-csi/external-provisioner/compare/a44abfbfb181d32d7c47d048cdf320c1efee0420..6f779f3c7eb8082c534aada2a95d796eaac8ce23.

PS. However the output from `release-tools/go-get-kubernetes.sh 1.21.0` was still not the one that I expected. It updated some of the dependencies to `v0.21.1-rc.0`.

```
$ release-tools/go-get-kubernetes.sh 1.21.0
# [...]
go get: upgraded k8s.io/apimachinery v0.21.0 => v0.21.1-rc.0
SUCCESS
```

From the output I see that the script fetches the `kubernetes-1.21.0` branch instead of the corresponding tag that I want. However this is just a guess and the small issue with the script is a separate topic for enhancement.

msau42: Given that the client-go issue seems to be resolved, I prefer to remove the replace statements to keep this easier to comprehend and update. 

This is fine for now

PR #620, opened 2021-05-04
Fix migration metric registration

### LINKED ISSUE #619: Error collecting metrics with migratable CSI driver
When used with migratable CSI driver, such as `ebs.csi.aws.com`, the external-provisioner v2.2.0 returns 500 response on its metrics endpoint:

```
$ curl -v localhost:8202/metrics
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8202 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8202 (#0)
> GET /metrics HTTP/1.1
> Host: localhost:8202
> User-Agent: curl/7.61.1
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Tue, 04 May 2021 16:20:47 GMT
< Content-Length: 243
< 
An error has occurred while serving metrics:

gathered metric family process_start_time_seconds has help "[ALPHA] Start time of the process since unix epoch in seconds." but should have "Start time of the process since unix epoch in seconds."
```

jsafrane:
/kind bug
Don't register `process_start_time_seconds metric` in migration metrics manager to prevent double registration, resulting in this error:

```
gathered metric family process_start_time_seconds has help "[ALPHA] Start time of the process since unix epoch in seconds." but should have "Start time of the process since unix epoch in seconds."
```

This happens only when a migratable CSI driver is used. With a regular hostpath CSI driver, the metrics work without any issues.

Fixes #619

cc @pohly @Jiawei0227 

```release-note
Fixed reporting of metrics when a migratable CSI driver is used.
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -222,7 +222,10 @@ func main() {
 		klog.V(2).Infof("Supports migration from in-tree plugin: %s", supportsMigrationFromInTreePluginName)
 
 		// Create a new connection with the metrics manager with migrated label
-		metricsManager = metrics.NewCSIMetricsManagerWithOptions(provisionerName, metrics.WithMigration())
+		metricsManager = metrics.NewCSIMetricsManagerWithOptions(provisionerName,
+			// Will be provided via default gatherer.
+			metrics.WithProcessStartTime(false),
+			metrics.WithMigration())
 		migratedGrpcClient, err := ctrl.Connect(*csiEndpoint, metricsManager)
 		if err != nil {
 			klog.Error(err.Error())


pohly (approved):
/lgtm

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -222,7 +222,10 @@ func main() {
 		klog.V(2).Infof("Supports migration from in-tree plugin: %s", supportsMigrationFromInTreePluginName)
 
 		// Create a new connection with the metrics manager with migrated label
-		metricsManager = metrics.NewCSIMetricsManagerWithOptions(provisionerName, metrics.WithMigration())
+		metricsManager = metrics.NewCSIMetricsManagerWithOptions(provisionerName,
+			// Will be provided via default gatherer.
+			metrics.WithProcessStartTime(false),
```
msau42, line 227: Should we remove the processstarttime metric from csi-lib-utils now that it's included by default?

msau42:
Should this fix be cherry-picked?

cmd/csi-provisioner/csi-provisioner.go
pohly: Some sidecars only use the csi-lib-utils code and nothing else, in which case processtarttime still needs to be added there.

msau42: I thought the plan was to also add the base metrics to all sidecars?

pohly: Yes, but that's just a plan. I don't know who's working on it. We can only remove it once all sidecars are updated.

pohly:
> Should this fix be cherry-picked?

Right now it would be simpler to release 2.2.1 from master (all it has are the two bug fixes that need to be cherry-picked) and then fast-forward `release-2.2`, but if you prefer cherry-picking, then please merge https://github.com/kubernetes-csi/external-provisioner/pull/626

gnufied:
Is there a way to catch this from happening? We had some metric e2e tests - https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_metrics.go . I don't think we ported these to CSI suite.

pohly:
> Is there a way to catch this from happening? We had some metric e2e tests - https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_metrics.go .

That looks like it tests metrics data in kube-controller-manager. What we want to test are metrics exposed by the sidecars. Retrieving those from an e2e.test isn't easy because from outside of the cluster, port-forwarding has to be used to reach the metrics endpoint of each container. But I found a solution: https://github.com/intel/pmem-csi/blob/devel/test/e2e/pod/dial.go

Simply retrieving the metrics data would already be useful. But ideally we should also validate that the returned data meets expectations, which means we need to define what those expectations are.

But whether that would have caught this issue here is still uncertain. Do we have e2e tests involving a driver where migration is enabled?

PR #630, opened 2021-05-08
Map PV access modes to CSI access modes based on driver capability

chrishenzie:
**What type of PR is this?**
/kind feature

See [KEP#2485](https://github.com/kubernetes/enhancements/pull/2489) for more details.

This PR introduces mappings between Kubernetes PV access modes and CSI access modes. If the CSI driver supports the `SINGLE_NODE_MULTI_WRITER` controller capability, map ReadWriteOnce to `SINGLE_NODE_MULTI_WRITER` and ReadWriteOncePod to `SINGLE_NODE_SINGLE_WRITER`. If the CSI driver does not support this controller capability, map ReadWriteOnce to `SINGLE_NODE_WRITER` (current behavior preserved for backwards compatibility).

Fixes #

The CSI spec change [PR#476 is here](https://github.com/container-storage-interface/spec/pull/476) and just requires merging. Until then, I have a placeholder commit using my forked version of the CSI spec.

Similarly for K8s, the API change and feature addition [PR#102028 is here](https://github.com/kubernetes/kubernetes/pull/102028) and still is in the early stages of review. Until then, I have a placeholder commit using my forked version of the K8s API.

```release-note
Adds mappings for PV access modes to new CSI access modes: `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER`.
```


pkg/controller/controller.go
```diff
@@ -470,23 +470,70 @@ func getAccessMode(pvcAccessMode v1.PersistentVolumeAccessMode) *csi.VolumeCapab
...
 	}
+}
 
+func (p *csiProvisioner) getVolumeCapabilities(
+	claim *v1.PersistentVolumeClaim,
+	sc *storagev1.StorageClass,
+	fsType string,
+) []*csi.VolumeCapability {
+	supportsSingleNodeMultiWriter := false
+	if p.controllerCapabilities[csi.ControllerServiceCapability_RPC_SINGLE_NODE_MULTI_WRITER] {
```
jingxu97, line 528: what if controller only has SINGLE_NODE_SINGLE_WRITER?

chrishenzie: The capability name does not make it entirely clear, but in the CSI spec we define that the `SINGLE_NODE_MULTI_WRITER` capability means the driver supports the `SINGLE_NODE_SINGLE_WRITER` and/or `SINGLE_NODE_MULTI_WRITER` access modes.

The reason the capability is named `SINGLE_NODE_MULTI_WRITER` is because this access mode is missing from the CSI spec. This applies to drivers that strictly followed the NodePublishVolume rules which made `SINGLE_NODE_WRITER` == `SINGLE_NODE_SINGLE_WRITER`).

humblec: Indeed, the capabilities in CSI spec  and its indirect capability pass through is confusing a lot :) . Regardless, with the addition of explicit `SINGLE_NODE_SINGLE_WRITER and SINGLE_NODE_MULTI_WRITER` in recent spec, the situation is getting improved. `ReadWriteOncePod` is much needed one too.  :+1: .

humblec:
@chrishenzie ( just as a suggestion) it may be easy to have `table/matrix` representation of `ToCSIAccessMode` logic in source code comment or in the PR to make sure we are covering all the scenario..

chrishenzie:
@humblec I've added the tables to https://github.com/kubernetes-csi/csi-lib-utils/pull/89, PTAL, I plan on depending on this library across our sidecars.

humblec:
> @humblec I've added the tables to [kubernetes-csi/csi-lib-utils#89](https://github.com/kubernetes-csi/csi-lib-utils/pull/89), PTAL, I plan on depending on this library across our sidecars, I just need to update this one to do that.

@chrishenzie Thanks :+1: , it really helps !

chrishenzie:
/hold

Rebased on #656 which is blocked on release of v1.22.

PR #635, opened 2021-05-23
Add SupportsBlock method to provisionWrapper for being able to use block device

### LINKED ISSUE #631: SupportsBlock method does not work
Hi, currently, block devices can not use because provisionWrapper struct doesn't have SupportsBlock method.
I think this bug probably has occurred when https://github.com/kubernetes-csi/external-provisioner/pull/586 PR was merged.

ref:
- https://github.com/kubernetes-csi/external-provisioner/blob/359a0e068e079bfa79df75db31e03f32d7743b1f/cmd/csi-provisioner/csi-provisioner.go#L482-L483
- https://github.com/kubernetes-csi/external-provisioner/blob/359a0e068e079bfa79df75db31e03f32d7743b1f/pkg/capacity/provision.go#L26-L87

/kind bug

bells17:
**What type of PR is this?**

/kind bug

Add SupportsBlock method to provisionWrapper for being able to use block device.

Fixes #631

```release-note
Fix a bug that not being able to use block device mode when enable a storage capacity tracking mode.
```

### DIFF
--- a/pkg/capacity/provision.go
+++ b/pkg/capacity/provision.go
@@ -30,6 +30,7 @@ type provisionWrapper struct {
 
 var _ controller.Provisioner = &provisionWrapper{}
 var _ controller.BlockProvisioner = &provisionWrapper{}
+var _ controller.Qualifier = &provisionWrapper{}
 
 func NewProvisionWrapper(p controller.Provisioner, c *Controller) controller.Provisioner {
 	return &provisionWrapper{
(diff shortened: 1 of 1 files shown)


pohly:
/assign
/ok-to-test

I will check why this was not detected by tests...

pohly:
> I will check why this was not detected by tests...

Two reasons:
- We don't have pull jobs in the external-provisioner repo for "distributed provisioning". Given that this feature is getting used and thus important, we should add such a job to detect regressions before merging them. There is a pull-kubernetes-csi-external-provisioner-distributed-on-kubernetes-master job, but it couldn't be enabled at the time because it depended on a newer csi-driver-host-path release. It should be doable now.
- The only periodic job, distributed-on-master, was not using canary images. This looks like an oversight, I don't think I did that intentionally.

pohly:
`pull-kubernetes-csi-external-provisioner-distributed-on-kubernetes-1-20` is a new job which should cover this functionality.

pohly:
> pull-kubernetes-csi-external-provisioner-distributed-on-kubernetes-1-20 is a new job which should cover this functionality.

... or perhaps not. The deployment itself tries to capacity tracking, but only does so if enabled in Kubernetes. That wasn't the case in 1.20.

I need to use 1.21.

pohly:
`pull-kubernetes-csi-external-provisioner-distributed-on-kubernetes-1-21` is passing here and about to fail in a test PR (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-provisioner/637/pull-kubernetes-csi-external-provisioner-distributed-on-kubernetes-1-21/1397214642060660736).

This confirms that the functionality is now covered.

bells17:
Thank you.

Note to Self:
> pull-kubernetes-csi-external-provisioner-distributed-on-kubernetes-1-21

it was addded  from [this PR](https://github.com/kubernetes/test-infra/pull/22337)

Issue #638, opened 2021-05-31
test flake: capacity metrics check fails


pohly:
Sometimes, `go test ./pkg/capacity` fails. I can reproduce this locally by running it continuously until it fails:
```
$ i=0; while go  test -count=1 ./pkg/capacity/; do echo $i; i=$((i + 1)); done
ok  	github.com/kubernetes-csi/external-provisioner/pkg/capacity	12.248s
0
ok  	github.com/kubernetes-csi/external-provisioner/pkg/capacity	12.243s
1
ok  	github.com/kubernetes-csi/external-provisioner/pkg/capacity	12.245s
2
ok  	github.com/kubernetes-csi/external-provisioner/pkg/capacity	12.246s
3
ok  	github.com/kubernetes-csi/external-provisioner/pkg/capacity	12.246s
4
--- FAIL: TestCapacityController (0.00s)
    --- FAIL: TestCapacityController/two_segments,_two_classes,_four_objects_updated (0.01s)
        capacity.go:336: INFO onTopologyChanges: topology changed added=[0x27ef470 = layer0: foo 0x27ef490 = layer0: bar] removed=[]
        capacity.go:371: INFO onSCAddOrUpdate: updated or added storageclass="direct-sc"
        capacity.go:479: INFO onSCAddOrUpdate: enqueuing storageclass="direct-sc" workitem={segment:0x27ef470 storageClassName:direct-sc}
        capacity.go:479: INFO onSCAddOrUpdate: enqueuing storageclass="direct-sc" workitem={segment:0x27ef490 storageClassName:direct-sc}
        capacity.go:371: INFO onSCAddOrUpdate: updated or added storageclass="triple-sc"
        capacity.go:479: INFO onSCAddOrUpdate: enqueuing storageclass="triple-sc" workitem={segment:0x27ef470 storageClassName:triple-sc}
        capacity.go:336: INFO onTopologyChanges: topology changed added=[0x27ef470 = layer0: foo 0x27ef490 = layer0: bar] removed=[]
        capacity.go:479: INFO onSCAddOrUpdate: enqueuing storageclass="triple-sc" workitem={segment:0x27ef490 storageClassName:triple-sc}
        capacity.go:479: INFO onTopologyChanges: enqueuing workitem={segment:0x27ef470 storageClassName:direct-sc}
        capacity.go:479: INFO onTopologyChanges: enqueuing workitem={segment:0x27ef490 storageClassName:direct-sc}
[...]

PR #639, opened 2021-05-31
RFC: structured, contextual logging

pohly:
**What type of PR is this?**
/kind cleanup

Embracing go-logr as logger has several advantages:
- adding a certain prefix and/or set of objects can be done consistently with less code duplication
- it's even possible for the caller to do that, something that doesn't work when the prefix is part of the format string
- in Go tests, all output can be associated with the currently running test via https://github.com/kubernetes/klog/pull/240

The latter was needed to debug https://github.com/kubernetes-csi/external-provisioner/issues/638

```release-note
Log messages are structured (fixed message, additional information in key/value pairs).
```


pkg/capacity/capacity.go
```diff
@@ -228,24 +232,28 @@ var _ metrics.StableCollector = &Controller{}
...
-	klog.Info("Started Capacity Controller")
+	logger.Info("started controller")
 	<-ctx.Done()
-	klog.Info("Shutting down Capacity Controller")
+	logger.Info("shutting down controller")
 }
 
-func (c *Controller) prepare(ctx context.Context) {
+func (c *Controller) prepare(ctx context.Context, logger logr.Logger) {
+	logger = logger.WithName("prepare")
```
pohly, line 255: I have mixed feelings about explicitly passing a logger here and in the other functions which already take a context.

One the one hand it makes it explicit that the function is going to use a `logr.Logger` and needs it.

On the other hand, it is a redundant parameter because `ctx` could be used to provide the logger. A function that retrieves the logger from the context and uses `klogr.New` as fallback, i.e. never returns nil, would be ideal.

At least here we *can* add a parameter. There are plenty of established APIs which take a context, but cannot be changed easily to also take a logger - see client-go.

This is one of the open questions around the usage of `logr` in Kubernetes...

pohly: For comparison, in `capacity_test.go` I used `logr.FromContext` because that file controlled the context and could be sure to not get nil.

go.mod
```diff
@@ -61,3 +61,6 @@ replace k8s.io/component-base => k8s.io/component-base v0.21.0
 replace k8s.io/component-helpers => k8s.io/component-helpers v0.21.0
 
 replace k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.21.0
+
+// WIP
+replace k8s.io/klog/v2 => github.com/pohly/klog/v2 v2.4.1-0.20210527141230-ac596814502c
```
xing-yang, line 66: What's the plan for this?

pkg/capacity/capacity.go
```diff
@@ -228,24 +232,28 @@ var _ metrics.StableCollector = &Controller{}
...
-			c.runWorker(ctx)
+			c.runWorker(ctx, logger)
 		}, time.Second)
 	}
 
-	go wait.UntilWithContext(ctx, func(ctx context.Context) { c.pollCapacities() }, c.pollPeriod)
+	go wait.UntilWithContext(ctx, func(ctx context.Context) { c.pollCapacities(logger) }, c.pollPeriod)
 
-	klog.Info("Started Capacity Controller")
+	logger.Info("started controller")
```
xing-yang, line 249: What's the reason for starting an info log message with lower case?

pkg/capacity/capacity.go
```diff
@@ -377,24 +389,25 @@ func (c *Controller) onSCDelete(sc *storagev1.StorageClass) {
 	if sc.Provisioner != c.driverName {
 		return
 	}
+	logger := c.logger.WithName("onSCDelete").WithValues("storageclass", klog.KObj(sc))
 
-	klog.V(3).Infof("Capacity Controller: storage class %s was removed", sc.Name)
+	logger.V(3).Info("removed")
```
xing-yang, line 394: Can you show an example of the output of this changed log msg vs the original log msg?

go.mod
pohly: I proposed to have the code in klog: https://github.com/kubernetes/klog/pull/240

It's currently on hold because the logr API changes need to be dealt with first.

pkg/capacity/capacity.go
pohly: It's not a full sentence, therefore initial capital letter looked odd. I don't know whether there is some guidance on this.

It's also consistent with error messages. For those the official guidance is to start with lower case because the error might get wrapped.

pkg/capacity/capacity.go
pohly: Looks like I have a coverage gap in capacity_test.go - onSCDelete is not called. Will fix.

In the meantime, here's the corresponding output from onSCAddedOrUpdated, as printed by `go test`. With these changes:
```
capacity.go:373: INFO onSCAddOrUpdate: updated or added storageclass="triple-sc"
...
capacity.go:481: INFO onSCAddOrUpdate: enqueuing storageclass="triple-sc" workitem={segment:0x27ef430 storageClassName:triple-sc}
```

Without them:
```
I0608 17:48:40.721547  866697 capacity.go:361] Capacity Controller: storage class triple-sc was updated or added
...
I0608 17:48:40.721526  866697 capacity.go:468] Capacity Controller: enqueuing {segment:0x27e7350 storageClassName:other-sc}
```

Note that the "enqueuing" messages without this PR lacks context. It's not clear why `addWorkItem` was called. With contextual logging, the `onSCAddOrUpdate` function name and the storage class get passed down and are added to the log message.

This is important once things start to happen in parallel. When everything is sequential, one can read the log from top to bottom and remember which values were logged earlier. But when run in parallel, it's not clear whether log output actually follows the one printed directly above it.

This is a problem in our driver logs where it is hard to associate a gRPC error response with the corresponding call.

pohly: Here's the log output for a complete testcase:
```
=== RUN   TestRefresh
=== RUN   TestRefresh/truncated_topology
=== PAUSE TestRefresh/truncated_topology
=== CONT  TestRefresh/truncated_topology
    capacity.go:338: INFO onTopologyChanges: topology changed added=[0x27ef430 = layer0: foo+ layer1: X+ layer2: A 0x27ef450 = layer0: foo+ layer1: X+ layer2: B] removed=[]
    capacity.go:373: INFO onSCAddOrUpdate: updated or added storageclass="direct-sc"
    capacity.go:481: INFO onSCAddOrUpdate: enqueuing storageclass="direct-sc" workitem={segment:0x27ef430 storageClassName:direct-sc}
    capacity.go:481: INFO onSCAddOrUpdate: enqueuing storageclass="direct-sc" workitem={segment:0x27ef450 storageClassName:direct-sc}
    capacity.go:373: INFO onSCAddOrUpdate: updated or added storageclass="triple-sc"
    capacity.go:481: INFO onSCAddOrUpdate: enqueuing storageclass="triple-sc" workitem={segment:0x27ef430 storageClassName:triple-sc}
    capacity.go:481: INFO onSCAddOrUpdate: enqueuing storageclass="triple-sc" workitem={segment:0x27ef450 storageClassName:triple-sc}
    capacity.go:338: INFO onTopologyChanges: topology changed added=[0x27ef430 = layer0: foo+ layer1: X+ layer2: A 0x27ef450 = layer0: foo+ layer1: X+ layer2: B] removed=[]
    capacity.go:481: INFO onTopologyChanges: enqueuing workitem={segment:0x27ef430 storageClassName:direct-sc}
    capacity.go:481: INFO onTopologyChanges: enqueuing workitem={segment:0x27ef450 storageClassName:direct-sc}
    capacity.go:481: INFO onTopologyChanges: enqueuing workitem={segment:0x27ef430 storageClassName:triple-sc}
    capacity.go:481: INFO onTopologyChanges: enqueuing workitem={segment:0x27ef450 storageClassName:triple-sc}
    capacity.go:276: INFO prepare: initial state topology segments=2 storage classes=2 potential CSIStorageCapacity objects=4
    capacity.go:288: INFO prepare: checking for existing CSIStorageCapacity objects
--- PASS: TestRefresh (0.00s)
```

Note that this would be impossible to do without this PR because log output from different test cases would be mixed.

pkg/capacity/capacity.go
xing-yang: I'd like to see if the structured log KEP has any official guidance on this.  The official guidance for error message is only referring to the returned error message, not error messages in the logs.

pohly: The logr example uses lower case, incidentally also with a "starting" message:
https://github.com/go-logr/logr#typical-usage

But the Kubernetes documentation says "Start from a capital letter": https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#remove-string-formatting-from-log-message

It also has some other recommendations. Once we agree to go further with this and the klog PR is merged, I'll revisit the log messages and update them accordingly.

pohly:
/remove-lifecycle rotten

I will come back to this once contextual logging is usable in Kubernetes. It's planned for alpha in 1.24.

pohly:
/remove-lifecycle rotten

I was looking for a volunteer to continue with this, but so far without luck. I'll probably finish this myself.

bells17:
Hi @pohly 

> I was looking for a volunteer to continue with this, but so far without luck. I'll probably finish this myself.

Would it be acceptable if I were to take on this development task?

pohly:
@bells17: help with this would be very welcome. Feel free to take my branch, rebase it and continue in a new PR.

I think with this PR and https://github.com/kubernetes-csi/node-driver-registrar/pull/259 it is technically clear how to use component-base/logs. The rest of the conversion can go as described in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md

PR #643, opened 2021-06-10
Add command line arguments to configure leader election options

### LINKED ISSUE #556: Add command line arguments for leader election parameters
The leader election package specified 3 variables that are related to the time interval of acquiring and renewing leases:

```
// leaderElection is a convenience wrapper around client-go's leader election library.
type leaderElection struct {
...
	leaseDuration time.Duration
	renewDeadline time.Duration
	retryPeriod   time.Duration
}
```

The leader election documentation states:

```
// Larger clusters often have a more lenient SLA for API latency. This should be
// taken into account when configuring the client. The rate of leader transitions
// should be monitored and RetryPeriod and LeaseDuration should be increased
// until the rate is stable and acceptably low. 
```

Does it make sense to introduce command line arguments for these three parameters so that storage vendors can specify their own values. 

I've noticed in our deployments for vSphere CSI driver that external sidecars are actually restarting multiple times (temporarily entering the `CrashLoopBackOff` state) because they lose leader election lease. 
For example:

```
# kubectl get pods -n vmware-system-csi
NAME                                     READY   STATUS    RESTARTS   AGE
vsphere-csi-controller-8554589bd-6kbxt     5/6     CrashLoopBackOff   456          12h
...
# kubectl get pods -n vmware-system-csi
NAME                                     READY   STATUS    RESTARTS   AGE
vsphere-csi-controller-8554589bd-6kbxt   6/6     Running   824        23h
```

NOTE this applies to all sidecars, but just starting the discussion here. 

/kind feature

RaunakShah:
**What type of PR is this?**

/kind feature

This PR adds command line arguments for leader election options - Lease Duration, Renew Deadline and Retry Period. Currently the defaults are 15, 10 and 5 seconds respectively. Without this change, there's no way for a customer/storage plugin to configure those values. 

This PR makes those parameters configurable. Storage plugins deploying external-provisioner can tune those parameters in their YAMLs accordingly. 

Fixes #556

Testing:

Created Provisioner with new arguments and verified that Pod is running:
```
# cat vsphere-csi-driver.yaml
...
        - name: csi-provisioner
          args:
            - "--v=4"
            - "--timeout=300s"
            - "--csi-address=$(ADDRESS)"
            - "--kube-api-qps=100"
            - "--kube-api-burst=100"
            - "--leader-election"
            - "--default-fstype=ext4"
            - "leader-election-lease-duration=30"
            - "leader-election-renew-deadline=20"
            - "leader-election-retry-period=10"
...
# kubectl get pod -n vmware-system-csi
NAME                                      READY   STATUS    RESTARTS   AGE
vsphere-csi-controller-74f57f8754-x2bcj   6/6     Running   0          21m
```

```release-note
Add command line arguments to configure leader election options
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,12 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--leader-election-namespace`: Namespace where leader election object will be created. It is recommended that this parameter is populated from Kubernetes DownwardAPI with the namespace where the external-provisioner runs in.
 
+* `--leader-election-lease-duration <duration>`: Duration, in seconds, that non-leader candidates will wait to force acquire leadership. Defaults to 15 seconds.
+
+* `--leader-election-renew-deadline <duration>`: Duration, in seconds, that the acting leader will retry refreshing leadership before giving up. Defaults to 10 seconds.
+
+* `--leader-election-retry-period <duration>`: Duration, in seconds, the LeaderElector clients should wait between tries of actions. Defaults to 5 seconds.
+
 * `--timeout <duration>`: Timeout of all calls to CSI driver. It should be set to value that accommodates majority of `ControllerCreateVolume` and `ControllerDeleteVolume` calls. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details. 15 seconds is used by default.
 
 * `--retry-interval-start <duration>`: Initial retry interval of failed provisioning or deletion. It doubles with each failure, up to `--retry-interval-max` and then it stops increasing. Default value is 1 second. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details.
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -85,6 +85,10 @@ var (
 	httpEndpoint            = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including metrics and leader election health check, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--metrics-address` and `--http-endpoint` can be set.")
 	metricsPath             = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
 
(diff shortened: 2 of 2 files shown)


ggriffiths:
Looks good once the go fmt issue is fixed

RaunakShah:
@ggriffiths yup thats the plan! Thanks for the quick review!

xing-yang:
Should there be "--" in front of these new options?

```# cat vsphere-csi-driver.yaml
...
        - name: csi-provisioner
          args:
            - "--v=4"
            - "--timeout=300s"
            - "--csi-address=$(ADDRESS)"
            - "--kube-api-qps=100"
            - "--kube-api-burst=100"
            - "--leader-election"
            - "--default-fstype=ext4"
            - "leader-election-lease-duration=30"
            - "leader-election-renew-deadline=20"
            - "leader-election-retry-period=10"
...
```

xing-yang:
> there's no way for a customer/storage plugin to configure those values.

My understanding is that these are tunable options when starting the external-provisioner so whoever deploys the CSI driver can configure them?

RaunakShah:
@xing-yang 

> Should there be "--" in front of these new options?

Fixed the description! And redeployed with the `--`. 

> My understanding is that these are tunable options when starting the external-provisioner so whoever deploys the CSI driver can configure them?

That's correct, they can be set before creating the pod. I've also added the parameters to the documentation page.

xing-yang:
@RaunakShah, I see the following in PR description:
```
This PR adds command line arguments for leader election options - Lease Duration, Renew Deadline and Retry Period. Currently the defaults are 15, 10 and 5 seconds respectively and there's no way for a customer/storage plugin to configure those values.
```
If these options are tunable, why you said "there's no way for a customer/storage plugin to configure those values"?

RaunakShah:
>  Do you mean, without this change, they can't be configured; but with this change, they are tunable? 

@xing-yang This is correct, apologies for the confusion. I've updated the PR description with more clarity!

jingxu97:
@RaunakShah Thanks for the PR. Do you mind making the similar changes to other csi sidecars? Thanks!

PR #655, opened 2021-07-27
WIP: use csi-lib-utils for CSI secret handling logic

### LINKED ISSUE #653: Make CSI secret handling logic reusable
CSI secret can be defined as a template as defined [here](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html#storageclass-secrets). Current implementation of external-provisioner embeds [this logic](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L1581) in the controller codes as private scoped, which isn't reusable from other controllers.

There are use cases, like PVC transfer and secret protection, to use this logic from other controllers, therefore it needs to be changed to public scoped and move to a kind of util packages.

mkimuram:
**What type of PR is this?**
/kind cleanup

Use csi-lib-utils for CSI secret handling logic. These codes are moved to csi-lib-utils by https://github.com/kubernetes-csi/csi-lib-utils/pull/95 .

Fixes #653

```release-note

```


mkimuram:
Still WIP, because kubernetes-csi/csi-lib-utils#95 needs to be tested before this is merged.
Only tested with unit tests, currently.

PR #661, opened 2021-08-05
Fix recommended Kubernetes version in README

chrishenzie:
**What type of PR is this?**
/kind cleanup
/kind documentation

Fixes an issue in the README where the recommended Kubernetes version is set to one higher than it should be.

See for more details:
https://kubernetes-csi.github.io/docs/project-policies.html#recommended-version

Fixes #

```release-note
NONE
```

/assign @xing-yang

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This information reflects the head of this branch.
 
 | Compatible with CSI Version | Container Image | [Min K8s Version](https://kubernetes-csi.github.io/docs/kubernetes-compatibility.html#minimum-version) | [Recommended K8s Version](https://kubernetes-csi.github.io/docs/kubernetes-compatibility.html#recommended-version) |
 | ------------------------------------------------------------------------------------------ | -------------------------------| --------------- | ------------- |
-| [CSI Spec v1.5.0](https://github.com/container-storage-interface/spec/releases/tag/v1.5.0) | k8s.gcr.io/sig-storage/csi-provisioner | 1.17 | 1.22 |
+| [CSI Spec v1.5.0](https://github.com/container-storage-interface/spec/releases/tag/v1.5.0) | k8s.gcr.io/sig-storage/csi-provisioner | 1.17 | 1.21 |
 
 ## Feature status
 


README.md
```diff
@@ -13,7 +13,7 @@ This information reflects the head of this branch.
 
 | Compatible with CSI Version | Container Image | [Min K8s Version](https://kubernetes-csi.github.io/docs/kubernetes-compatibility.html#minimum-version) | [Recommended K8s Version](https://kubernetes-csi.github.io/docs/kubernetes-compatibility.html#recommended-version) |
 | ------------------------------------------------------------------------------------------ | -------------------------------| --------------- | ------------- |
-| [CSI Spec v1.5.0](https://github.com/container-storage-interface/spec/releases/tag/v1.5.0) | k8s.gcr.io/sig-storage/csi-provisioner | 1.17 | 1.22 |
+| [CSI Spec v1.5.0](https://github.com/container-storage-interface/spec/releases/tag/v1.5.0) | k8s.gcr.io/sig-storage/csi-provisioner | 1.17 | 1.21 |
```
msau42, line 16: Why not 1.22? That's needed for your access mode changes right?

chrishenzie: > The recommended Kubernetes version specifies the lowest Kubernetes version needed where all its supported features will function correctly.

I might be confused on the wording here, are features referring to features added to the CSI sidecar? The feature in this case would be the new CSI access modes. Customers can leverage the new `SINGLE_NODE_MULTI_WRITER` access mode with older Kubernetes versions if they supply "ReadWriteOnce". If they want to leverage `SINGLE_NODE_SINGLE_WRITER` they will need Kubernetes 1.22 in order to supply ReadWriteOncePod. So in that case, then 1.22 would recommended if you want to leverage both of the new CSI access modes that comprise the feature.

chrishenzie: @xing-yang

msau42: The recommended version is the version of Kubernetes needed for all features of the sidecar to function correctly. Since your access modes change depends on the new access mode in 1.22, I think it is the right version to recommend.

jsafrane: +1 to @msau42, the recommended version should support all features that the sidecar supports, hence 1.22 for access modes.

PR #662, opened 2021-08-10
 Check error on mockserver start and remove unused clientset arg

humblec:

> /kind cleanup

```release-note-none

```


pkg/controller/controller_test.go
```diff
@@ -106,8 +106,10 @@ func createMockServer(t *testing.T, tmpdir string) (*gomock.Controller,
 		Identity:   identityServer,
 		Controller: controllerServer,
 	})
-	drv.StartOnAddress("unix", filepath.Join(tmpdir, "csi.sock"))
-
+	err := drv.StartOnAddress("unix", filepath.Join(tmpdir, "csi.sock"))
+	if err != nil {
+		t.Fatalf("Failed to start mock server: %v", err)
```
chrishenzie, line 111: It looks like the calling function will also `t.Fatal()` if `createMockServer()` returns an error. Maybe we can return an error here instead which would also allow us to remove the `t` argument.

chrishenzie (changes requested):

pkg/controller/controller_test.go
humblec: @chrishenzie , yeah, that would be better , adjusted the patchset. PTAL , Thanks :+1:

PR #670, opened 2021-09-02
distributed provisioning: avoid errors from lib about foreign nodes

### LINKED ISSUE #669: distributed provisioning: PVC for other node not ignored silently

pohly:
**What type of PR is this?**
/kind cleanup

The provisioner receives PVCs for all nodes. This is necessary to support
provisioning of PVCs with immediate binding. When it receives a PVC with a
selected node that isn't the local one, the fake node informer didn't have an
object for that node, which caused this error and a corresponding event:

```
  E0902 17:33:24.229890       1 controller.go:981] error syncing claim "f5008265-3c43-4385-a78d-4fa3f72fb611": failed to get target node: node "aks-workerpool-15818640-vmss00000a" not found
```

Fixes #669

This could be avoided if the lib didn't look up the node for the
provisioner. We discussed that a while back and decided against it at the
time. Given that situation, the simplest solution is to ensure that the lib
always gets a node object.

```release-note
distributed provisioning: avoid harmless errors about "failed to get target node"
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -110,6 +110,33 @@ var (
 	version             = "unknown"
 )
 
+type nodeDeploymentLister struct {
+	listersv1.NodeLister
+	localNodeName string
+}
+
+func (n nodeDeploymentLister) Get(name string) (*v1.Node, error) {
+	if name == n.localNodeName {
+		// A copy for the local was stored beforehand.
+		return n.NodeLister.Get(name)
+	}
+	// For every other node we return a minimal object to keep
+	// sig-storage-lib-external-provisioner happy:
+	// https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/853cdd7cba6e20753071cdb3e07fe37a6f2fc5a3/controller/controller.go#L1326
+	//
+	// When it calls ShouldProvision, we return false and the
+	// fake node object never gets used.
+	//
+	// If we don't do this, sig-storage-lib-external-provisioner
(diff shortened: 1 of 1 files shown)


pohly:
@wongma7, @jsafrane : what do you think?

The alternative is to change or extend the sig-storage-lib-external-provisioner API (don't look up Node there, instead pass node name). The solution from this PR seems simpler.

jsafrane:
To me it looks quite fragile. sig-storage-lib-external-provisioner uses the nodeLister only after `shouldProvision` returns `true`. Can the external-provisioner filter out provisioning calls for other nodes in its `ShouldProvision` implementation?

pohly:
> Can the external-provisioner filter out provisioning calls for other nodes in its ShouldProvision implementation?

It already does that: https://github.com/kubernetes-csi/external-provisioner/blob/d6861307ef31020e93610d9708acc640adbacdcb/pkg/controller/controller.go#L1256

I think what happened here is that a claim was originally meant to be provisioned on the node (ShouldProvision returned true), but then got rescheduled. Why it then keeps trying to obtain the node and call Provision is a bit unclear - perhaps that is the real problem. Let me check that.

/hold

pohly:
> ShouldProvision returned true

Good news, bad news... The bad news is that `ShouldProvision` in v2.2.x wasn't getting called at all when enabling storage capacity tracking. The dood news is that this was fixed as part of https://github.com/kubernetes-csi/external-provisioner/pull/635, so the fix will be in v3.0.0.

This wasn't obvious because even at log -v5 there was no output about that check. I wonder whether that should be added.

But for now:
/close

PR #680, opened 2021-11-04
Add TLS support for diagnostics HTTP server

VirrageS:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: It adds TLS support for diagnostics HTTP server. In our cluster we require the metrics to be exposed as HTTPS. Not having this support stops us from using metrics from `external-provisioner`.

I'm not fully sure about the naming of the flags. I've looked at https://kubernetes.io/docs/setup/best-practices/certificates/ but there isn't seem to be consensus about the flag naming for TLS files. 

**Special notes for your reviewer**: NONE

```release-note
Add TLS support for diagnostics/metrics/leader election health status HTTP server by adding new flags: `--http-endpoint-key-file` and `--http-endpoint-cert-file`.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,13 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--cloning-protection-threads <num>`: Number of simultaneously running threads, handling cloning finalizer removal. Defaults to `1`.
 
-* `--http-endpoint`: The TCP network address where the HTTP server for diagnostics, including metrics and leader election health check, will listen (example: `:8080` which corresponds to port 8080 on local host). The default is empty string, which means the server is disabled.
+* `--http-endpoint`: The TCP network address where the HTTP(S) server for diagnostics, including pprof, metrics and leader election health check, will listen (example: `:8080` which corresponds to port 8080 on local host). The default is empty string, which means the server is disabled.
 
 * `--metrics-path`: The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.
+
+* `--http-endpoint-key-file`: The path to TLS key file used for running HTTPS server. The default is empty string, which means that TLS will not be used when running HTTP server. Both `--http-endpoint-key-file` and `--http-endpoint-cert-file` flags must be set to enable TLS support.
+
+* `--http-endpoint-cert-file`: The path to TLS cert file used for running HTTPS server. The default is empty string, which means that TLS will not be used when running HTTP server. Both `--http-endpoint-key-file` and `--http-endpoint-cert-file` flags must be set to enable TLS support.
 
 * `--extra-create-metadata`: Enables the injection of extra PVC and PV metadata as parameters when calling `CreateVolume` on the driver (keys: "csi.storage.k8s.io/pvc/name", "csi.storage.k8s.io/pvc/namespace", "csi.storage.k8s.io/pv/name")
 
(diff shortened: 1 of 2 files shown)


VirrageS:
I signed it

README.md
```diff
@@ -69,10 +69,14 @@ Note that the external-provisioner does not scale with more replicas. Only one e
 
 * `--cloning-protection-threads <num>`: Number of simultaneously running threads, handling cloning finalizer removal. Defaults to `1`.
 
-* `--http-endpoint`: The TCP network address where the HTTP server for diagnostics, including metrics and leader election health check, will listen (example: `:8080` which corresponds to port 8080 on local host). The default is empty string, which means the server is disabled.
+* `--http-endpoint`: The TCP network address where the HTTP(S) server for diagnostics, including metrics and leader election health check, will listen (example: `:8080` which corresponds to port 8080 on local host). The default is empty string, which means the server is disabled.
 
 * `--metrics-path`: The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.
 
+* `--http-endpoint-key-file`: The path to TLS key file used for running HTTPS server. The default is empty string, which means that TLS will not be used when running HTTP server. Both `--http-endpoint-key-file` and `--http-endpoint-cert-file` flags must be set to enable TLS support.
```
Jiawei0227, line 76: Do we need a CA file as well? Who signed this http certificate?

VirrageS: No, no need for CA file. The user must provide signed cert file.

Jiawei0227: no need for CA file means this  cert is a self-signed cert?

VirrageS: > In a typical public-key infrastructure (PKI) scheme, the certificate issuer is a certificate authority (CA), usually a company that charges customers to issue certificates for them.

So the CA issues the certificate and we just need to provide path to it. It can be self-signed cert (local CA).

I don't see a reason why would we need CA file? Could you explain that?

Jiawei0227: Bare with me if my understand is not correct, I am not very good at certs. The provisioner itself will serve as a HTTPs server, so if there is a client want to talk to this server, the server need to verify the client cert is signed by the correct CA. So basically supplying it a CA file will tell the HTTPs server which CA to trust. This is my understanding, I might be wrong but happy to learn it

Jiawei0227: Bare with me if my understand is not correct, I am not very good at certs. The provisioner itself will serve as a HTTPs server, so if there is a client want to talk to this server, the server need to verify the client cert is signed by the correct CA. So basically supplying it a CA file will tell the HTTPs server which CA to trust. This is my understanding, I might be wrong but happy to learn it

Jiawei0227: cc @msau42 to help me with my understanding

VirrageS: As far I understand this, the client doesn't need to provide the certs. The client certs work as authentication but this isn't in the scope of this MR. This MR deals with server certs. What this means is that the client issues a request and wants to check if the server is indeed valid, secure and trusted entity (not some kind of impostor).

> In SSL handshake process, a client makes a request for a secured connection, the server shows its public certificate for the acceptance. The client has already trusted certificate lists stored in called truststore. If the presented certificate matches with the prescribed list of certificates, an encrypted connection could be established.

Jiawei0227: oh I see. So here in your configuration you did not enable client-side verification on the HTTPs server. So thats why the CA is not needed. I guess it is okay

Jiawei0227:
/lgtm
cc @xing-yang for approval

jsafrane:
I am not sure about this. HTTPS is not only about encryption. See https://github.com/kubernetes-csi/external-attacher/pull/331#issuecomment-965015345

VirrageS:
@jsafrane @xing-yang  is there anything that can be done here to make this mergable?

PR #681, opened 2021-11-10
delete the invalid case in  verifyAndGetSecretNameAndNamespaceTemplate

jianglingxia:
**What type of PR is this?**

> /kind cleanup

Fixes # the invalid case in function verifyAndGetSecretNameAndNamespaceTemplate

```release-note

```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -145,7 +145,7 @@ var (
 	}
 
 	provisionerSecretParams = secretParamsMap{
-		name: "Provisioner",
+		name:                         "Provisioner",
 		deprecatedSecretNameKey:      provisionerSecretNameKey,
 		deprecatedSecretNamespaceKey: provisionerSecretNamespaceKey,
 		secretNameKey:                prefixedProvisionerSecretNameKey,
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -145,31 +145,31 @@ var (
 	}
 
 	provisionerSecretParams = secretParamsMap{
-		name:                         "Provisioner",
+		name: "Provisioner",
```
chrishenzie, line 148: Did you mean to change this indentation? I'm pretty sure `gofmt` will add this back when its ran

chrishenzie (changes requested):
/ok-to-test

pkg/controller/controller.go
jianglingxia: I run gofmt in normal format, but the file is indented when I save it

jianglingxia:
@chrishenzie can you help to review this commit? thanks

chrishenzie:
Hi, thanks for the ping. Quick question, I have less context here, can you provide context why this case is not possible? Also, can you squash your indentation commit into the previous one?

PR #688, opened 2021-12-10
bugfix: get capacity grpc request should have timeout

bai3shuo4:
/kind bug

bug
If one node csidriver fails or it does not set any timeout for grpc, this sync capacity will stuck here. If we only use one thread to process CSIStoragecapacities, it will stuck forever. Provisioning pv or delete all has timeout, it should have timeout for GetCapacity as well

Fixes #

```release-note
Add getCapacity request timeout to avoid hang forever
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -74,7 +74,7 @@ var (
 	workerThreads        = flag.Uint("worker-threads", 100, "Number of provisioner worker threads, in other words nr. of simultaneous CSI calls.")
 	finalizerThreads     = flag.Uint("cloning-protection-threads", 1, "Number of simultaneously running threads, handling cloning finalizer removal")
 	capacityThreads      = flag.Uint("capacity-threads", 1, "Number of simultaneously running threads, handling CSIStorageCapacity objects")
-	operationTimeout     = flag.Duration("timeout", 10*time.Second, "Timeout for waiting for creation or deletion of a volume")
+	operationTimeout     = flag.Duration("timeout", 10*time.Second, "Timeout for waiting for volume operation (creation, deletion, capacity queries)")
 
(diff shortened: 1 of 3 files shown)


cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -475,6 +475,7 @@ func main() {
 			factoryForNamespace.Storage().V1beta1().CSIStorageCapacities(),
 			*capacityPollInterval,
 			*capacityImmediateBinding,
+			*operationTimeout,
```
pohly, line 478: Setting some timeout makes sense, I just wonder whether we should use `operationTimeout` for it. It's currently defined as
```
        operationTimeout     = flag.Duration("timeout", 10*time.Second, "Timeout for waiting for creation or deletion of a volume")                                                                                                                                                                                                                                                                    
```

Perhaps change that into `Timeout for volume operations (creation, deletion. capacity queries)`?

bai3shuo4: Creation and deletion all use same timeoutOperation. I think we should keep this pattern. We don't need define too much timeout which is unnecessary

pohly: I agree, but then the description should be changed to include the new usage. Querying capacity is not "creation or deletion of a volume" (current description).

bai3shuo4: Oh！I miss it, I will add it

bai3shuo4: @pohly done~

pohly (approved):
/lgtm

bai3shuo4:
@pohly Sorry, I fix the unit test problem, why all this check stales here? How to recover it

bai3shuo4:
> Can you add a release note?
Done

bai3shuo4:
@pohly Hi, I think it's ok to merge :)

pohly:
/lgtm
/assign @xing-yang 

For approval.

PR #696, opened 2022-01-25
maximumVolumeSize should be updated when it changes

mrpre:
**What type of PR is this?**
>
> /kind api-change

 /kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

`maximumVolumeSize` should be updated but not. 
NONE

NONE
NONE
```release-note
bug fix, no release-note
```

### DIFF
--- a/pkg/capacity/capacity.go
+++ b/pkg/capacity/capacity.go
@@ -642,8 +642,9 @@ func (c *Controller) syncCapacity(ctx context.Context, item workItem) error {
 		// scenario that we end up creating two objects for the same work item, the second
 		// one will be recognized as duplicate and get deleted again once we receive it.
 	} else if capacity.Capacity.Value() == quantity.Value() &&
+		sizesAreEqual(capacity.MaximumVolumeSize, maximumVolumeSize) &&
 		(c.owner == nil || c.isOwnedByUs(capacity)) {
-		klog.V(5).Infof("Capacity Controller: no need to update %s for %+v, same capacity %v and correct owner", capacity.Name, item, quantity)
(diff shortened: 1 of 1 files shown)


pohly:
Good catch. Thanks for sharing. Can you also sign the CLA?

/ok-to-test

mrpre:
CLA  signed and all tests passed @pohly @humblec @xing-yang

humblec:
Thanks @mrpre .. lgtm..

pkg/capacity/capacity.go
```diff
@@ -642,8 +642,10 @@ func (c *Controller) syncCapacity(ctx context.Context, item workItem) error {
 		// scenario that we end up creating two objects for the same work item, the second
 		// one will be recognized as duplicate and get deleted again once we receive it.
 	} else if capacity.Capacity.Value() == quantity.Value() &&
+		capacity.MaximumVolumeSize == maximumVolumeSize &&
+		(maximumVolumeSize == nil || capacity.MaximumVolumeSize.Value() == maximumVolumeSize.Value()) &&
```
pohly, line 646: This gets hard to review for correctness.

Let's move this into a helper function:
```
func sizesAreEqual(a, b *resource.Quantity) bool {
   if a == b {
     // Both nil or the same value.
     return true
  }
  if a == nil {
     // b must be non-nil.
     return false
  }
  // Both not nil, compare values.
  return a.Value() == b.Value()
}
```

pohly (approved):
/lgtm
/assign @jsafrane 

For approval.

PR #699, opened 2022-01-26
relax same storageclass restriction

amacaskill:
**What type of PR is this?**
> /kind feature

The restriction that cloning is only supported within the same Storage Class, restricts PVC cloning in some cases. For example, if you want to clone from a zonal volume --> regional volume, the replication-type for these two volumes must be different, so they require different Storage Classes. Currently snapshot allows for restoring from a snapshot with a different StorageClass, and it leaves the validation up to the CSI Drivers. No validation is done in the external provisioner to make sure the Storage Classes are compatible. For PVC cloning, validation will also be up to the CSI Drivers.

Fixes https://github.com/kubernetes-csi/docs/issues/490

```release-note
Users will be able to clone a volume with a different Storage Class. The destination volume no longer has to be the same storage class as the source. 
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -963,11 +963,6 @@ func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentV
 		return nil, fmt.Errorf("the requested PVC (%s) storageclass cannot be empty", claim.Name)
 	}
 
-	if *sourcePVC.Spec.StorageClassName != *claim.Spec.StorageClassName {
-		return nil, fmt.Errorf("the source PVC and destination PVCs must be in the same storage class for cloning.  Source is in %v, but new PVC is in %v",
-			*sourcePVC.Spec.StorageClassName, *claim.Spec.StorageClassName)
-	}
-
 	capacity := claim.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
 	requestedSize := capacity.Value()
 	srcCapacity := sourcePVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -4148,9 +4148,10 @@ func TestProvisionFromPVC(t *testing.T) {
 			expectErr:        true,
 		},
 		"provision with pvc data source different storage classes": {
-			clonePVName: pvName,
-			volOpts:     generatePVCForProvisionFromPVC(srcNamespace, srcName, fakeSc2, requestedBytes, ""),
-			expectErr:   true,
+			clonePVName:      pvName,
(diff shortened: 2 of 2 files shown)


mattcary:
/ok-to-test
/lgtm

An approver in this repo should look at this too to make sure we haven't misunderstood the fix.

pkg/controller/controller.go
```diff
@@ -963,11 +963,6 @@ func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentV
 		return nil, fmt.Errorf("the requested PVC (%s) storageclass cannot be empty", claim.Name)
 	}
 
-	if *sourcePVC.Spec.StorageClassName != *claim.Spec.StorageClassName {
```
msau42, line 966: Should we add some validation that known first class attributes like fstype are compatible?

mattcary: I thought the consensus was that things like that were driver specific and should be handled by the, eg, pd or filestore driver (where the test might be different).

Even though snapshots can't validate anything, the only problems we've seen are on the other side where the over-validation of enforcing storageclasses to be identical prevent cloning. So I'm inclined to have more relaxed validation which ultimately unblocks workloads.

msau42: fstype is a first class field in Kubernetes/CSI, and it's information that cannot be easily obtained in the current CSI CreateVolume call since it's not an attribute that block storage really cares about.

amacaskill: So to summarize, we should validate Fstype in the provisioner because right now we don't have way to validate that the fstype of the source block volume is compatible with the fstype of the destination block volume. `createVolumeRequest.GetVolumeCapability().GetBlock()` doesn't have any field `FsType`.  I believe we have a way to validate fstype for Filesystem volumes, `createVolumeRequest.GetVolumeCapability().GetMount().FsType`, correct?. In the Kubernetes Sig storage meeting, someone threw out the idea of adding documentation to say that some params like fstype may not be compatible if you have a different storage class. Could we do this instead of validating?

jsafrane: >  I believe we have a way to validate fstype for Filesystem volumes, `createVolumeRequest.GetVolumeCapability().GetMount().FsType`, correct

Yes, a CSI driver can get the destination fsType as you suggest above. But it has no way to get the source volume fsType, it gets only `volume_id` of the source volume / `shapshot_id` of the source snapshot.

I don't have a strong opinion about this PR. If the external-provisioner can't check fsType when restoring snapshots (they don't have fsType), why would it check it in cloning? Should we add fsType to snapshots?

msau42: Sounds good to me. My only ask is we make sure cloning and snapshot documentation has a note that the user needs to make sure source and target parameters are compatible.

xing-yang: +1

PR #705, opened 2022-02-18
capacity tracking: add benchmark for topology discovery

### LINKED ISSUE #466: storage capacity: scalability of pkg/capacity/topology/nodes.go
For the sake of simplicity, nodes.go uses slice to store all known segments and iterates over all of them to determine whether a new one has to be added.

We need:
- a test benchmark which scales this up to a high number of nodes
- if the slice turns out to be slow, a better data structure:
  - A tree with topology key/value pairs on the edges and the segment with those pairs in the leaf.
  - Serialize the slice in a string, then use that string as key in a map.

A map of segments does not work because a segment is a slice, and slices cannot be used as key in a map.

pohly:
**What type of PR is this?**

/kind cleanup

When external-provisioner runs in central mode, it needs to collect topology
labels from CSINode and Node to determine which topology segments exist in the
cluster. This benchmark measures how well the code for that works for different
combinations of cluster size and number of segments (determined by the number
of nodes per segment). It measures that for the initial segment creation and
for a refresh where nothing changes.

benchstat shows that the numbers are stable (± 3%). The following script can be
used to turn the output of benchstat into a csv that can be imported into
Google Docs when using the configuration that iterates over node sizes from 0
to 10000 in steps of 100:

```
for t in initial refresh; do echo $t; echo -e "nodes\tsegment size 1\tsegment size 10\tsegment size 100\tsegment size 1000\tsegment size 10000"; for n in $(seq 0 100 10000) ; do echo -n "$n"; for s in 1 10 100 1000 10000; do echo -n -e "\t$(grep numNodes=$n/segmentSize=$s/$t benchstat.out | sed -e 's/[^ ]* *\([^ ]*\) .*/\1/' -e 's/µs//' -e 's/ms/*1000/' | bc -l | tr . , )"; done; echo; done; echo; done >/tmp/data.csv
```

Fixes #466

Here's a Google Doc with the numbers and charts that plot the numbers:
https://docs.google.com/spreadsheets/d/1QzG_FeXm0gZwy7sTfgFMhW7sg9Flr0TfETY8leyG3e8/edit?usp=sharing

My conclusion is that the current code scales well enough (not linear, but doesn't explode either) and doesn't need to be changed.

```release-note
NONE
```

### DIFF
--- a/pkg/capacity/topology/nodes_test.go
+++ b/pkg/capacity/topology/nodes_test.go
@@ -31,7 +31,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
+	coreinformersv1 "k8s.io/client-go/informers/core/v1"
+	storageinformersv1 "k8s.io/client-go/informers/storage/v1"
 	fakeclientset "k8s.io/client-go/kubernetes/fake"
+	corelistersv1 "k8s.io/client-go/listers/core/v1"
+	storagelistersv1 "k8s.io/client-go/listers/storage/v1"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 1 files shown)


ggriffiths (approved):
/lgtm

pohly:
/assign @msau42

For approval.

PR #706, opened 2022-02-18
capacity: add benchmark for CSIStorageCapacity update

### LINKED ISSUE #467: storage capacity: patch vs. update of CSIStorageCapacity
capacity.go currently uses `Update` to change the capacity value. It might be more efficient to use `Patch`:
- write a benchmark
- compare both approaches

pohly:
**What type of PR is this?**
/kind cleanup

It may be possible to optimize updating capacity by using patching. Before
trying that we need a benchmark.

Fixes #467

This benchmark acts as a client with basically unlimited local rate
limits. When run long enough, the results are stable:

```
$ go test -run=xxx -bench=. -count=5 -benchtime=1m . | tee /tmp/log
...
$ $GOPATH/bin/benchstat /tmp/log
name               time/op
CapacityUpdate-36  3.93ms ± 1%
```

```release-note
NONE
```

### DIFF
--- a/pkg/capacity/README.md
+++ b/pkg/capacity/README.md
@@ -1,0 +1,9 @@
+To benchmark the CSIStorageCapacity update code, use:
+
+```
+KUBECONFIG=<some config file> go test -bench=. -run=xxx .
+```
+
+Running repeatedly with `-count=5` and filtering the output with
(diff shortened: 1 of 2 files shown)


pohly:
There doesn't seem to be enough interest in having these benchmarks included in the repo.

/close

PR #711, opened 2022-03-07
capacity: use separate client and lower QPS/Burst settings

pohly:
**What type of PR is this?**
/kind feature

The rationale behind this change is that it allows external-provisioner to deal
with volumes without being slowed down by the concurrent CSIStorageCapacity
updates. Those updates are less important and the scheduler can deal with
outdated information, so enforcing a lower rate by default is acceptable.

If few changes are needed, updating CSIStorageCapacity keeps up with volume
changes. If many volumes get created or deleted quickly, throttling the
updating is beneficial because multiple changes can be combined into a single
update.

This has helped in practice during a scale test with 100 nodes and 10000
pods+volumes. Without this change, there were intermittent problems with an
unresponsive apiserver. With this change, no such problems occurred.

```release-note
QPS and burst can be set separately for storage capacity tracking. The defaults are lower than before to reduce the risk of overloading the apiserver.
```


pohly:
/hold

I'm currently rerunning the scale test, let's wait with merging until that is completed.

pohly:
/hold cancel

The test completed two more times with this change.

PR #713, opened 2022-03-10
controller: add pv secret annotation support to forget SC

### LINKED ISSUE #330: Improve Deletion Secret Handling
Today, the external-provisioner will look at StorageClass for deletion secret. If the secret does not exist, delete is called any way. 

The problem is that the StorageClass may be deleted or mutated (deleted and recreated with different parameters). This will result in volumes being unable to delete. Ideally we want the deletion secret on the PV object. However, @liggitt pointed out that this would result in asymmetry of the API (provision is handled by a higher layer controller and specified in StorageClass, having that controller then look at PV object for the delete operation seems wrong). That said, we do want to better handle this case.

So as a compromise, the proposal is to 1) add a reference to the deletion secret as an annotation on the PV object (instead of a first class field), and 2) to better document why you shouldn't have deletion secrets.

For 1, the *proposed change* is to add a new flag to external-provisioner that says controller requires deletion secret, if this is set by SP, the external-provisioner should store a reference to the provision secret in an annotation on the PV object, and when deleting if the flag is set and the PV object has the annotation, fetch and pass the secret in the CSI `DeleteSecret` call.

And 2 is being tacked in https://github.com/kubernetes-csi/docs/issues/189#issuecomment-520621418.

humblec:

> /kind feature
Fixes #330 

-->
```release-note
Previous to this release, if the SC does not exist at time of PVC deletion where provisioner need credentials to delete volume which was passed as SC parameter at creation time, the PV deletion will fail as it can not fetch credentials by reading the SC from the API server. However this workflow is improved from this release which allows the PV deletion to succeed even if SC was deleted or does not exist at time of PV deletion. 
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1217,7 +1217,7 @@ func (p *csiProvisioner) handleSecretsForDeletion(ctx context.Context, volume *v
 			}
 			req.Secrets = credentials
 		} else if annDeletionSecretName == "" && annDeletionSecretNamespace == "" {
-			klog.V(2).Infof("volumes without secrets detected.. continue deleting the volumes")
+			klog.V(2).Infof("volume %s does not need any deletion secrets", volume.Name)
 		}
 	} else {
 		err := p.getSecretsFromSC(ctx, volume, migratedVolume, req)
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -132,6 +132,11 @@ const (
 	annStorageProvisioner     = "volume.kubernetes.io/storage-provisioner"
 	annSelectedNode           = "volume.kubernetes.io/selected-node"
 
+	// Annotation for secret name and namespace will be added to the pv object
+	// and used at pvc deletion time.
+	annDeletionProvisionerSecretRefName      = "volume.kubernetes.io/provisioner-deletion-secret-name"
+	annDeletionProvisionerSecretRefNamespace = "volume.kubernetes.io/provisioner-deletion-secret-namespace"
+
 	snapshotNotBound = "snapshot %s not bound"
 
 	pvcCloneFinalizer = "provisioner.storage.kubernetes.io/cloning-protection"
@@ -497,11 +502,17 @@ func (p *csiProvisioner) getVolumeCapabilities(
 	return volumeCaps, nil
 }
 
+type deletionSecretParams struct {
+	name      string
+	namespace string
+}
+
 type prepareProvisionResult struct {
-	fsType         string
-	migratedVolume bool
-	req            *csi.CreateVolumeRequest
-	csiPVSource    *v1.CSIPersistentVolumeSource
+	fsType              string
+	migratedVolume      bool
+	req                 *csi.CreateVolumeRequest
+	csiPVSource         *v1.CSIPersistentVolumeSource
+	provDeletionSecrets *deletionSecretParams
 }
 
 // prepareProvision does non-destructive parameter checking and preparations for provisioning a volume.
@@ -689,13 +700,21 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		req.Parameters[pvcNamespaceKey] = claim.GetNamespace()
 		req.Parameters[pvNameKey] = pvName
 	}
+	deletionAnnSecrets := new(deletionSecretParams)
+
+	if provisionerSecretRef != nil {
+		deletionAnnSecrets.name = provisionerSecretRef.Name
+		deletionAnnSecrets.namespace = provisionerSecretRef.Namespace
+	}
 
 	return &prepareProvisionResult{
-		fsType:         fsType,
-		migratedVolume: migratedVolume,
-		req:            &req,
-		csiPVSource:    csiPVSource,
+		fsType:              fsType,
+		migratedVolume:      migratedVolume,
+		req:                 &req,
+		csiPVSource:         csiPVSource,
+		provDeletionSecrets: deletionAnnSecrets,
 	}, controller.ProvisioningNoChange, nil
+
 }
 
 func (p *csiProvisioner) Provision(ctx context.Context, options controller.ProvisionOptions) (*v1.PersistentVolume, controller.ProvisioningState, error) {
@@ -836,6 +855,16 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 				CSI: result.csiPVSource,
 			},
 		},
+	}
+
+	// Set annDeletionSecretRefName and namespace in PV object.
+	if result.provDeletionSecrets != nil {
+		klog.V(5).Infof("createVolumeOperation: set annotation [%s/%s] on pv [%s].", annDeletionProvisionerSecretRefNamespace, annDeletionProvisionerSecretRefName, pv.Name)
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, annDeletionProvisionerSecretRefName, result.provDeletionSecrets.name)
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, annDeletionProvisionerSecretRefNamespace, result.provDeletionSecrets.namespace)
+	} else {
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, annDeletionProvisionerSecretRefName, "")
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, annDeletionProvisionerSecretRefNamespace, "")
 	}
 
 	if options.StorageClass.ReclaimPolicy != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -838,6 +857,14 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 		},
 	}
 
+	// Set AnnDeletionSecretRefName and namespace
+
+	if result.provDeletionSecrets != nil {
```
mkimuram, line 862: @humblec 

Should we consider adding empty string for "volume.kubernetes.io/provisioner-deletion-secret-name" and "volume.kubernetes.io/provisioner-deletion-secret-namespace" annotations if Secret is not used (and making `Delete` not fail and skip adding credentials in this case)?

The current logic falls back to get name and namespace for Secret from StorageClass if these annotations are not found. However, it can't distinguish whether lack of annotations is caused by (1) no Secret required or (2) provisioned by old provisioner. As a result, volumes without Secret will fail to be deleted when StorageClass is not found, even after new provisioner is used on deploy.

humblec: Nice catch @mkimuram ! , updated to accomodate that condition too..

mkimuram:
@humblec I've created unit tests for this PR https://github.com/mkimuram/external-provisioner/commit/f0b88ab1868ffaaa3dd409a650caf83c0a615f80 on top of this PR's commit. Please check it. 
Also, please let me know your idea on how we collaborate, could you merge my test codes to this PR or should I create another PR?

humblec:
> @humblec I've created unit tests for this PR [mkimuram@f0b88ab](https://github.com/mkimuram/external-provisioner/commit/f0b88ab1868ffaaa3dd409a650caf83c0a615f80) on top of this PR's commit. Please check it. Also, please let me know your idea on how we collaborate, could you merge my test codes to this PR or should I create another PR?

@mkimuram Thats cool. I have added your commit on top of this PR :+1: , lets see how it goes!

pkg/controller/controller.go
```diff
@@ -132,6 +132,11 @@ const (
 	annStorageProvisioner     = "volume.kubernetes.io/storage-provisioner"
 	annSelectedNode           = "volume.kubernetes.io/selected-node"
 
+	// Annotation for secret name and namespace will be added to the pv object
+	// and used at pvc deletion time.
+	AnnDeletionProvisionerSecretRefName      = "volume.kubernetes.io/provisioner-deletion-secret-name"
+	AnnDeletionProvisionerSecretRefNamespace = "volume.kubernetes.io/provisioner-deletion-secret-namespace"
```
jsafrane, line 138: Why is it exported outside of the package?

pkg/controller/controller.go
```diff
@@ -1154,37 +1183,56 @@ func (p *csiProvisioner) Delete(ctx context.Context, volume *v1.PersistentVolume
...
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      volume.Spec.ClaimRef.Name,
-					Namespace: volume.Spec.ClaimRef.Namespace,
-				},
-			})
-			if err != nil {
-				return fmt.Errorf("failed to get secretreference for volume %s: %v", volume.Name, err)
-			}
 
+	if metav1.HasAnnotation(volume.ObjectMeta, AnnDeletionProvisionerSecretRefName) && metav1.HasAnnotation(volume.ObjectMeta, AnnDeletionProvisionerSecretRefNamespace) {
```
jsafrane, line 1187: This prefers annotations over StorageClass even when the StorageClass still exists. Is this intended behavior? What if the StorageClass was re-created with updated secrets?

pkg/controller/controller.go
```diff
@@ -1154,37 +1183,56 @@ func (p *csiProvisioner) Delete(ctx context.Context, volume *v1.PersistentVolume
...
+		annDeletionSecretNamespace := volume.Annotations[AnnDeletionProvisionerSecretRefNamespace]
+		if annDeletionSecretName != "" && annDeletionSecretNamespace != "" {
+			provisionerSecretRef := &v1.SecretReference{}
+			provisionerSecretRef.Name = annDeletionSecretName
+			provisionerSecretRef.Namespace = annDeletionSecretNamespace
 			credentials, err := getCredentials(ctx, p.client, provisionerSecretRef)
 			if err != nil {
 				// Continue with deletion, as the secret may have already been deleted.
-				klog.Errorf("Failed to get credentials for volume %s: %s", volume.Name, err.Error())
+				klog.Errorf("failed to get credentials for volume %s:%s", volume.Name, err.Error())
```
jsafrane, line 1197: Please don't remove the space

pkg/controller/controller.go
```diff
@@ -1154,37 +1183,56 @@ func (p *csiProvisioner) Delete(ctx context.Context, volume *v1.PersistentVolume
 	req := csi.DeleteVolumeRequest{
 		VolumeId: volumeId,
 	}
-	// get secrets if StorageClass specifies it
```
jsafrane, line 1157: IMO the whole Secrets handling now deserves its own function, Delete() has grown too much.

pkg/controller/controller_test.go
```diff
@@ -3876,6 +3925,206 @@ func TestDelete(t *testing.T) {
...
+					},
+				},
+			},
+			storageClass: &storagev1.StorageClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "sc-name",
+				},
+				Parameters: map[string]string{
+					prefixedProvisionerSecretNameKey:      "provisionersecret",
+					prefixedProvisionerSecretNamespaceKey: defaultSecretNsName,
```
jsafrane, line 3999: Use a different key-value par than `getDefaultProvisinerSecrets()` to see where the final secret comes from. This applies to all tests below.

mkimuram: @jsafrane  Thank you for your review.

`Non-empty provisioner secret is set as PV annotation and modified StorageClass exists` and `Non-existent provisioner secret is set as PV annotation` already checks where the final secret comes from.
Former one has invalid secret specified in the StorageClass and verifies that the right secret is taken from annotations (we may still need to discuss if it should be the expected behavior as commented [here](https://github.com/kubernetes-csi/external-provisioner/pull/713#discussion_r829110566)).
Latter one has invalid secret specified in the annotations and verifies that it fails to get the secret. 

However, we may be able to improve `getDefaultProvisinerSecrets()` to prepare two secrets and check that either of the expected secret is set, to make it clear.
(Also, `getDefaultProvisinerSecrets()` should be renamed to something like `getProvisinerSecretCandidates()`?)

pkg/controller/controller.go
mkimuram: I think that other `Secret`s are stored in `csiSource` on [provision](https://github.com/kubernetes-csi/external-provisioner/blob/32c27a57ce311cf7b0fedefd6cc850845a41e341/pkg/controller/controller.go#L672-L679) and won't be changed when `StorageClass` is updated, so the same behavior, prioritizing the value on provision, would be also expected for provisioner-secret.

Although, we might still be able to discuss if the current behavior of the other `Secret`s is good or not.

pkg/controller/controller.go
humblec: unexported

pkg/controller/controller.go
humblec: sure :)

pkg/controller/controller.go
humblec: yeah, considering we give preference to the spec source once PV is bound, I believe SC recreation should also get less preference here.

pkg/controller/controller.go
humblec: sure, moved into a seperate funcs.

ggriffiths (commented):
Mostly looks good, just one question for the provision changes

pkg/controller/controller.go
```diff
@@ -838,6 +857,16 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
...
 	}
 
+	// Set annDeletionSecretRefName and namespace in PV object.
+	if result.provDeletionSecrets != nil {
+		klog.V(5).Infof("createVolumeOperation: set annotation [%s/%s] on pv [%s].", annDeletionProvisionerSecretRefNamespace, annDeletionProvisionerSecretRefName, pv.Name)
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, annDeletionProvisionerSecretRefName, result.provDeletionSecrets.name)
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, annDeletionProvisionerSecretRefNamespace, result.provDeletionSecrets.namespace)
+	} else {
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, annDeletionProvisionerSecretRefName, "")
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, annDeletionProvisionerSecretRefNamespace, "")
```
ggriffiths, line 867: Is this needed? Won't the annotation already be unset if this is a new PV?

mkimuram: Setting "" to the annotations is needed for deletion path. If we don't do it here, or on provision, and these annotations are not found on deletion, the controller can't distinguish:
1. The volume doesn't require Secret for deletion
2. The volume requires Secret for deletion but provisioned with old provisioner that doesn't add the annotations

As a result, deletion of the case 1 fails if SC is deleted. By setting "", we can avoid it like [this](https://github.com/kubernetes-csi/external-provisioner/pull/713/files/d6f20f85c31b3ea94c5a0612a344a8928ce7162f#diff-243ebed2765f75e6a54f57167212fefb08c3b2a85967ad2acbc0eb78919019c1R1219-R1221).

pkg/controller/controller.go
jsafrane: ack

pkg/controller/controller_test.go
humblec: @jsafrane could you please confirm/share your thought on above , so that the unit test can also be adjusted..

pkg/controller/controller.go
humblec: Thanks @mkimuram . 
@ggriffiths hope above clarifies it..

pkg/controller/controller.go
```diff
@@ -1154,6 +1183,52 @@ func (p *csiProvisioner) Delete(ctx context.Context, volume *v1.PersistentVolume
...
+			provisionerSecretRef.Name = annDeletionSecretName
+			provisionerSecretRef.Namespace = annDeletionSecretNamespace
+			credentials, err := getCredentials(ctx, p.client, provisionerSecretRef)
+			if err != nil {
+				// Continue with deletion, as the secret may have already been deleted.
+				klog.Errorf("failed to get credentials for volume %s: %s", volume.Name, err.Error())
+			}
+			req.Secrets = credentials
+		} else if annDeletionSecretName == "" && annDeletionSecretNamespace == "" {
+			klog.V(2).Infof("volumes without secrets detected.. continue deleting the volumes")
```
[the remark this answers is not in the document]
humblec: Done.

pkg/controller/controller.go
ggriffiths: Ah yes, meant to reply. Thanks for the clarification!

PR #717, opened 2022-03-21
fix managed-by label being too long when the node name is long.

### LINKED ISSUE #707: distributed provisioner: managed-by label can fail if node name too long
The value of the `csi.storage.k8s.io/managed-by` label when the external provisioner had enableNodeDeployment set can be too large if the node name is too long. The value is calculated [here](https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L447-L450) and then passed to the capacity controller. The capacity controller sets that value as a label on the CsiStorageCapacity object as is [here](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/capacity/capacity.go#L622). Labels have a maximum length of 63 characters and if the node name is long enough, combined with the `external-provisioner` prefix on the label value, it will be too large and the lister will fail.

My suggestion would be to hash the value and put that as the label value, and then use the same hash function in the [check](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/capacity/capacity.go#L833) that confirms the right managed by label is set.

awels:
Added a hashing function that only hashes the passed in base and
prefix if the combined size is > the maxLength passed into the function
In this case the maximum length of a label.

Signed-off-by: Alexander Wels <awels@redhat.com>

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #707 

```release-note
storage capacity: managed-by label potentially too long with a long node name and enableNodeDeployment set to true
```

### DIFF
--- a/cmd/csi-provisioner/util.go
+++ b/cmd/csi-provisioner/util.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"hash/fnv"
+)
+
+// getNameWithMaxLength returns a name given a base ("deployment-5") and a suffix ("deploy")
+// It will first attempt to join them with a dash. If the resulting name is longer
+// than maxLength: if the suffix is too long, it will truncate the base name and add
+// an 8-character hash of the [base]-[suffix] string.  If the suffix is not too long,
+// it will truncate the base, add the hash of the base and return [base]-[hash]-[suffix]
+// copied from https://github.com/openshift/library-go/blob/master/pkg/build/naming/namer.go
+func getNameWithMaxLength(base, suffix string, maxLength int) string {
+	name := fmt.Sprintf("%s-%s", base, suffix)
+	if len(name) <= maxLength {
+		return name
+	}
+
+	baseLength := maxLength - 10 /*length of -hash-*/ - len(suffix)
+
+	// if the suffix is too long, ignore it
+	if baseLength < 0 {
+		prefix := base[0:min(len(base), max(0, maxLength-9))]
+		// Calculate hash on initial base-suffix string
+		shortName := fmt.Sprintf("%s-%s", prefix, hash(name))
+		return shortName[:min(maxLength, len(shortName))]
+	}
+
+	prefix := base[0:baseLength]
+	// Calculate hash on initial base-suffix string
+	return fmt.Sprintf("%s-%s-%s", prefix, hash(base), suffix)
+}
+
+// hash calculates the hexadecimal representation (8-chars)
+// of the hash of the passed in string using the FNV-a algorithm
+func hash(s string) string {
+	hash := fnv.New32a()
+	hash.Write([]byte(s))
+	intHash := hash.Sum32()
+	result := fmt.Sprintf("%08x", intHash)
+	return result
+}
+
+// max returns the greater of its 2 inputs
+func max(a, b int) int {
+	if b > a {
+		return b
+	}
+	return a
+}
+
+// min returns the lesser of its 2 inputs
+func min(a, b int) int {
+	if b < a {
+		return b
+	}
+	return a
+}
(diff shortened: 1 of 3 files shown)


awels:
I would like to add some functional testing, but I can't find any in this repo. Do I need to add some tests somewhere else?

pohly:
With functional testing you mean running inside a cluster? That is covered by the Prow jobs which deploy this sidecar as part of the csi-driver-host-path deployment. There is one job for "distributed" which enables storage capacity tracking.

/ok-to-test

pohly:
Thanks for the PR, BTW. Looks good overall. Can you replace "BugFix: " with "storage capacity: " in the change note? The kind of change is covered by the PR label.

cmd/csi-provisioner/util.go
```diff
@@ -0,0 +1,70 @@
...
+*/
+
+package main
+
+import (
+	"fmt"
+	"hash/fnv"
+)
+
+// copied from https://github.com/openshift/library-go/blob/master/pkg/build/naming/namer.go
```
pohly, line 24: Can you copy the full function description from https://github.com/openshift/library-go/blob/edab248e63516c65a93467eaa8224c86d69f5de9/pkg/build/naming/namer.go#L10-L14?

pohly: Actually, using https://github.com/openshift/library-go/blob/master/pkg/build/naming/namer.go directly as a dependency may be better. The module doesn't get released (no tags), but it has a compatible license and we only need that one package from it.

awels: Yeah, I was afraid I would pull in a large part of Open Shift just for that one function, which I would like to avoid. Let me see what happens.

awels:
Well I would like to write some tests specifically to test the long node names, which I am guessing is currently not covered by the existing tests. But I did not find any location I could add tests, unless I need to do that in the kubernetes project itself.

cmd/csi-provisioner/util.go
pohly: Looks good to me. As expected, changes to go.mod and vendor are minimal. What probably becomes slower are the "go mod" operations when github.com/openshift/library-go hasn't been retrieved yet, but that shouldn't be a concern.

Let's get a second opinion.... @jsafrane: should we fork github.com/openshift/library-go/pkg/build/naming or pull it in as dependency?

awels: @jsafrane Could you take a look?

jsafrane: openshift/library-go does not guarantee stable API and it does not have tags / releases. While we pull just a small package that we probably won't ever need to re-vendor, I am worried by dumb CVE scanning tools that may catch "unsafe" library-go in go.sum, when the unsafe functionality is actually never used.

I would prefer to copy the function, but I don't have strong opinion.

awels: Since we already have both versions it is easy enough to just copy, essentially it will be this https://github.com/kubernetes-csi/external-provisioner/pull/717/commits/995d58b079d6c5958668232e44606f046f2d67b3 but with the format fixes.

pohly: Let's copy the function.

awels: Okay give me a bit to update the PR, I am basically going to revert the last commit, and then fix the small formatting issue

pohly:
> Well I would like to write some tests specifically to test the long node names, which I am guessing is currently not covered by the existing tests. But I did not find any location I could add tests, unless I need to do that in the kubernetes project itself.

You would have to bring up a cluster with long node names. I suppose the KinD cluster that is used for testing might support that, but if we change that in the one job that we have testing, then we loose coverage of the case with short names.

I suspect the unit tests will have to be sufficient for test coverage of this code.

awels:
Okay, I am somewhat obsessive on writing functional tests for stuff, but I think the unit test proves that if you call the function with a longer name, it will properly modify the name to be both consistent, and of appropriate size. Plus once it makes it into my QE organization runs, they have node names with long lengths (this is how we found the problem) that will prove it works.

cmd/csi-provisioner/util.go
pohly: As we are back at copying the function, can you address my first comment (copy the function comment, too)?

awels: Ahhh of course, sorry about that.

pohly (approved):
Squash the commits and we are good to go, eh, merge.

awels:
Squashed into single commit.

pohly (approved):
/lgtm
/assign @jsafrane 

For approval.

maya-r:
Hi all, do you have any idea when one could expect this commit to be in a tagged release? thanks.

pohly:
We will prepare new sidecars for Kubernetes 1.25. external-provisioner should become available shortly after Kubernetes 1.25.0 because we depend on client-go from 1.25.

akalenyu:
> We will prepare new sidecars for Kubernetes 1.25. external-provisioner should become available shortly after Kubernetes 1.25.0 because we depend on client-go from 1.25.

Hi
Would it make sense to backport this to `release-3.1`?
IIUC any setup with long node names is going to cause this problem

pohly:
Sounds reasonable. Can you prepare cherry-pick PRs?

akalenyu:
> Sounds reasonable. Can you prepare cherry-pick PRs?

Done, https://github.com/kubernetes-csi/external-provisioner/pull/739
Should this also be in releases < 3?

pohly:
I think we'll drop support for 2.1.x soon, but 2.2.x might get another update. If you prepare a PR, then it'll get included.

PR #729, opened 2022-04-24
controller: pass sc information to downstream csi

krisdiano:
**What type of PR is this?**

> /kind feature

When developing localpv, it is necessary to record some meta information, such as the last used disk, so that the disk allocation strategy can be based on the meta information, which may be a rotation training strategy, but users may create different scs based on the same csi, and csi cannot Distinguish the sc corresponding to pv, so the strategy is problematic.

NONE

NONE

```release-note
external-privisioner As a controller, it should pass some information to the csi information, so that the csi can be refined for the user.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -140,6 +140,10 @@ const (
 	snapshotNotBound = "snapshot %s not bound"
 
 	pvcCloneFinalizer = "provisioner.storage.kubernetes.io/cloning-protection"
+
+	// The kv that needs to be automatically inserted into CreateVolumeRequest.Parameters
+	parameterPrefix = "X-K8S-PROVISIONER-"
+	ParameterScName = parameterPrefix + "SC-NAME"
 )
 
 var (
@@ -613,6 +617,10 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 			RequiredBytes: int64(volSizeBytes),
 		},
 	}
+	if req.Parameters == nil {
+		req.Parameters = make(map[string]string)
+	}
+	req.Parameters[ParameterScName] = sc.Name
 
 	if claim.Spec.DataSource != nil && (rc.clone || rc.snapshot) {
 		volumeContentSource, err := p.getVolumeContentSource(ctx, claim, sc)


pkg/controller/controller.go
```diff
@@ -140,6 +140,10 @@ const (
 	snapshotNotBound = "snapshot %s not bound"
 
 	pvcCloneFinalizer = "provisioner.storage.kubernetes.io/cloning-protection"
+
+	// The kv that needs to be automatically inserted into CreateVolumeRequest.Parameters
+	parameterPrefix = "X-K8S-PROVISIONER-"
+	ParameterScName = parameterPrefix + "SC-NAME"
```
Madhu-1, line 146: if you want the SC name to be passed to the CSI driver, add one more parameter to the storageclass `kubernetes-sc-name: sc-name` and it will get automatically passed to the CSI driver?

awels: So there are parameters you can pass to the SC, and those parameters are passed to the CSI provisioner. I think that would solve your issue?

krisdiano: > if you want the SC name to be passed to the CSI driver, add one more parameter to the storageclass `kubernetes-sc-name: sc-name` and it will get automatically passed to the CSI driver?

in perpareProvision function.

krisdiano: > So there are parameters you can pass to the SC, and those parameters are passed to the CSI provisioner. I think that would solve your issue?

yes，prepareProvision has sc information，the parameters are passed to the sci，so I define a constant key here and add sc name to the parameters behind.

krisdiano: > So there are parameters you can pass to the SC, and those parameters are passed to the CSI provisioner. I think that would solve your issue?

I got it, you mean I add it to sc when created, I did that before, the problem is `ParameterScName`  is required,  If it does not exist, I am expecting an error when creating，because delayed error messages are not user friendly, and users will be confused about this parameter, so I did it in external-provisioner.

krisdiano: yap I think so. But many coworkers didn't think so. In the end, we maintain our external provisioner forked here.
Thanks, It can be closed.

humblec:
Closing this as per https://github.com/kubernetes-csi/external-provisioner/pull/729#discussion_r859805116

PR #730, opened 2022-04-30
Fix DeleteVolume() timeout for cleanupVolume()

xinydev:
**What type of PR is this?**

/kind bug

Every `DeleteVolume()` request needs a same `connectionTimeout`, not share a total `connectionTimeout` among all `DeleteVolume()` requests. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1811,10 +1811,10 @@ func checkError(err error, mayReschedule bool) controller.ProvisioningState {
 func cleanupVolume(ctx context.Context, p *csiProvisioner, delReq *csi.DeleteVolumeRequest, provisionerCredentials map[string]string) error {
 	var err error
 	delReq.Secrets = provisionerCredentials
-	deleteCtx, cancel := context.WithTimeout(ctx, p.timeout)
-	defer cancel()
 	for i := 0; i < deleteVolumeRetryCount; i++ {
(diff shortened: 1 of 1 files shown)


pohly:
Can you explain why this change in the logic is better?

xinydev:
> Can you explain why this change in the logic is better?

Now if the first `DeleteVolume()` times out, the subsequent retry is meaningless, because the `deleteCtx` is not updated, it will definitely time out. This is just my opinion, please correct me if I'm wrong, thanks

pohly:
> Now if the first DeleteVolume() times out, the subsequent retry is meaningless, because the deleteCtx is not updated, it will definitely time out. 

Yes, and perhaps that was intentional to ensure that relatively long timeouts don't add up. The fast retry within the allotted time slot might still be useful for other failures.

I haven't written the code, but to me that also makes sense, therefore this PR is not obviously better.

xinydev:
> > Now if the first DeleteVolume() times out, the subsequent retry is meaningless, because the deleteCtx is not updated, it will definitely time out.
> 
> Yes, and perhaps that was intentional to ensure that relatively long timeouts don't add up. The fast retry within the allotted time slot might still be useful for other failures.
> 
> I haven't written the code, but to me that also makes sense, therefore this PR is not obviously better.

That make sense! Thank you for the explanation.

PR #731, opened 2022-05-02
clone pvc labels to pv labels

andyxning:
**What type of PR is this?**
> /kind feature

Setting pv labels by cloning pvc's labels. We may add some custom labels to pvc, we need these labels also available to pv so we can list pv with label selector.

Fixes #None

```release-note
Setting pv labels by cloning pvc's labels
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -839,7 +839,8 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 	result.csiPVSource.ReadOnly = pvReadOnly
 	pv := &v1.PersistentVolume{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: pvName,
+			Name:   pvName,
+			Labels: claim.ObjectMeta.Labels,
 		},
 		Spec: v1.PersistentVolumeSpec{
 			AccessModes:  options.PVC.Spec.AccessModes,


humblec:
imo, Having all PVC labels on PV may not be ideal, considering the ownership of these objects are different.  however lets get thoughts from others too @msau42 @jsafrane

jsafrane:
I agree with @humblec, PV and PVC is a different object with a different purpose and blindly copying labels from one to the other is IMO bad. For example, copying bad topology labels could make a PV not usable.

At the same time, I see a good use case of copying *some* labels from PVC to PV, controlled by some policy. Just brainstorming an idea: there could be a list of labels to copy in a new StorageClass field.

Another question is *when* to copy these labels. If user adds a new label to a PVC that is already bound to a PV, should this label be copied to the PV too? Or when a new label is added to the policy, e.g. StorageClass suggested above, should it re-sync all existing PVCs/PVs?

This will need broader discussion among sig-storage folks. cc @gnufied @msau42 @xing-yang @saad-ali

andyxning:
I agree that this way may not be a good choice to pass some labels from pvc to pv. All i want to do is we need a way to pass some labels to pv from pvc.

> Another question is when to copy these labels. If user adds a new label to a PVC that is already bound to a PV, should this label be copied to the PV too? Or when a new label is added to the policy, e.g. StorageClass suggested above, should it re-sync all existing PVCs/PVs?

This is not the objective for my initial PR. The coordination of pvc labels and pv labels is not needed, imho. After initial settings of pv labels, the updation of pvc labels does not need to propagate to pv since they are not the relation like deployment and pod.

andyxning:
Friendly Ping @xing-yang @jsafrane @humblec

jsafrane:
It almost feels like there should be pvc.spec.persistentVolumeTemplate, at least with annotations and labels. And perhaps with a way how to transform _some_ PVC fields into PV labels / annotations.

Issue #732, opened 2022-05-02
external-provisioner yields whole stack trace when it loses connection to the CSI driver


ialidzhikov:
/sig storage
/kind bug

**What happened**:
external-provisioner yields whole stack trace when it loses connection to the CSI driver.

**What you expected to happen**:
external-provisioner to only log connection lost and to exit (without logging thousands of lines of stack trace that is not useful for anyone).

**How to reproduce it**:

```
{"log":"2022-05-01T18:20:45.305918834Z stderr F E0501 18:20:45.305666       1 connection.go:131] Lost connection to unix:///var/lib/csi/sockets/pluginproxy/csi.sock."}
{"log":"2022-05-01T18:20:45.306436522Z stderr F F0501 18:20:45.306382       1 connection.go:87] Lost connection to CSI driver, exiting"}
{"log":"2022-05-01T18:20:46.106055473Z stderr F goroutine 82 [running]:"}
{"log":"2022-05-01T18:20:46.106122391Z stderr F k8s.io/klog/v2.stacks(0xc00000e001, 0xc000bec1e0, 0x57, 0x1cd)"}
{"log":"2022-05-01T18:20:46.106166734Z stderr F \t/workspace/vendor/k8s.io/klog/v2/klog.go:1026 +0xb9"}
{"log":"2022-05-01T18:20:46.106178188Z stderr F k8s.io/klog/v2.(*loggingT).output(0x2606640, 0xc000000003, 0x0, 0x0, 0xc000472f50, 0x2554978, 0xd, 0x57, 0x0)"}
{"log":"2022-05-01T18:20:46.106196151Z stderr F \t/workspace/vendor/k8s.io/klog/v2/klog.go:975 +0x19b"}
{"log":"2022-05-01T18:20:46.106206488Z stderr F k8s.io/klog/v2.(*loggingT).printf(0x2606640, 0x3, 0x0, 0x0, 0x0, 0x0, 0x19a688e, 0x26, 0x0, 0x0, ...)"}
{"log":"2022-05-01T18:20:46.106231016Z stderr F \t/workspace/vendor/k8s.io/klog/v2/klog.go:750 +0x191"}
{"log":"2022-05-01T18:20:46.1062379Z stderr F k8s.io/klog/v2.Fatalf(...)"}
{"log":"2022-05-01T18:20:46.10624548Z stderr F \t/workspace/vendor/k8s.io/klog/v2/klog.go:1514"}
[...]

PR #734, opened 2022-05-05
[release-2.2] Update github.com/kubernetes-csi/csi-lib-utils

ialidzhikov:
/sig storage
/kind bug

Part of https://github.com/kubernetes-csi/external-provisioner/issues/732

**Does this PR introduce a user-facing change?**:

```release-note
external-provisioner now exits without output about goroutines when it fails to connect to the CSI driver.
```

PR #735, opened 2022-05-05
[release-2.1] Update github.com/kubernetes-csi/csi-lib-utils

ialidzhikov:
/sig storage
/kind bug

Part of https://github.com/kubernetes-csi/external-provisioner/issues/732

```release-note
external-provisioner now exits without output about goroutines when it fails to connect to the CSI driver.
```


pohly:
Looks like the release-2.1 branch was not prepared to work with Kubernetes 1.23. It's old and will soon be unsupported. Do you really need it to be updated?

ialidzhikov:
We are still using external-provisioner@2.1.2. I would appreciate if it is possible to cut a new patch release from the release-2.1 branch that includes this PR. If it is not possible we have to consider update to external-provisioner@2.2.x.

pohly:
Why can you not update to a newer release?

ialidzhikov:
Ok, I will then update to 2.2.x when a fix containing https://github.com/kubernetes-csi/external-provisioner/pull/734 is cut (2.2.3).

PR #737, opened 2022-05-18
test: add tirvy vulnerability scanner github action

andyzhangx:
**What type of PR is this?**
/kind failing-test

test: add tirvy vulnerability scanner github action

Fixes #

```release-note
none
```

### DIFF
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -1,0 +1,33 @@
+name: Trivy vulnerability scanner
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.16
+      
+      - name: Build an image from Dockerfile
+        run: |
+          make
+          docker build -t test/image:latest -f Dockerfile --output=type=docker --label revision=latest .
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: 'test/image:latest'
+          format: 'table'
+          exit-code: '1'
+          ignore-unfixed: true
+          vuln-type: 'os,library'
+          severity: 'CRITICAL,HIGH,MEDIUM,LOW,UNKNOWN'


pohly:
Same concern as in other PRs which add GitHub actions in specific repos: we should have a consistent policy for all Kubernetes-CSI repos.

mauriciopoppe:
agreed, maybe they could be symlinks from `.github/workflows/<.yaml>` to `release-tools/workflows/<.yaml>`

pohly:
I agree, such symlinks might work (haven't tried it). But before we dive into implementation details I would like to have a discussion about how we use these additional checks. For example, why only check the master branch? Isn't it even more important to check supported release branches? And the elephant in the room: what do we do if such a check fails? Who signs up to deal with it?

yonatankahana:
@pohly 

> I agree, such symlinks might work (haven't tried it)

i have tested that and it seems that github does not resolve links, and the action will [fail to run](https://github.com/yonatankahana/github-actions-symlink/actions/runs/2455646122/workflow)... 

> why only check the master branch? Isn't it even more important to check supported release branches?

it is more important to check PRs.
in my version for `nfs-subdir-external-provisioner` (https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner/pull/211) it test both `master` and PRs. 


> what do we do if such a check fails? Who signs up to deal with it?

the `SECURITY_CONTACTS` (or any other volunteer) can be in charge of creating an issue with the CVE details. they can fix it themselves or tag the issue with the `help wanted` and `kind/failing-test` labels.


there is also a question whether we should fail the test if the CVE's are still 'unfixed'?

pohly:
> it is more important to check PRs.

Why? Because a PR adds a new dependency which is vulnerable? We don't add much new code, so this seems unlikely.

A much more common situation will be that a new vulnerability is detected in an existing dependency, and that then affects both the master branch and all release branches.

andyzhangx:
> What I've seen is that most of the vulnerability reports come from using a base image that is not distroless (e.g. if it's any of the builder images in https://github.com/kubernetes/release/blob/master/images/build/debian-base/variants.yaml), as @pohly says it usually doesn't come from new go code or by adding new dependencies.
> 
> > Isn't it even more important to check supported release branches?
> 
> Yes, I think we could have periodic jobs that use the image scanner on master and previous releases.

@mauriciopoppe This vuln scanner gh action would scan both image and go binary, so even it's distroless base image, there could be still vulnerability issue. Not sure how to make this change into csi release-tools

ConnorJC3:
/remove-lifecycle stale

Is there still interest in this? We at the EBS CSI Driver are also impacted by sidecar vulns.

msau42:
The current github action form is too dirsuptive, which blocks non-related PRs from merging. If there is a more out of band way to scan and open bug reports, then that would be preferable.

pohly:
I saw a comment on #sig-testing that folks are moving away from trivy because it generates too many false positives. The main problem is that it doesn't consider which code is actually used by a Go program - any vulnerability in a dependency is assumed to apply.

So far, I have not found it useful for the sidecars.

PR #739, opened 2022-05-24
[release-3.1] backport 'fix managed-by label being too long when the node name is long.'

### LINKED ISSUE #707: distributed provisioner: managed-by label can fail if node name too long

akalenyu:
Signed-off-by: Alexander Wels <awels@redhat.com>

**What type of PR is this?**
/kind bug

Cherry-pick of #717 for the release-3.1 branch, as discussed in https://github.com/kubernetes-csi/external-provisioner/pull/717#issuecomment-1126920943.

Fixes #707

```release-note
storage capacity: managed-by label potentially too long with a long node name and enableNodeDeployment set to true
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -37,6 +37,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	validation "k8s.io/apimachinery/pkg/util/validation"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


pohly (approved):
/lgtm
/ok-to-test

pohly:
I have approver rights on master, but not on this release branch.

/assign @xing-yang 

For approval.

akalenyu:
Big thanks!
@pohly Would it be ok to ask for a 3.1.z release that contains this?

pohly:
I am going to prepare new releases now.

PR #753, opened 2022-06-27
fix: changed the csistoragecapacity check namespace

### LINKED ISSUE #752: CSIStorageCapacity check fails if the pod does not have permissions in "default"
The CSIStorageCapacity check attempts to create an invalid resource in the "default" namespace. If the pod does not have permissions for "default", the call fails with a fatal error.

https://github.com/kubernetes-csi/external-provisioner/pull/710#issuecomment-1166905128

nbalacha:
The CSIStorageCapacity v1 check was attempting to create an invalid
object in the default namespace. This would fail if the pod did not
have permissions to the namespace. This will now use the namespace
specified in the NAMESPACE env variable.

Signed-off-by: N Balachandran <nibalach@redhat.com>

**What type of PR is this?**
/kind bug

The CSIStorageCapacity version check attempts to create an invalid resource in the default namespace. If the pod does not have permissions on "default" , this check returns with a fatal error.

Fixes #752 

```release-note
The CSIStorageCapacity version check will now use the namespace specified in the NAMESPACE env variable. This avoids a fatal unexpected error when checking for the V1 CSIStorageCapacity API error during startup.
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -493,7 +493,7 @@ func main() {
 				Name: "#%123-invalid-name",
 			},
 		}
-		createdCapacity, err := clientset.StorageV1().CSIStorageCapacities("default").Create(ctx, invalidCapacity, metav1.CreateOptions{})
+		createdCapacity, err := clientset.StorageV1().CSIStorageCapacities(namespace).Create(ctx, invalidCapacity, metav1.CreateOptions{})
 		switch {
 		case err == nil:
 			klog.Fatalf("creating an invalid v1.CSIStorageCapacity didn't fail as expected, got: %s", createdCapacity)


pohly:
/lgtm

Before I approve and start preparing releases, was this tested on a cluster where the failure occurred?

pohly:
Also, can the change note be extended with "This avoids a fatal `unexpected error when checking for the V1 CSIStorageCapacity API` error during startup.?

nbalacha:
> /lgtm
> 
> Before I approve and start preparing releases, was this tested on a cluster where the failure occurred?

Thanks. Yes, I tried it out on the same cluster where the original image failed.

nbalacha:
> Also, can the change note be extended with "This avoids a fatal `unexpected error when checking for the V1 CSIStorageCapacity API` error during startup.?

I've added this in the PR (sorry, didn't know what you meant by change note). Let me know if I need to change it elsewhere.

pohly:
> I've added this in the PR (sorry, didn't know what you meant by change note).

Yes, that one.

PR #783, opened 2022-09-06
correct the fmt errors on the controller files.

humblec:
Reference #https://github.com/kubernetes-csi/csi-release-tools/pull/203

> /kind cleanup

-->
```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1668,7 +1668,7 @@ func verifyAndGetSecretNameAndNamespaceTemplate(secret secretParamsMap, storageC
 }
 
 // getSecretReference returns a reference to the secret specified in the given nameTemplate
-//  and namespaceTemplate, or an error if the templates are not specified correctly.
+// and namespaceTemplate, or an error if the templates are not specified correctly.
 // no lookup of the referenced secret is performed, and the secret may or may not exist.
 //
 // supported tokens for name resolution:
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -1668,7 +1668,9 @@ func verifyAndGetSecretNameAndNamespaceTemplate(secret secretParamsMap, storageC
 }
 
 // getSecretReference returns a reference to the secret specified in the given nameTemplate
-//  and namespaceTemplate, or an error if the templates are not specified correctly.
+//
+//	and namespaceTemplate, or an error if the templates are not specified correctly.
+//
```
jsafrane, line 1673: This looks very odd, I'd prefer if you just deleted the extra spaces + empty lines.

humblec: hmm , the CI failure we see here https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-release-tools/203/pull-kubernetes-csi-release-tools-external-provisioner/1567039207589810176#1:build-log.txt%3A169 made me to run `gofmt on this repo` which brought this change. I am wondering going ahead with the default styling would help us to avoid manual corrections in future.. wdyt? or Am I missing someting here @jsafrane

humblec: oh..ok.. looks like that extra space caused the `gofmt` to split the line with intendation..etc.. I have corrected as you suggested @jsafrane . ptal.. thanks.

PR #793, opened 2022-09-30
Fix string pointer comparison for source volume mode conversion

RaunakShah:
**What type of PR is this?**

/kind bug

This PR fixes a bug where string addresses were being compared for equality instead of the actual string values. 
Due to this bug, even if the source volume mode is unchanged, the VolumeSnapshotContent object needed the `allowVolumeModeChange` annotation to be set to true for PVC creation to succeed. 

Fixes #

```release-note
Fix string pointer comparison for source volume mode conversion
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1149,7 +1149,7 @@ func (p *csiProvisioner) getSnapshotSource(ctx context.Context, claim *v1.Persis
 	}
 
 	if p.preventVolumeModeConversion {
-		if snapContentObj.Spec.SourceVolumeMode != nil && claim.Spec.VolumeMode != nil && snapContentObj.Spec.SourceVolumeMode != claim.Spec.VolumeMode {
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -1149,7 +1149,7 @@ func (p *csiProvisioner) getSnapshotSource(ctx context.Context, claim *v1.Persis
 	}
 
 	if p.preventVolumeModeConversion {
-		if snapContentObj.Spec.SourceVolumeMode != nil && claim.Spec.VolumeMode != nil && snapContentObj.Spec.SourceVolumeMode != claim.Spec.VolumeMode {
+		if snapContentObj.Spec.SourceVolumeMode != nil && claim.Spec.VolumeMode != nil && *snapContentObj.Spec.SourceVolumeMode != *claim.Spec.VolumeMode {
```
pohly, line 1152: What if one pointer is nil and the other isn't? What should happen in that case?

PR #796, opened 2022-10-14
fix: add defence logic for PV without claime ref

zezaeoh:
**What type of PR is this?**

> /kind bug

App crashes when deleting PVs without ClaimRef.
```
I1014 03:34:47.331898       1 controller.go:1471] delete "pvc-a6f2adab-036b-4833-b5c7-eaf84626c351": started
E1014 03:34:47.332068       1 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 257 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x1805020?, 0x29703c0})
	/workspace/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:75 +0x99
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc0005af920?})
	/workspace/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:49 +0x75
panic({0x1805020, 0x29703c0})
	/go/pkg/csiprow.XXXXDeFLHD/go-1.18/src/runtime/panic.go:838 +0x207
github.com/kubernetes-csi/external-provisioner/pkg/controller.(*csiProvisioner).getSecretsFromSC(0xc0003b8640, {0x1cdf738, 0xc0005b4300}, 0xc00071e500, 0x0, 0xc0006f08c0)
	/workspace/pkg/controller/controller.go:1272 +0x2a2
...
```

Fixes #None

```release-note
None
```


RaunakShah:
Can you add some logs to the description to show that this error is triggered?

pkg/controller/controller.go
```diff
@@ -1283,6 +1283,9 @@ func (p *csiProvisioner) getSecretsFromSC(ctx context.Context, volume *v1.Persis
 			}
 
 			// Resolve provision secret credentials.
+			if volume.Spec.ClaimRef == nil {
+				return fmt.Errorf("failed to get secretreference for volume %s: claim reference not exists", volume.Name)
```
RaunakShah, line 1287: nit: `not exists` -> `does not exist`

zezaeoh: thx! 😄

zezaeoh:
@RaunakShah 

Could you review again? :)

pkg/controller/controller.go
```diff
@@ -1283,6 +1283,9 @@ func (p *csiProvisioner) getSecretsFromSC(ctx context.Context, volume *v1.Persis
 			}
 
 			// Resolve provision secret credentials.
+			if volume.Spec.ClaimRef == nil {
```
Madhu-1, line 1286: If we dont have the secrets, dont we need to allow deletion? lets the CSI driver try to handle it

Madhu-1 (commented):
Commits need to be squashed into one.

pkg/controller/controller.go
zezaeoh: Actually, `action-runner-controller` deletes PV's  ClaimRef intentionally to make PV orphan and reuse.
So, i think we should leave this PV as orphan

Madhu-1: Yes, that is for the `ReclaimPolicy=Retained` PV right, not for the `ReclaimPolicy=Delete`?

zezaeoh: Oh.. you are right! @Madhu-1 
I'll follow your suggestion :) thx!!

zezaeoh:
@Madhu-1 @RaunakShah 

Could you review again?

pkg/controller/controller.go
Madhu-1: Change LGTM, you need review or approval from @msau42 @jsafrane

RaunakShah:
@zezaeoh changes look good, can you add also add some testing results to the description? For example, the logs showing that `csi-provisioner` doesn't crash but also that the CSI driver eventually does not delete the volume?

jsafrane:
@zezaeoh do you know how it's possible that a volume for PV without `ClaimRef` is deleted? That should be nearly impossible.

PR #800, opened 2022-11-01
Pass PVC annotations to req.Parameters when --extra-create-metadata is enabled

### LINKED ISSUE #86: Adding PVC's annotation to CreateVolumeRequest.Parameters
As the title says, there is a great possibility that the application-specific `Operator` would leverage `PVC`'s `Annotation` for storing custom data. The custom data will tell storage provisioner how  CSI drive create volume correctly.

Here is already existing logic for provision volume  : 

```
	// Create a CSI CreateVolumeRequest and Response
	req := csi.CreateVolumeRequest{

		Name:       pvName,
		Parameters: options.Parameters,
		VolumeCapabilities: []*csi.VolumeCapability{
			{
				AccessType: accessType,
				AccessMode: accessMode,
			},
		},
		CapacityRange: &csi.CapacityRange{
		},
	}
```

Note : more [details](https://github.com/kubernetes-csi/external-provisioner/blob/bc0b6718292ab0cd76697d997059f060e3228afb/pkg/controller/controller.go#L298)

CSI Plugin specifies that `CreateVolumeRequest`'s field `Parameters` is `OPTIONAL` and opaque.
What I am suggesting is that we could pass in `PVC`'s `Annotation` in addition to `Storageclass.Parameter`.

vinli-cn:
**What type of PR is this?**
/kind feature

Adding PVC's annotation to CreateVolumeRequest.Parameters when --extra-create-metadata is enabled, fix issue #86

Fixes #86 

```release-note
Adding PVC's annotation to CreateVolumeRequest.Parameters when --extra-create-metadata is enabled
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -113,9 +114,10 @@ const (
 	nodePublishSecretNamespaceKey = "csiNodePublishSecretNamespace"
 
 	// PV and PVC metadata, used for sending to drivers in the  create requests, added as parameters, optional.
-	pvcNameKey      = "csi.storage.k8s.io/pvc/name"
-	pvcNamespaceKey = "csi.storage.k8s.io/pvc/namespace"
-	pvNameKey       = "csi.storage.k8s.io/pv/name"
+	pvcNameKey                 = "csi.storage.k8s.io/pvc/name"
+	pvcNamespaceKey            = "csi.storage.k8s.io/pvc/namespace"
+	pvcNamespaceAnnotationsKey = "csi.storage.k8s.io/pvc/namespace/annotations"
+	pvNameKey                  = "csi.storage.k8s.io/pv/name"
 
 	snapshotKind     = "VolumeSnapshot"
 	snapshotAPIGroup = snapapi.GroupName       // "snapshot.storage.k8s.io"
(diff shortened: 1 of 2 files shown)


ggriffiths (changes requested):
@vinli-cn Your PR title says "pass namespace annotations", but your code is passing PVC annotations to the driver. 

Passing PVC annotations to a CSI Driver request has been discussed at length in this thread:
https://github.com/kubernetes-csi/external-provisioner/issues/86

I would read up on the thread and see if you have anything else to add as to why PVC annotations should be added to the CSI create volume request.

pkg/controller/controller.go
```diff
@@ -716,9 +718,19 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 	}
 
 	if p.extraCreateMetadata {
+		pvcAnnotationMap := claim.GetAnnotations()
```
humblec, line 721: this is geting annotations from PVC and not namespace as mentioned in the PR.

vinli-cn:
> @vinli-cn Your PR title says "pass namespace annotations", but your code is passing PVC annotations to the driver.
> 
> Passing PVC annotations to a CSI Driver request has been discussed at length in this thread: #86
> 
> I would read up on the thread and see if you have anything else to add as to why PVC annotations should be added to the CSI create volume request.

thanks for reviewing the PR.

I updated the title and description of this PR according to issue #86, user might need to make use of the extra PVC's annotations.

Could you please review again and share your insights? 


Thanks!

ggriffiths:
Hi @vinli-cn, thank you for updating the PR description!

I think your PR title is still incorrect:
> Pass namespace annotations to driver when --extra-create-metadata is enabled

Can you edit this to the below so we're clear what's being proposed?
> Pass PVC annotations

Anyway, @vinli-cn, you should read up on the thread in issue #86. Many people have proposed this change in the past, but the maintainers decided to not add it.

If you think you have a compelling use case as to why this should be supported, send a message in #86.  Discussion in this current pull request will not result in it being merged.

ggriffiths (changes requested):
Feel free to re-request review if this proposal is accepted in the issue I linked.

PR #802, opened 2022-11-03
update sidecars to latest version

humblec:
> /kind cleanup

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/storage-capacity.yaml
+++ b/deploy/kubernetes/storage-capacity.yaml
@@ -22,7 +22,7 @@ spec:
       serviceAccount: csi-provisioner
       containers:
         - name: csi-provisioner
-          image: registry.k8s.io/sig-storage/csi-provisioner:v3.1.0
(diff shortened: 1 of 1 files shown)


deploy/kubernetes/storage-capacity.yaml
```diff
@@ -22,7 +22,7 @@ spec:
       serviceAccount: csi-provisioner
       containers:
         - name: csi-provisioner
-          image: registry.k8s.io/sig-storage/csi-provisioner:v3.1.0
+          image: registry.k8s.io/sig-storage/csi-provisioner:v3.3.0
```
msau42, line 25: Can we make these templates so we don't need to remember to update the images every release?

humblec: will open an issue for the improvement/automation @msau42 .

humblec: @msau42 can we get this in regardless?

Issue #804, opened 2022-11-03
Add provisioning volumes from cross-namespace data sources


ttakahashi21:
**Is your feature request related to a problem?/Why is this needed**

Please see [KEP-3294](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/3294-provision-volumes-from-cross-namespace-snapshots)

**Describe the solution you'd like in detail**

method for provisioning volumes from cross-namespace snapshot

**Describe alternatives you've considered**

Please see [KEP-3294 Alternatives](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/3294-provision-volumes-from-cross-namespace-snapshots#alternatives)

**Additional context**

Discuss the API definition [here](https://github.com/kubernetes/kubernetes/pull/113186). 
The code may change due to API specification changes.

msau42:
/close
Fixed by #805

PR #805, opened 2022-11-03
Alpha support for provision volumes from cross-namespace data sources

ttakahashi21:
**What type of PR is this?**
/kind feature

Please see [CrossNamespaceVolumeDataSource](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/3294-provision-volumes-from-cross-namespace-snapshots)

Part of #https://github.com/kubernetes-csi/external-provisioner/issues/804

Discuss the API definition https://github.com/kubernetes/kubernetes/pull/113186.
The code may change due to API specification changes.

```release-note
- [KEP]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/3294-provision-volumes-from-cross-namespace-snapshots
```


pkg/controller/controller.go
```diff
@@ -57,6 +59,7 @@ import (
 	"github.com/kubernetes-csi/csi-lib-utils/rpc"
 	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
 	snapclientset "github.com/kubernetes-csi/external-snapshotter/client/v6/clientset/versioned"
+	referenceGrantv1alpha2 "sigs.k8s.io/gateway-api/pkg/client/listers/apis/v1alpha2"
```
humblec, line 62: just wondering are we fine to use alphav2 of RefGrants? considering we have v1beta1 avaiable , cant we or better to use it ?

pkg/features/features.go
```diff
@@ -43,6 +50,7 @@ func init() {
 // defaultKubernetesFeatureGates consists of all known feature keys specific to external-provisioner.
 // To add a new feature, define a key for it above and add it here.
 var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	Topology:             {Default: false, PreRelease: featuregate.GA},
-	HonorPVReclaimPolicy: {Default: false, PreRelease: featuregate.Alpha},
+	Topology:                         {Default: false, PreRelease: featuregate.GA},
+	HonorPVReclaimPolicy:             {Default: false, PreRelease: featuregate.Alpha},
+	CrossNamespaceSourceProvisioning: {Default: false, PreRelease: featuregate.Alpha},
```
humblec, line 55: we also need to add this feature gate to the readme doc

humblec:
@ttakahashi21 thanks.  I have few review comments . Apart from that, can we add some tests to this PR ?

pkg/controller/controller.go
xing-yang: +1

pkg/controller/controller.go
```diff
@@ -1081,47 +1126,114 @@ func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentV
...
+}
+
+// getSnapshotSourceDataSourceRef2 verifies DataSourceRef2.Kind of type VolumeSnapshot, making sure that the requested Snapshot is available/ready
+// returns the VolumeContentSource for the requested snapshot
+func (p *csiProvisioner) getSnapshotSourceDataSourceRef2(ctx context.Context, claim *v1.PersistentVolumeClaim, sc *storagev1.StorageClass) (*csi.VolumeContentSource, error) {
+	if claim.Spec.DataSourceRef2.Namespace == "" {
+		return p.getSnapshotSourceInternal(ctx, claim, sc, claim.Namespace, claim.Spec.DataSourceRef2.Name)
+	}
+
+	if ok, err := p.IsGranted(ctx, claim.Spec.DataSourceRef2.Namespace, claim.Spec.DataSourceRef2.Name, claim); err != nil || !ok {
```
xing-yang, line 1190: Do we check somewhere if the PVC namespace is the same as the VolumeSnapshot namespace?  If they are the same, then referenceGrant is not needed.

pkg/controller/controller.go
ttakahashi21: The ReferenceGrant is not yet in beta. Therefore, I'm using Alpha2.
See the URL below.
https://github.com/kubernetes-sigs/gateway-api

pkg/features/features.go
ttakahashi21: Could you tell me where should i write it?

pkg/controller/controller.go
humblec: @ttakahashi21 isnt this https://github.com/kubernetes-sigs/gateway-api/tree/main/pkg/client/listers/apis/v1beta1  available ?

ttakahashi21: Thank you for letting me know.

ttakahashi21: I was able to use it by adding the following to my go.mod file:
> require sigs.k8s.io/gateway-api v0.5.1-0.20221110191515-2d586cc49ab4

pkg/features/features.go
ttakahashi21: Added CrossNamespaceVolumeDataSource in [README.md](https://github.com/kubernetes-csi/external-provisioner/pull/805/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).

pkg/controller/controller.go
```diff
@@ -582,6 +589,21 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
...
+	if utilfeature.DefaultFeatureGate.Enabled(features.CrossNamespaceVolumeDataSource) {
+		if claim.Spec.DataSourceRef != nil && claim.Spec.DataSourceRef.Namespace != nil && len(*claim.Spec.DataSourceRef.Namespace) > 0 {
+			switch claim.Spec.DataSourceRef.Kind {
+			case snapshotKind:
+				if *(claim.Spec.DataSourceRef.APIGroup) != snapshotAPIGroup {
+					return nil, controller.ProvisioningFinished, fmt.Errorf("the PVC source does not belong to the right APIGroup. Expected %s, Got %s", snapshotAPIGroup, *(claim.Spec.DataSourceRef.APIGroup))
+				}
+				rc.snapshot = true
+			default:
+				return nil, controller.ProvisioningFinished, fmt.Errorf("the PVC source specified via DataSourceRef with non-empty namespace does not belong to the right Kind.  Expected %s, Got %s", snapshotKind, claim.Spec.DataSourceRef.Kind)
```
msau42, line 602: Are we allowing pvc source and data populators too?

pkg/controller/controller.go
```diff
@@ -1081,47 +1131,115 @@ func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentV
...
+				break
+			}
+		}
+		// Skip unrelated policy by checking From field
+		if !validFrom {
+			continue
+		}
+
+		for _, to := range grant.Spec.To {
+			if to.Group != snapshotAPIGroup || to.Kind != snapshotKind {
```
msau42, line 1161: Can we make the function more generic to handle non-snapshot data sources?

pkg/controller/controller.go
```diff
@@ -582,6 +589,21 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 			}
 		}
 	}
+
+	if utilfeature.DefaultFeatureGate.Enabled(features.CrossNamespaceVolumeDataSource) {
+		if claim.Spec.DataSourceRef != nil && claim.Spec.DataSourceRef.Namespace != nil && len(*claim.Spec.DataSourceRef.Namespace) > 0 {
```
msau42, line 594: I think we should move the check for feature gate and return error earlier at this place. There's a number of csi calls we make after this that may be nice to avoid.

pkg/controller/controller.go
```diff
@@ -631,7 +653,17 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		},
 	}
 
-	if claim.Spec.DataSource != nil && (rc.clone || rc.snapshot) {
+	if claim.Spec.DataSourceRef != nil && claim.Spec.DataSourceRef.Namespace != nil && len(*claim.Spec.DataSourceRef.Namespace) > 0 {
+		if utilfeature.DefaultFeatureGate.Enabled(features.CrossNamespaceVolumeDataSource) && rc.snapshot {
+			volumeContentSource, err := p.getVolumeContentSourceFromXnsDataSource(ctx, claim, sc)
```
msau42, line 658: Do we need a new function, or can we modify `getVolumeContentSource`?

ttakahashi21: Yes. Is possible. Fixed code https://github.com/kubernetes-csi/external-provisioner/pull/805/commits/1f9df93e646bbd126bf7bcc172a7e2a3a2da4e8e
Could you check code?

pkg/controller/controller.go
ttakahashi21: I added feature gate check code early in the prepare Provision function.

pkg/controller/controller.go
ttakahashi21: Yes. Added. https://github.com/kubernetes-csi/external-provisioner/commit/1f9df93e646bbd126bf7bcc172a7e2a3a2da4e8e

deploy/kubernetes/rbac.yaml
```diff
@@ -57,6 +57,9 @@ rules:
   - apiGroups: ["storage.k8s.io"]
     resources: ["volumeattachments"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["gateway.networking.k8s.io"]
```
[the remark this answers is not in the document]
ttakahashi21: I see. But if I comment it out, CI fails. Is it possible to test without commenting out in the CI side test?

go.mod
```diff
@@ -74,13 +76,13 @@ require (
...
 	k8s.io/kube-openapi v0.0.0-20220803164354-a70c9af30aea // indirect
-	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed // indirect
+	k8s.io/utils v0.0.0-20221107191617-1a15be271d1d // indirect
 	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace k8s.io/api => k8s.io/api v0.25.2
+replace k8s.io/api => k8s.io/api v0.0.0-20221110160654-5cb32024090c
```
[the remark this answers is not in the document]
ttakahashi21: I set it temporarily because I need to use the API added for k8s 1.26 for successful CI testing.  Therefore, I will change it when the release tag is attached.

pkg/controller/controller.go
```diff
@@ -582,6 +594,28 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 			}
 		}
 	}
+	if utilfeature.DefaultFeatureGate.Enabled(features.CrossNamespaceVolumeDataSource) {
+		if claim.Spec.DataSourceRef != nil {
```
msau42, line 598: This logic looks the same as the DataSource logic. Can we create a common function that takes in Name and Kind?

pkg/controller/controller.go
```diff
@@ -980,29 +1032,71 @@ func removePrefixedParameters(param map[string]string) (map[string]string, error
...
+		switch claim.Spec.DataSourceRef.Kind {
+		case snapshotKind:
+			return p.getSnapshotSource(ctx, claim, sc)
+		case pvcKind:
+			return p.getPVCSource(ctx, claim, sc)
+		default:
+			// For now we shouldn't pass other things to this function, but treat it as a noop and extend as needed
+			return nil, nil
+		}
+	} else {
```
msau42, line 1045: Do we need the if/else? The logic looks exactly the same.

pkg/controller/controller.go
```diff
@@ -1081,47 +1175,117 @@ func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentV
 	return volumeContentSource, nil
 }
 
+func (p *csiProvisioner) IsGranted(ctx context.Context, claim *v1.PersistentVolumeClaim) (bool, error) {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.CrossNamespaceVolumeDataSource) {
+		return false, nil
+	}
+
+	// Get all ReferenceGrants in snapshot's namespace
```
msau42, line 1183: update comment to say "data source" instead of "snapshot"

pkg/controller/controller.go
```diff
@@ -1081,47 +1175,117 @@ func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentV
 	return volumeContentSource, nil
 }
 
+func (p *csiProvisioner) IsGranted(ctx context.Context, claim *v1.PersistentVolumeClaim) (bool, error) {
```
msau42, line 1178: Add a todo that this function should eventually be replaced by a common kubernetes library

pkg/controller/controller.go
```diff
@@ -1081,47 +1175,117 @@ func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentV
...
+				break
+			}
+		}
+		// Skip unrelated policy by checking From field
+		if !validFrom {
+			continue
+		}
+
+		for _, to := range grant.Spec.To {
+			if claim.Spec.DataSourceRef.APIGroup != nil && (string(to.Group) != *claim.Spec.DataSourceRef.APIGroup || string(to.Kind) != claim.Spec.DataSourceRef.Kind) {
```
msau42, line 1205: What if `dataSourceRef.APIGroup == nil` and `to.Group is set`?

Also it should be `groupMatches` and `kindMatches` right?

pkg/controller/controller.go
```diff
@@ -1081,47 +1175,117 @@ func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentV
...
+			}
+		}
+
+		if allowed {
+			break
+		}
+	}
+
+	if !allowed {
+		return false, nil
```
msau42, line 1220: I think this could just be `return allowed, nil`

pkg/controller/controller.go
```diff
@@ -924,7 +964,19 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 }
 
 func (p *csiProvisioner) setCloneFinalizer(ctx context.Context, pvc *v1.PersistentVolumeClaim) error {
-	claim, err := p.claimLister.PersistentVolumeClaims(pvc.Namespace).Get(pvc.Spec.DataSource.Name)
+	var claim *v1.PersistentVolumeClaim
+	var err error
+
+	if pvc.Spec.DataSourceRef != nil {
```
msau42, line 970: How much of this logic can be simplified across all the data sources if we had a common function that can normalize DataSource and DataSourceRef? Something like:
```
func dataSource(claim *v1.PersistentVolumeClaim) (ObjectReference, error)
```

I see that this pattern repeats many times for each of the data source types.

go.mod
```diff
@@ -29,6 +29,8 @@ require (
 	sigs.k8s.io/sig-storage-lib-external-provisioner/v8 v8.0.0
 )
 
+require sigs.k8s.io/gateway-api v0.5.1-0.20221110191515-2d586cc49ab4
```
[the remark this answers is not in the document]
ttakahashi21: > Also is there a release tag of gateway-api we can use?
There is none.

@robscott
Could you please create a tag that includes the ReferenceGrant API Beta?

pkg/controller/controller.go
```diff
@@ -537,6 +544,11 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		return nil, controller.ProvisioningFinished, errors.New("storage class was nil")
 	}
 
+	if !utilfeature.DefaultFeatureGate.Enabled(features.CrossNamespaceVolumeDataSource) &&
+		claim.Spec.DataSourceRef != nil && claim.Spec.DataSourceRef.Namespace != nil && len(*claim.Spec.DataSourceRef.Namespace) > 0 {
+		return nil, controller.ProvisioningFinished, fmt.Errorf("error handling for DataSourceRef Type %s with non-empty namespace by Name %s: CrossNamespaceVolumeDataSource feature disabled", claim.Spec.DataSourceRef.Kind, claim.Spec.DataSourceRef.Name)
```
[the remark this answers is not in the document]
ttakahashi21: Fixed.

pkg/controller/controller.go
ttakahashi21: Fixed. https://github.com/kubernetes-csi/external-provisioner/pull/805/files#diff-243ebed2765f75e6a54f57167212fefb08c3b2a85967ad2acbc0eb78919019c1R1121

pkg/controller/controller.go
ttakahashi21: I fixed it as follows. Is this the fix you expected?
https://github.com/kubernetes-csi/external-provisioner/pull/805/files#diff-243ebed2765f75e6a54f57167212fefb08c3b2a85967ad2acbc0eb78919019c1R1145-R1149

pkg/controller/controller.go
ttakahashi21: Fixed.

pkg/controller/controller.go
ttakahashi21: Fixed.

pkg/controller/controller.go
ttakahashi21: Changed to check the featuregate in the dataSource function. Also, since it is confirmed at the beginning of the prepareProvision function, the confirmation of featuregate removed in the "IsGrant ", "getSnapshotSource ", and "getPVCSource " function.

pkg/controller/controller.go
ttakahashi21: Fixed.

ttakahashi21:
https://github.com/kubernetes-csi/external-provisioner/pull/805#discussion_r1028712172
I'm sorry. I was mistaken. No error occurs.

pkg/controller/controller.go
ttakahashi21: Fixed.

pkg/controller/controller.go
```diff
@@ -923,8 +940,16 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 	return pv, controller.ProvisioningFinished, nil
 }
 
-func (p *csiProvisioner) setCloneFinalizer(ctx context.Context, pvc *v1.PersistentVolumeClaim) error {
-	claim, err := p.claimLister.PersistentVolumeClaims(pvc.Namespace).Get(pvc.Spec.DataSource.Name)
+func (p *csiProvisioner) setCloneFinalizer(ctx context.Context, pvc *v1.PersistentVolumeClaim, dataSource *v1.ObjectReference) error {
+	var claim *v1.PersistentVolumeClaim
+	var err error
+
+	if len(dataSource.Namespace) > 0 {
```
msau42, line 947: Can `dataSource` also set the namespace field so we don't need this check?

pkg/controller/controller.go
```diff
@@ -993,16 +1018,28 @@ func (p *csiProvisioner) getVolumeContentSource(ctx context.Context, claim *v1.P
...
 // returns the VolumeContentSource for the requested PVC
-func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentVolumeClaim, sc *storagev1.StorageClass) (*csi.VolumeContentSource, error) {
-	sourcePVC, err := p.claimLister.PersistentVolumeClaims(claim.Namespace).Get(claim.Spec.DataSource.Name)
+func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentVolumeClaim, sc *storagev1.StorageClass, dataSource *v1.ObjectReference) (*csi.VolumeContentSource, error) {
+
+	var sourcePVC *v1.PersistentVolumeClaim
+	var err error
+	if len(dataSource.Namespace) > 0 {
+		if claim.Namespace != dataSource.Namespace {
+			if ok, err := p.IsGranted(ctx, claim); err != nil || !ok {
```
msau42, line 1027: Can `IsGranted` also be moved to `dataSource`?

pkg/controller/controller.go
msau42: btw does the gateway controller have a similar method?

pkg/controller/controller_test.go
```diff
@@ -4470,6 +4840,92 @@ func generatePVCForProvisionFromPVC(srcNamespace, srcName, scName string, reques
 	return provisionRequest
 }
 
+// generatePVCForProvisionFromPVC returns a ProvisionOptions with the requested settings
+func generatePVCForProvisionFromXnsPVC(srcNamespace, srcName, scName, kind string, refpvcNamespace, apiGroup *string, requestedBytes int64, volumeMode string, setnamespace bool) controller.ProvisionOptions {
+	deletePolicy := v1.PersistentVolumeReclaimDelete
+
+	var provisionRequest controller.ProvisionOptions
+	if setnamespace {
+		provisionRequest = controller.ProvisionOptions{
```
msau42, line 4849: It looks like almost everything is the same and the only difference is the namespace field. I think you can move this out of this `if` block and only change the namespace inside.

pkg/controller/controller_test.go
```diff
@@ -4671,12 +5134,170 @@ func TestProvisionFromPVC(t *testing.T) {
...
+			refGrant:    newRefGrant("refGrant1", srcNamespace, ns1, coreapiGrp, "PersistentVolumeClaim"),
+		},
 	}
 
 	for k, tc := range testcases {
 		tc := tc
 		t.Run(k, func(t *testing.T) {
-			t.Parallel()
+			if !tc.xnsEnabled {
+				t.Parallel()
```
msau42, line 5299: Why does the xns feature require serial tests?

pkg/controller/controller_test.go
```diff
@@ -4803,11 +5448,28 @@ func TestProvisionFromPVC(t *testing.T) {
...
+							},
+						}
+					} else if tc.volOpts.PVC.Spec.DataSource != nil {
+						volumeSource = csi.VolumeContentSource_Volume{
+							Volume: &csi.VolumeContentSource_VolumeSource{
+								VolumeId: tc.volOpts.PVC.Spec.DataSource.Name,
+							},
+						}
+					}
+				} else {
```
msau42, line 5467: Can this be collapsed into 2 conditions?

pkg/controller/controller.go
```diff
@@ -1081,47 +1118,108 @@ func (p *csiProvisioner) getPVCSource(ctx context.Context, claim *v1.PersistentV
 	return volumeContentSource, nil
 }
 
+// TODO: This function should eventually be replaced by a common kubernetes library.
```
msau42, line 1121: For now this is a pretty important function from a security perspective. Can we move this out into a util file and have extensive unit tests for it?

pkg/controller/controller.go
msau42: It looks much better now thanks!

pkg/controller/controller_test.go
ttakahashi21: Because the test did not work with feature gate enabled when it was executed in parallel test.

pkg/controller/controller_test.go
ttakahashi21: Because the test did not work with feature gate enabled when it was executed in parallel test.

pkg/controller/controller.go
ttakahashi21: It is possible.

pkg/controller/controller.go
ttakahashi21: Is it possible. Fixed code.

pkg/controller/controller.go
robscott: I'm not sure I get the context here. Agree that ReferenceGrant is only useful/necessary to enable cross-namespace references. Same namespace references should not require a ReferenceGrant.

pkg/controller/controller_test.go
msau42: Hm that seems to imply there is some global shared state somewhere. I think it would be good to investigate more into that.

pkg/controller/controller_test.go
ttakahashi21: Fixed.

pkg/controller/controller_test.go
ttakahashi21: Fixed.

ttakahashi21:
@msau42 
https://github.com/kubernetes-csi/external-provisioner/pull/805#discussion_r1036460110
I found two similar logics.

[cillum]
https://github.com/cilium/cilium/pull/21902/files#diff-c97054ccf2e8ead85e142ecda08c521b76ce13f924cc79dac145d380cd5bcf13R152-R174

[envoyproxy]
https://github.com/envoyproxy/gateway/pull/172/files#diff-be48a4416dad5f211bed7f15e1cc13e8cdfaa1378a7d5be349d8bb06e4b2054eR706-R745

pkg/controller/controller_test.go
ttakahashi21: `DefaultFeatureGate` defined in [utilfeature](https://github.com/kubernetes/apiserver/blob/master/pkg/util/feature/feature_gate.go#L32) seems a global value, therefore setting [value](https://github.com/kubernetes-csi/external-provisioner/blob/3bfb91586acdcce7c8cd1796ebe0ab2770cf880d/vendor/k8s.io/component-base/featuregate/testing/feature_gate.go#L35) during tests causes changes to a global value in each test.

pkg/controller/util_test.go
```diff
@@ -0,0 +1,175 @@
...
+	fromNamespace := "ns1"
+	fromsrcName := "test-dataSource"
+	toNamespace := "ns2"
+
+	type testcase struct {
+		expectErr    bool
+		volOpts      controller.ProvisionOptions
+		refGrantList []*gatewayv1beta1.ReferenceGrant
+	}
+	testcases := map[string]testcase{
```
msau42, line 64: Can you add a test case for nil datasource apigroup?

pkg/controller/util_test.go
```diff
@@ -0,0 +1,175 @@
...
+		"Allowed to access dataSource for xns PVC with refGrant": {
+			volOpts:      generatePVCForProvisionFromXnsdataSource(toNamespace, fromsrcName, "", pvcapiKind, &fromNamespace, &coreapiGrp, requestedBytes, "", true),
+			refGrantList: newRefGrantList("refGrant1", fromNamespace, coreapiGrp, pvcapiKind, toNamespace, coreapiGrp, pvcapiKind, "", false),
```
msau42, line 67: I know we do this in our other test files, but I think it's a little difficult to read the test case with this style of constructor because I need to remember the ordering of the arguments and which field they correspond to. It's perhaps more lines, but I think it's more readable to directly reference the object in the test case definition.

If you want to reduce the boilerplate, maybe you can still have a constructor, but have it take a `ReferenceGrantFrom` and a `ReferenceGrantTo` as arguments?

Similar comment for `generatePVC...` Keep the constructor to reduce on the common boilerplate, but have it take `DataSourceRef` type as an input, which is what changes across each test case.

msau42 (commented):
Thanks for all the great work! I just have a few unit test suggestions.

pkg/controller/controller_test.go
ttakahashi21: I added the link of the issue I created as a comment

pkg/controller/controller.go
```diff
@@ -923,8 +940,12 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 	return pv, controller.ProvisioningFinished, nil
 }
 
-func (p *csiProvisioner) setCloneFinalizer(ctx context.Context, pvc *v1.PersistentVolumeClaim) error {
-	claim, err := p.claimLister.PersistentVolumeClaims(pvc.Namespace).Get(pvc.Spec.DataSource.Name)
+func (p *csiProvisioner) setCloneFinalizer(ctx context.Context, pvc *v1.PersistentVolumeClaim, dataSource *v1.ObjectReference) error {
+	var claim *v1.PersistentVolumeClaim
```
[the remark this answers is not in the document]
ttakahashi21: Fixed.

pkg/controller/util_test.go
ttakahashi21: https://github.com/kubernetes-csi/external-provisioner/pull/805/files#diff-640957054096359bf678ef9f2789673a60814fba8286165ad298f5eba881536aR215-R222
I added.

pkg/controller/util_test.go
ttakahashi21: util_test.go created a base test case. And I modified the base test case and modified it to add the test case.
Along with the above modifications, TestProvisionFromSnapshot and TestProvisionFromPVC have also been modified.

In util.go I added the following just in case:
https://github.com/kubernetes-csi/external-provisioner/pull/805/files#diff-ded124c02add3467ce103641d0cd28436ea69cf37c5906470b3d519009d950b7R16-R20

As it turned out to be necessary in the test case below.
https://github.com/kubernetes-csi/external-provisioner/pull/805/files#diff-640957054096359bf678ef9f2789673a60814fba8286165ad298f5eba881536aR166-R170

This condition is unnecessary in the controller because it is acquired by specifying the namespace, but I added it just in case.
https://github.com/kubernetes-csi/external-provisioner/pull/805/files#diff-243ebed2765f75e6a54f57167212fefb08c3b2a85967ad2acbc0eb78919019c1R1936-R1937

msau42:
/lgtm
/approve

Thanks for adding all the comprehensive unit tests!

For the release note, can you update it to say "Add support for cross-namespace data sources alpha feature"

PR #808, opened 2022-11-08
Replace templates in SC parameter with PVC annotations

vinli-cn:
**What type of PR is this?**
/kind feature

Replace templates in SC parameter with PVC annotations, this fixes issue 428 in csi-driver-smb repo: https://github.com/kubernetes-csi/csi-driver-smb/issues/428

This change has been tested manually with SMB CSI Driver which uses SC parameter subDir in SMB mount path, code example as below

> Storage Class
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: smb
provisioner: smb.csi.k8s.io
parameters:
  subDir: ${pvc.annotations['subDirInPVC']}
```

> PVC
```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: pvc-smb
  annotations:
    subDirInPVC: "dir1"
```

Fixes #

```release-note
Replace SC parameter template with PVC annotations
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -121,10 +121,10 @@ const (
 	snapshotAPIGroup = snapapi.GroupName       // "snapshot.storage.k8s.io"
 	pvcKind          = "PersistentVolumeClaim" // Native types don't require an API group
 
-	tokenPVNameKey                 = "pv.name"
-	tokenPVCNameKey                = "pvc.name"
-	tokenPVCNameSpaceKey           = "pvc.namespace"
-	tokenPVAnnoationTemplatePrefix = "pvc.annotations"
+	tokenPVNameKey                  = "pv.name"
+	tokenPVCNameKey                 = "pvc.name"
+	tokenPVCNameSpaceKey            = "pvc.namespace"
+	tokenPVCAnnoationTemplatePrefix = "pvc.annotations"
 
 	ResyncPeriodOfCsiNodeInformer = 1 * time.Hour
 
@@ -598,14 +598,14 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 
 	pvcAnnotationsForSCParam := map[string]string{}
 	for k, v := range claim.GetAnnotations() {
-		pvcAnnotationsForSCParam[tokenPVAnnoationTemplatePrefix+"['"+k+"']"] = v
+		pvcAnnotationsForSCParam[tokenPVCAnnoationTemplatePrefix+"['"+k+"']"] = v
 	}
 
 	fsTypesFound := 0
 	fsType := ""
 	for k, v := range sc.Parameters {
 		// Replace SC parameters with PVC annoations
-		if strings.Contains(v, tokenPVAnnoationTemplatePrefix) {
+		if strings.Contains(v, tokenPVCAnnoationTemplatePrefix) {
 			resolvedName, err := resolveTemplate(v, pvcAnnotationsForSCParam)
 
 			if err != nil {
@@ -1765,7 +1765,7 @@ func getSecretReference(secretParams secretParamsMap, storageClassParams map[str
 			nameParams[tokenPVCNameKey] = pvc.Name
 			nameParams[tokenPVCNameSpaceKey] = pvc.Namespace
 			for k, v := range pvc.Annotations {
-				nameParams[tokenPVAnnoationTemplatePrefix+"['"+k+"']"] = v
+				nameParams[tokenPVCAnnoationTemplatePrefix+"['"+k+"']"] = v
 			}
 		}
 		resolvedName, err := resolveTemplate(nameTemplate, nameParams)
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -121,9 +121,10 @@ const (
 	snapshotAPIGroup = snapapi.GroupName       // "snapshot.storage.k8s.io"
 	pvcKind          = "PersistentVolumeClaim" // Native types don't require an API group
 
-	tokenPVNameKey       = "pv.name"
-	tokenPVCNameKey      = "pvc.name"
-	tokenPVCNameSpaceKey = "pvc.namespace"
+	tokenPVNameKey                 = "pv.name"
+	tokenPVCNameKey                = "pvc.name"
+	tokenPVCNameSpaceKey           = "pvc.namespace"
+	tokenPVAnnoationTemplatePrefix = "pvc.annotations"
 
 	ResyncPeriodOfCsiNodeInformer = 1 * time.Hour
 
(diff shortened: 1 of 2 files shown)


andyzhangx:
@vinli-cn could you add more details in PR description? thanks.

pkg/controller/controller.go
```diff
@@ -595,9 +595,24 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
...
+	for k, v := range claim.GetAnnotations() {
+		pvcAnnotationsForSCParam["pvc.annotations['"+k+"']"] = v
+	}
+
 	fsTypesFound := 0
 	fsType := ""
 	for k, v := range sc.Parameters {
+		// Replace SC parameters with PVC annoations
+		if strings.Contains(v, "pvc.annotations") {
+			resolvedName, err := resolveTemplate(v, pvcAnnotationsForSCParam)
```
andyzhangx, line 608: write like this:
```
if resolvedName, err := resolveTemplate(v, pvcAnnotationsForSCParam); err != nil {

}
```

pkg/controller/controller.go
```diff
@@ -595,9 +595,24 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
...
+
+			if err != nil {
+				return nil, controller.ProvisioningFinished, fmt.Errorf("error resolving value %q: %v", v, err)
```
andyzhangx, line 611: are you sure you want to return error here just because resolving value failure, I think it's better logging here.

pkg/controller/controller.go
vinli-cn: If we write it like this, variable resolvedName can't be reached out of "if {}" scope (line 614)

pkg/controller/controller.go
vinli-cn: If a parameter follows "${pvc.annoations" pattern but it can't find any match in pvc.annoations, it should be a human mistake, either bad annotation name in SC parameter or missing annotation in pvc.annoations.

Given that, I think it's better to error it out.

pkg/controller/controller.go
```diff
@@ -121,9 +121,10 @@ const (
 	snapshotAPIGroup = snapapi.GroupName       // "snapshot.storage.k8s.io"
 	pvcKind          = "PersistentVolumeClaim" // Native types don't require an API group
 
-	tokenPVNameKey       = "pv.name"
-	tokenPVCNameKey      = "pvc.name"
-	tokenPVCNameSpaceKey = "pvc.namespace"
+	tokenPVNameKey                 = "pv.name"
+	tokenPVCNameKey                = "pvc.name"
+	tokenPVCNameSpaceKey           = "pvc.namespace"
+	tokenPVAnnoationTemplatePrefix = "pvc.annotations"
```
andyzhangx, line 127: it's `tokenPVCAnnoationTemplatePrefix `

vinli-cn: thanks for pointing that out, fixed

andyzhangx: still not fixed? `tokenPVAnnoationTemplatePrefix`

andyzhangx (commented):
pls add ut for this PR, and also I found `resolveTemplate` does not have any ut, I think you could submit a separate PR for resolveTemplate ut

xing-yang:
In the past, we have rejected enhancement requests like this for portability reasons.  I think we need to have a discuss on this.
/hold

xing-yang:
See this issue here: https://github.com/kubernetes-csi/external-provisioner/issues/86

andyzhangx:
> See this issue here: #86

@xing-yang this is not [Adding PVC's annotation to CreateVolumeRequest.Parameters](https://github.com/kubernetes-csi/external-provisioner/issues/86#top), and since `csi.storage.k8s.io/node-publish-secret-name` already supports `pvc.annotations` translation, this PR extends the translation support for other storage class fields, as long as it's defined as `${pvc.annotations['team.example.com/key']}`, it would be translated for other fields.

```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: fast-storage
provisioner: csi-driver.team.example.com
parameters:
  type: pd-ssd
  csi.storage.k8s.io/node-publish-secret-name: ${pvc.annotations['team.example.com/key']}
  csi.storage.k8s.io/node-publish-secret-namespace: ${pvc.namespace}
```

vinli-cn:
> See this issue here: #86

As Andy mentioned above, external-provisioner can replace templates in secret parameters in Storage Class with values from PVC annotations already, and this PR extends similar "translation" for other parameters in Storage Class, eg: ${pvc.annotations['subDirInPVC']} in my example. We've seen lots of user cases from community.

This PR doesn't have any correlation with #86. In fact, I've sent below PR for similar issue in #86 but it's being discussed now.
https://github.com/kubernetes-csi/external-provisioner/pull/800

jsafrane:
I understand that passing annotations (or any other per-PVC configuration) to CSI driver can have mane useful use cases and is frequently asked for, my personal favorite is a way to tag volumes in a cloud by user (PVC author) provided tags. 

However, it's a double edged sword.

1. As mentioned above, it makes PVCs not portable. This is a huge change, so far you could take PVCs to any Kubernetes with any CSI driver and they would work. All platform specific info was in StorageClass. How a random Helm chart author of a stateful application knows *what* annotations a  particular CSI driver supports? So far they needed to care only about StorageClass name.

2. It opens doors for security issues. For example, https://github.com/kubernetes-csi/csi-driver-smb/issues/428 - with that feature, anyone who can create a PVC can actually read data of other PVs that are dynamically provisioned from the same StorageClass, just by guessing the subdirectory name. It may work well in clusters where users trust each other, but it absolute no-go for multi-tenant clusters. It's not obvious from StorageClass parameter description that it lowers security. My experience is that CSI driver authors do not care or do not understand security too much and we end up with very insecure clusters.

Not that I'd like to have some way how to pass custom parameters from PVC to PV and to CSI driver, however, it must be done carefully. I don't have a magic solution that would solve both problems mentioned above, but it may exist.

andyzhangx:
> I understand that passing annotations (or any other per-PVC configuration) to CSI driver can have mane useful use cases and is frequently asked for, my personal favorite is a way to tag volumes in a cloud by user (PVC author) provided tags.
> 
> However, it's a double edged sword.
> 
> 1. As mentioned above, it makes PVCs not portable. This is a huge change, so far you could take PVCs to any Kubernetes with any CSI driver and they would work. All platform specific info was in StorageClass. How a random Helm chart author of a stateful application knows _what_ annotations a  particular CSI driver supports? So far they needed to care only about StorageClass name.
> 2. It opens doors for security issues. For example, [Cannot mount an existing share via subDir parameter csi-driver-smb#428](https://github.com/kubernetes-csi/csi-driver-smb/issues/428) - with that feature, anyone who can create a PVC can actually read data of other PVs that are dynamically provisioned from the same StorageClass, just by guessing the subdirectory name. It may work well in clusters where users trust each other, but it absolute no-go for multi-tenant clusters. It's not obvious from StorageClass parameter description that it lowers security. My experience is that CSI driver authors do not care or do not understand security too much and we end up with very insecure clusters.
> 
> Not that I'd like to have some way how to pass custom parameters from PVC to PV and to CSI driver, however, it must be done carefully. I don't have a magic solution that would solve both problems mentioned above, but it may exist.

@jsafrane thanks for the comments. Regrading to point 1 & 2, I am thinking is it possible to add a feature flag for this feature(by default disable), so if users want this feature, they could turn it on in csi-provisioner themselves, by that way, this feature won't affect other csi-provisioner users, and it could also satisfy specific csi driver users.

PR #846, opened 2022-12-28
Seed random generator before creating provisioner identity

### LINKED ISSUE #845: CSI provider identity doesn't seed random generator
I noticed several CSI drivers have `8081` as part of their random identity as seen in the `volumeAttributes` field in PVs.

[Relevant code](https://github.com/kubernetes-csi/external-provisioner/blob/1fcfdc84579b0ccf3918bcc9fb791a741ef87c77/cmd/csi-provisioner/csi-provisioner.go#L283):

```go
	// Generate a unique ID for this provisioner
	timeStamp := time.Now().UnixNano() / int64(time.Millisecond)
	identity := strconv.FormatInt(timeStamp, 10) + "-" + strconv.Itoa(rand.Intn(10000)) + "-" + provisionerName
```

I don't think this code seeds the Go math random generator, which will always generate `8081`:

```go
package main

import (
        "fmt"
        "math/rand"
        "strconv"
)

func main() {
        fmt.Println(strconv.Itoa(rand.Intn(10000)))
}
```
```shell
$ go run main.go
8081
```

The Go `math/random` generator needs to be seeded properly, otherwise it will generate the same sequence, in this case with an initial value of `8081`, which is not what you want for this code.

**Environment**:
- Driver version: v3.1.0
- Kubernetes version (use `kubectl version`): 1.20, 1.25
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

wvh:
The Go math/random package needs to be seeded otherwise it will generate the same sequence over and over.

Without seeding, all provisioners will have `8081` as part of their random ID, as that is the first number generated by the default sequence.

Use the millisecond timeStamp created just above as seed. This still links the random number to the time stamp, but that's probably good enough for its purpose.

/kind bug

Fixes #845

```release-note
NONE
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -280,6 +280,7 @@ func main() {
 
 	// Generate a unique ID for this provisioner
 	timeStamp := time.Now().UnixNano() / int64(time.Millisecond)
+	rand.Seed(timeStamp)
 	identity := strconv.FormatInt(timeStamp, 10) + "-" + strconv.Itoa(rand.Intn(10000)) + "-" + provisionerName
 	if *enableNodeDeployment {
 		identity = identity + "-" + node


cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -280,6 +280,7 @@ func main() {
 
 	// Generate a unique ID for this provisioner
 	timeStamp := time.Now().UnixNano() / int64(time.Millisecond)
+	rand.Seed(timeStamp)
```
pohly, line 283: I agree that `rand.Intn` wasn't useful. But is this any better?

The randomness of `identity` still depends entirely on `timeStamp` - if we can't improve that, then we might as well drop the `rand.Intn` part.

Does Go have an API for the system's random number pool?

wvh: The standard library has `math/rand`, which is not cryptographically secure, and `crypto/rand`, which should just read from `/dev/random` on Linux. The easiest option would be to read 2-4 bytes of dev random, hex encode them and use that instead.

Finding another seed value for math random is another option, but the most likely choices I can see are time-based and the driver name, which are already there in that string.

wvh: ... And process id is another classic option, which makes the ID string unlikely to collide combined with the time-based int already in the string.

pohly: I like using `crypto/rand`.

ggriffiths: +1 for `crypto/rand`, I have mostly seen that used to solve this issue with `math/rand`

ggriffiths (changes requested):
Suggested to use `crypto/rand` as per the earlier thread

wvh:
I can add a simple hex hash using `crypto/rand` instead of the integer. Do we still need to keep the timestamp then? I assume the purpose is uniqueness, no requirement for order.

(The current string has the form `ts-rand-driverName...`.)

pohly:
I'm in favor of removing the time stamp.

PR #867, opened 2023-02-07
Add tests for volume mode conversion feature

RaunakShah:
**What type of PR is this?**
/kind test

**What this PR does / why we need it**:

Follow up to https://github.com/kubernetes-csi/external-provisioner/pull/832
This adds further e2e tests described in the feature KEP - https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/3141-prevent-volume-mode-conversion#e2e-tests

Test output:

```
provision volumes with different volume modes from volume snapshot dataSource when the source volume mode is altered with permissions
/Users/raunakshah/go/src/github.com/kubernetes-csi/external-provisioner/test/e2e/storage/provision.go:154
  STEP: Creating a kubernetes client @ 02/07/23 14:18:59.983
  Feb  7 14:18:59.983: INFO: >>> kubeConfig: /Users/raunakshah/.kube/config
  STEP: Building a namespace api object, basename pvcs-from-volume-snapshots @ 02/07/23 14:18:59.99
  STEP: Waiting for a default service account to be provisioned in namespace @ 02/07/23 14:19:00.002
  STEP: Waiting for kube-root-ca.crt to be provisioned in namespace @ 02/07/23 14:19:00.007
  Feb  7 14:19:00.011: INFO: Running '/opt/homebrew/bin/kubectl --server=https://127.0.0.1:56677 --kubeconfig=/Users/raunakshah/.kube/config --namespace=kube-system describe deployment snapshot-controller'
  Feb  7 14:19:00.113: INFO: stderr: ""
[...]

PR #870, opened 2023-02-13
Rearrange cleanup of PVCs and VolumeSnapshots

RaunakShah:
**What type of PR is this?**
/kind cleanup

Follow up to https://github.com/kubernetes-csi/external-provisioner/pull/867#discussion_r1102800664

Fixes #

```release-note
NONE
```


test/e2e/storage/provision.go
```diff
@@ -85,6 +75,17 @@ var _ = ginkgo.Describe("provision volumes with different volume modes from volu
 				}
 			}
 		}
+		for _, claim := range local.pvcs {
```
msau42, line 78: Are these tests intended only to be run with CSI hostpath driver, or with any CSI driver?

For a general test that can be used against any csi driver, it needs to be able to support any ordering of deleting snapshots and pvcs. So what we've done with our generic test suite is to delete both PVCs and VolumeSnapshots in parallel, and then wait for both of them to complete.

cc @xing-yang

PR #901, opened 2023-03-28
Update dependency go modules for k8s v1.27.0-rc.0 and v1.27.0

sunnylovestiramisu:
Ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.27.0-rc.0
and
ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.27.0

```release-note
Update go version to 1.20
```


sunnylovestiramisu:
This change seems related because it is setting feature flag for tests:

https://github.com/kubernetes/component-base/commit/74434322d4c86da772639eb5cac43816d32e288b

msau42:
Note that the e2e jobs will download Kubernetes at different release versions and build the e2e binaries. So PR above does not seem relevant

```
HEAD is now at b46a3f88 Release commit for Kubernetes v1.26.0
Thu Apr 13 23:14:07 UTC 2023 go1.20.3 /home/prow/go/src/k8s.io/kubernetes$ git clean -fdx
Thu Apr 13 23:14:08 UTC 2023 go1.19 $ make WHAT=test/e2e/e2e.test -C/home/prow/go/src/k8s.io/kubernetes
```

Looking at the test code, `l` is not initialized until the `It` clause: https://github.com/kubernetes/kubernetes/blob/v1.26.0/test/e2e/storage/testsuites/volumeperf.go#L147

So I believe if the test case exits in the `BeforeEach`, it could be possible that the `AfterEach` will run without `l` being initialized.

go.mod
```diff
@@ -17,13 +17,13 @@ require (
...
 	google.golang.org/protobuf v1.30.0
-	k8s.io/api v0.26.3
-	k8s.io/apimachinery v0.26.3
-	k8s.io/apiserver v0.26.1
-	k8s.io/client-go v0.26.1
-	k8s.io/component-base v0.26.1
-	k8s.io/component-helpers v0.26.1
-	k8s.io/csi-translation-lib v0.26.3
+	k8s.io/api v0.27.0-rc.0
+	k8s.io/apimachinery v0.27.0-rc.0
```
msau42, line 21: searching for where the `ServerSideFieldValidation` feature gate is defined:
```
./vendor/k8s.io/apiserver/pkg/features/kube_features.go:        ServerSideFieldValidation: {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.29
./vendor/k8s.io/kubernetes/pkg/features/kube_features.go:       genericfeatures.ServerSideFieldValidation: {Default: true, PreRelease: featuregate.Beta},
```

I see that the same feature gate is defined but with different PreRelease values.

I think the kubernetes dependency at L36 also needs to be updated.

sunnylovestiramisu: but in 1.27 branch it is the same value: https://github.com/kubernetes/kubernetes/blob/release-1.27/pkg/features/kube_features.go#L1194

msau42: I mean this line should also be updated to 1.27: https://github.com/kubernetes-csi/external-provisioner/blob/master/go.mod#L36

The script is supposed to do this; seems like there's a bug

go.mod
```diff
@@ -17,13 +17,13 @@ require (
...
 	google.golang.org/grpc v1.54.0
 	google.golang.org/protobuf v1.30.0
-	k8s.io/api v0.26.3
-	k8s.io/apimachinery v0.26.3
-	k8s.io/apiserver v0.26.1
-	k8s.io/client-go v0.26.1
-	k8s.io/component-base v0.26.1
-	k8s.io/component-helpers v0.26.1
-	k8s.io/csi-translation-lib v0.26.3
+	k8s.io/api v0.27.0-rc.0
```
msau42, line 20: should we update to 0.27.0 now that it's available?

go.mod
```diff
@@ -17,13 +17,13 @@ require (
...
 	google.golang.org/protobuf v1.30.0
-	k8s.io/api v0.26.3
-	k8s.io/apimachinery v0.26.3
-	k8s.io/apiserver v0.26.1
-	k8s.io/client-go v0.26.1
-	k8s.io/component-base v0.26.1
-	k8s.io/component-helpers v0.26.1
-	k8s.io/csi-translation-lib v0.26.3
+	k8s.io/api v0.27.0
+	k8s.io/apimachinery v0.28.0-alpha.0
```
xing-yang, line 21: 0.27.0?

go.mod
```diff
@@ -101,73 +117,75 @@ require (
...
+replace k8s.io/apimachinery => k8s.io/apimachinery v0.27.0
 
-replace k8s.io/apiserver => k8s.io/apiserver v0.26.1
+replace k8s.io/apiserver => k8s.io/apiserver v0.27.0
 
-replace k8s.io/client-go => k8s.io/client-go v0.26.1
+replace k8s.io/client-go => k8s.io/client-go v0.27.0
 
-replace k8s.io/code-generator => k8s.io/code-generator v0.26.1
+replace k8s.io/code-generator => k8s.io/code-generator v0.27.1
```
xing-yang, line 149: 0.27.0?

PR #949, opened 2023-06-26
chore: remove refs to deprecated io/ioutil

testwill:
**What type of PR is this?**
> /kind cleanup

"io/ioutil" has been deprecated since Go 1.19: As of Go 1.16

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -19,7 +19,6 @@ package controller
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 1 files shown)


pkg/controller/controller_test.go
```diff
@@ -125,7 +124,7 @@ func createMockServer(t *testing.T, tmpdir string) (*gomock.Controller,
 }
 
 func tempDir(t *testing.T) string {
-	dir, err := ioutil.TempDir("", "external-provisioner-test-")
+	dir, err := os.MkdirTemp("", "external-provisioner-test-")
```
jsafrane, line 127: Perhaps `t.TempDir()` and remove `defer` somewhere below?

jsafrane: @testwill can you please fix this? The pr is OK otherwise.

jsafrane: I am terribly sorry, I was not aware that go does not support unix domain socket paths longer than 107 characters: https://github.com/golang/go/issues/6895.

`t.TempDir` creates paths with the test name in them and they're quite long. You can see the test failures in CI that I was waiting for you to fix.

Can you please go back to `os.MkdirTemp`?

PR #999, opened 2023-08-24
Update dependency go modules for k8s v1.28.0

Sneha-at:
Ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.28.0


go.mod
```diff
@@ -17,13 +17,13 @@ require (
 	github.com/stretchr/testify v1.8.4
 	google.golang.org/grpc v1.57.0
 	google.golang.org/protobuf v1.31.0
-	k8s.io/api v0.28.0-rc.0
```
sunnylovestiramisu, line 20: github.com/kubernetes-csi/csi-lib-utils v0.14.0 we should update this to v0.15.0

go.mod
```diff
@@ -139,57 +139,3 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.3.0 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
-
-replace k8s.io/api => k8s.io/api v0.28.0-rc.0
```
msau42, line 143: Did the script remove this?

Sneha-at: No, I removed all the replace statements as we do for all other sidecars. Should these be kept there for external-provisioner?

go.mod
Sneha-at: @msau42 @xing-yang should we use `csi-lib-utils:v0.15.0-rc0` instead? It does not have the changes for Optel package.

go.mod
msau42: I think if the script is adding it let's just keep it

go.mod
msau42: I think we can stay on v0.14.0 which should be equivalent

Sneha-at: Ok sounds good. Resolving the comment as no further changes are needed

go.mod
Sneha-at: Added these statements back in.

PR #1013, opened 2023-08-30
Group the rest of dependencies in dependabot

sunnylovestiramisu:
**What type of PR is this?**
/kind cleanup

Group the rest of dependencies so it does not open a single dependency update similar to this PR https://github.com/kubernetes-csi/external-provisioner/pull/1007

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -17,9 +17,11 @@ updates:
         - "sigs.k8s.io*"
     github-dependencies:
       patterns:
-        - "github.com*"
+        - "*"
       exclude-patterns:
         - "github.com/golang*"
+        - "k8s.io*"
+        - "sigs.k8s.io*"
   labels:
     - "area/dependency"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -20,6 +20,13 @@ updates:
         - "github.com*"
       exclude-patterns:
         - "github.com/golang*"
+    rest-dependencies:
+      patterns:
+        - "*"
```
msau42, line 25: I would just put these under the github-dependencies

sunnylovestiramisu: We already have a group named github-dependencies. Do we want to combine these two?

```
github-dependencies:
      patterns:
        - "github.com*"
      exclude-patterns:
        - "github.com/golang*"
     ```

sunnylovestiramisu: Updated based on my guess :)

PR #1068, opened 2023-10-26
Pass VolumeAttributesClass Parameters as MutableParameters to CreateVolume

### LINKED ISSUE #952: External Provisioner Change
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-e776b8fa-7fff-5a33-682c-79ecb0343e5e"><div dir="ltr" style="margin-left:0pt;" align="left">

Tasks
--
0. Feature gate for provisioner specific VolumeAttributesClass change
1. Add support for provision PVC with a VolumeAttributesClass

</div></b>

ConnorJC3:
**What type of PR is this?**

/kind feature

Implements the external-provisioner component of https://kep.k8s.io/3751

Fixes #952

There are several places where I was unsure of the best action, marked with "TODO" comments, please answer those if you have a better answer than I do.

Breakdown by commit:

- _Update to go1.21_  
Required because the Kubernetes dependency that will be upgrading in the following commit references the go1.21-only slices library.

- _[TODO FIX AFTER PRS MERGE] Bump CSI spec, csi-test, and Kubernetes dependencies_
  * Upgrades all k8s dependencies directly in k/k to the versions from https://github.com/kubernetes/kubernetes/pull/121104
  * Upgrades csi-test to the version from https://github.com/kubernetes-csi/csi-test/pull/492
  * Upgrades the CSI spec to v1.9.0

- _Add VolumeAttributesClass feature flag_  
Adds a default-off VAC feature flag.

- _Pass VolumeAttributesClass Parameters as MutableParameters to CreateVolume_  
The meat of the PR: Adds the funcionality for VACs, as well as associated unit tests.

```release-note
TODO will fill in later, unsure if this should be coordinated with other PRs
```


ConnorJC3:
Chasing down CI failures, but I believe they are largely (if not entirely) false positives caused by issues with the go version - this PR should be reasonably safe to review now, I don't anticipate any large or architectural changes.

pkg/controller/controller.go
```diff
@@ -596,6 +604,14 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		}
 	}
 
+	vacName := claim.Spec.VolumeAttributesClassName
+	// TODO: For reviewers - Should we reject volumes that have a VAC specified if the VAC feature gate is disabled?
```
sunnylovestiramisu, line 608: The csi spec states that "A Controller plugin MUST implement this RPC call if the plugin has the MODIFY_VOLUME controller capability. " 

To consider backward compatibility I do not think we should reject the call. Maybe if there is a way to send a warning to users indicate that the VAC is not applied because of feature gate etc.

pkg/features/features.go
```diff
@@ -41,6 +41,14 @@ const (
...
 	// Enable usage of Provision of PVCs from snapshots in other namespaces
 	CrossNamespaceVolumeDataSource featuregate.Feature = "CrossNamespaceVolumeDataSource"
+
+	// TODO: What are the appropriate owners here? Never implemented a feature flag before, did my best guess
+	// owner: @sunnylovestiramisu @ConnorJC3
+	// kep: https://kep.k8s.io/3751
+	// alpha: v1.29
+	//
+	// Pass VolumeAttributesClass parameters to supporting CSI drivers during CreateVolume
+	VolumeAttributesClass featuregate.Feature = "VolumeAttributesClass"
```
sunnylovestiramisu, line 51: The question here is also do we want a diff feature gate for the sidecars from the k8s feature gate. @msau42

pkg/controller/controller.go
```diff
@@ -742,6 +758,19 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		deletionAnnSecrets.namespace = provisionerSecretRef.Namespace
 	}
 
+	if utilfeature.DefaultFeatureGate.Enabled(features.VolumeAttributesClass) && vacName != nil {
+		vac, err := p.client.StorageV1alpha1().VolumeAttributesClasses().Get(ctx, *vacName, metav1.GetOptions{})
+		if err != nil {
+			return nil, controller.ProvisioningNoChange, err
```
sunnylovestiramisu, line 764: We were saying that even if the VAC does not exist yet, we still want to make sure later to retry provisioning. I just noticed that there seems no appropriate ProvisioningState right now for this. The closest is ProvisioningReschedule . Should we defer this to beta? @msau42

pkg/controller/controller.go
msau42: Why would it not be backwards compatible? A user would have to opt-in to specifying the VAC in their PVC.

sunnylovestiramisu: What if they downgrade their cluster and turn off the vac feature but the VAC is applied?

pkg/controller/controller.go
sunnylovestiramisu: Seems like the reconciling is happening in https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/master/controller/controller.go we may want to make some changes there too.

ConnorJC3: Would `ProvisioningInBackground` be better here? By my reading that should always cause a re-drive of provisioning after some delay.

pkg/controller/controller.go
ConnorJC3: Kubernetes doesn't support downgrades, although I'm not 100% sure what would happen if you disabled the API Server feature gate on a cluster with VACs already in it

pohly: Kubernetes supports downgrades. 

However, even with the help of the folks on #sig-architecture the best link that I could find was only the underlying design for API storage: https://kubernetes.io/docs/reference/using-api/deprecation-policy/

ConnorJC3: Maybe "doesn't support" is too strong of wording, but Kubernetes does not document any support for downgrading and it is widely considered unsafe: https://github.com/kubernetes/website/issues/12327

That said, I'm not sure it really matters for the purpose of this question: My question is basically, if the volume does specify a `VolumeAttributeClassName` (i.e. the feature flag is enabled at the API level and the user is using it) but the local `external-provisioner` feature flag is disabled, what should we do?

Should we either (1) provision the volume as normal - ignoring the fact it has a VAC or (2) refuse to provision the volume and demand the user turns on the VAC feature flag or remove the VAC from the volume?

ConnorJC3:
Bumped to use k8s 1.29.0-alpha.3 version. CI still broken because it uses go 1.20

pkg/controller/controller.go
```diff
@@ -596,6 +604,14 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		}
 	}
 
+	vacName := claim.Spec.VolumeAttributesClassName
+	// TODO: For reviewers - Should we reject volumes that have a VAC specified if the VAC feature gate is disabled?
+	// This will be a change in behavior (existing versions of external-provisioner out in the wild will just let those PVCs get provisioned),
+	// but it could prevent provisioning a PVC in a misleading way
+	if utilfeature.DefaultFeatureGate.Enabled(features.VolumeAttributesClass) && vacName != nil {
```
carlory, line 611: vacName maybe an empty string. for this case, modifyVolume should be false.

pkg/controller/controller.go
```diff
@@ -742,6 +758,19 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		deletionAnnSecrets.namespace = provisionerSecretRef.Namespace
 	}
 
+	if utilfeature.DefaultFeatureGate.Enabled(features.VolumeAttributesClass) && vacName != nil {
```
carlory, line 761: vacName maybe an empty string

pkg/controller/controller.go
```diff
@@ -742,6 +758,19 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
...
+		vac, err := p.client.StorageV1alpha1().VolumeAttributesClasses().Get(ctx, *vacName, metav1.GetOptions{})
+		if err != nil {
+			return nil, controller.ProvisioningNoChange, err
+		}
+
+		if vac.DriverName != p.driverName {
```
carlory, line 767: If an error occurs, the pvc will stuck in Pending state. users how to know the reason and how to fix it? It seems that the error message won't be found in the pvc's status field. Is it possible to omit an event to the pvc?

pkg/controller/controller.go
sunnylovestiramisu: This is the provisioner, it does not call modify volume

pkg/controller/controller.go
sunnylovestiramisu: We need to check existence of the VAC, not only if the vacName is an empty string or not.

pkg/controller/controller.go
sunnylovestiramisu: +1 to omit an event. We need to retry on this as well.

pkg/controller/controller.go
carlory: Do you means that the requiredCapabilities shouldn't contain the modifyVolume field? I know that the provsioner doesn't call the `ControllerModifyVolume` rpc. but its SPEC said that the MODIFY_VOLUME is required capability  when users create a pvc with a vac.

```
  // Plugin specific creation-time parameters passed in as opaque 
  // key-value pairs. These mutable_parameteres MAY also be
  // changed during the lifetime of the volume via a subsequent
  // `ControllerModifyVolume` RPC. This field is OPTIONAL.
  // The Plugin is responsible for parsing and validating these
  // parameters. COs will treat these as opaque.
  
  // Plugins MUST treat these 
  // as if they take precedence over the parameters field.
  // This field SHALL NOT be specified unless the SP has the
  // MODIFY_VOLUME plugin capability.
  map<string, string> mutable_parameters = 8 [(alpha_field) = true];
```

ConnorJC3: Added check against `*vacName == ""` in latest version

pkg/controller/controller.go
ConnorJC3: Added check against `*vacName == ""` in latest version

pkg/controller/controller.go
ConnorJC3: Newest version records an event here if the PVC and VAC's driver names are mismatched. Existing code should drive a retry as normal with provisioning failures.

ConnorJC3:
Rebased on latest `master` branch as well as incorporated the most recent feedback

Makefile
```diff
@@ -15,5 +15,7 @@
 CMDS=csi-provisioner
 all: build
 
+# TODO: Change in csi-release-tools before PR merge?
+export CSI_PROW_GO_VERSION_BUILD=1.21
```
jsafrane, line 19: Yes, please send a PR for csi-release-tools. All sidecar releases for 1.29 will need a new go.

pkg/controller/controller.go
```diff
@@ -742,6 +758,21 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
...
+			mismatchErrorMsg := fmt.Sprintf("VAC %s referenced in PVC is for driver %s which does not match driver name %s", *vacName, vac.DriverName, p.driverName)
+			p.eventRecorder.Event(claim, v1.EventTypeWarning, "Provisioning", mismatchErrorMsg)
```
jsafrane, line 769: sig-storage-lib-external-provisioner [will already send an event](https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/cd540b4cb6c7b01eaad8434c273d00aa13ce9081/controller/controller.go#L1474) when `Provision()` returns an error, no need to send another one here.

go.mod
```diff
@@ -1,9 +1,9 @@
 module github.com/kubernetes-csi/external-provisioner
 
-go 1.20
+go 1.21.3
```
jsafrane, line 3: This must be `1.21`, go mod does not allow patch version. See failed unit tests in this PR.

jsafrane: Pardon my ignorance, 1.21.3 is valid in 1.21. Still, all CI jobs fail because they uses 1.20.

pkg/controller/controller.go
ConnorJC3: Resolving this, reverted the change because of https://github.com/kubernetes-csi/external-provisioner/pull/1068#discussion_r1392414749

pkg/controller/controller.go
ConnorJC3: Thinking about it more deeply, I agree. As far as I know, there is no way to prevent a retry here - the external-provisioner will _always_ eventually retry. And as this error is potentially temporal (for example, if a networking issue blocks our access to the k8s api briefly), there is no way to know if a provisioning action is already ongoing or not.

pkg/controller/controller.go
ConnorJC3: Much better than my version, I stole it wholesale for the next revision

pkg/controller/controller_test.go
```diff
@@ -2282,6 +2303,130 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 			expectNoProvision:  true, // not owner and will not change that either
 			expectSelectedNode: "",   // not changed by ShouldProvision
 		},
+		"normal provision with VolumeAttributesClass": {
+			featureGates: map[featuregate.Feature]bool{
+				features.VolumeAttributesClass: true,
```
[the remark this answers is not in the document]
ConnorJC3: Added in latest revision

Makefile
ConnorJC3: Working on a PR, for now I modified the directory directly to hopefully unblock CI, although this will cause `make test` to fail because it detects non-upstreamed changes

jsafrane:
/approve
LGTM-is except for the release-tools hack.

ConnorJC3:
csi-release-tools PR here: https://github.com/kubernetes-csi/csi-release-tools/pull/238

jsafrane:
release-tools have go 1.21 now

PR #1125, opened 2023-12-05
Refactor (controller_tests): build the test case objects with all the information

### LINKED ISSUE #376: Refactor controller_tests to use builder pattern

hshitomi:
**What type of PR is this?**

/kind cleanup

Remove the conditional statements that customize the input test case object based on the different test scenarios in the function of running a test.

The function to build the test cases should build the test case objects with all the information that the function of running a test needs so that it makes the test code easier to read and understand.

Fixes #376

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -840,20 +840,13 @@ func TestGetSecretReference(t *testing.T) {
 type provisioningTestcase struct {
 	capacity                  int64 // if zero, default capacity, otherwise available bytes
 	volOpts                   controller.ProvisionOptions
-	notNilSelector            bool
-	makeVolumeNameErr         bool
-	getSecretRefErr           bool
-	getCredentialsErr         bool
-	volWithLessCap            bool
-	volWithZeroCap            bool
+	expectedMockReturn        func(t *testing.T, out *csi.CreateVolumeResponse, controllerServer *driver.MockControllerServer, createVolumeError error)
 	expectedPVSpec            *pvSpec
 	clientSetObjects          []runtime.Object
 	createVolumeError         error
 	expectErr                 bool
 	expectState               controller.ProvisioningState
-	expectCreateVolDo         func(t *testing.T, ctx context.Context, req *csi.CreateVolumeRequest)
 	withExtraMetadata         bool
-	skipCreateVolume          bool
 	deploymentNode            string // fake distributed provisioning with this node as host
 	immediateBinding          bool   // enable immediate binding support for distributed provisioning
 	expectSelectedNode        string // a specific selected-node of the PVC in the apiserver after the test, same as before if empty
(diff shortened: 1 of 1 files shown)


hshitomi:
@mauriciopoppe Let me start our conversation before I completely refactor the code.

My current focus is on the provisioning test cases. I have removed the if statements from `runProvisionTest()`. Could you please take a look? I want to confirm that I am on the right track.

Since I rewrote the test cases using method chaining (in `provisionTestcase()`), a lot of code has changed. However, as you can see, the core part of the change is not big and is around `provisioningTestcaseOption struct`, `TestProvision()` function and `TestShouldProvision()` function.

For now, I am using the method chaining rather than the functional option pattern because it seems a bit simpler to me. I know that the method to build the test case objects with all the information does not matter much. But if you prefer to use the functional option pattern, please let me know.

Thank you in advance for your time and guidance. Looking forward to your feedback.

pkg/controller/controller_test.go
```diff
@@ -1090,8 +1198,8 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 	}
 	vacName := "test-vac"
 	return requestedBytes, map[string]provisioningTestcase{
-		"normal provision": {
-			volOpts: controller.ProvisionOptions{
+		"normal provision": *provisionTestcaseBuilder().volOpts(
```
mauriciopoppe, line 1201: I think the builder pattern should be applied to things we want to build in an object to then test it externally. With this implementation both the things to build and the expectations are part of the same object.

Maybe `provisioningTestCase` shouldn't have setters for expectations.

mauriciopoppe, line 1201: I think it became a little bit harder to read with chaining, compare:

```
makeProvisionTestcaseBuilder().volOpts(
  ...
).foo(
  ...
).bar(
  ...
).build()
```

vs

```
makeProvisionTestcaseBuilder(
  withFoo(),
  withBar(),
  withVolOpts()
).build()
```

hshitomi: Thank you for your feedback. Let me confirm your suggestion.

We should create another function to set the expectations. The function is to set the following elements of the `provisioningTestcase` struct, which is the expectation of the test result.
```
	expectedPVSpec            *pvSpec
	expectedMockReturn        func(t *testing.T, out *csi.CreateVolumeResponse, controllerServer *driver.MockControllerServer, createVolumeError error)
	expectErr                 bool
	expectState               controller.ProvisioningState
	expectSelectedNode        string // a specific selected-node of the PVC in the apiserver after the test, same as before if empty
	expectNoProvision         bool   // if true, then ShouldProvision should return false
```

At the same time, the `provisionTestcase()` function should be used to build the test case, which means to set the following elements of the `provisioningTestcase` struct.
```
	capacity                  int64 // if zero, default capacity, otherwise available bytes
	volOpts                   controller.ProvisionOptions
	clientSetObjects          []runtime.Object
	createVolumeError         error
	withExtraMetadata         bool
	deploymentNode            string // fake distributed provisioning with this node as host
	immediateBinding          bool   // enable immediate binding support for distributed provisioning
	controllerPublishReadOnly bool
	featureGates              map[featuregate.Feature]bool
	pluginCapabilities        func() (rpc.PluginCapabilitySet, rpc.ControllerCapabilitySet)
```

Did I understand your suggestion properly?

hshitomi: OK. Let me implement it using the functional option pattern.

hshitomi: I split the first three test cases into builder (`buildProvisionTestcase()`) and expectation setter (`setProvisionTestcaseExpectation()`). Is this what you intended?

Or is the intention to further divide the `provisioningTestcase struct` into two, as shown below, and rewrite related parts based on that?
```
type provisioningTestcase struct {
	capacity                  int64 // if zero, default capacity, otherwise available bytes
	volOpts                   controller.ProvisionOptions
	clientSetObjects          []runtime.Object
	createVolumeError         error
	withExtraMetadata         bool
	deploymentNode            string // fake distributed provisioning with this node as host
	immediateBinding          bool   // enable immediate binding support for distributed provisioning
	controllerPublishReadOnly bool
	featureGates              map[featuregate.Feature]bool
	pluginCapabilities        func() (rpc.PluginCapabilitySet, rpc.ControllerCapabilitySet)
}

type provisioningTestcaseExpectation struct {
	expectedPVSpec            *pvSpec
	expectedMockReturn        func(t *testing.T, out *csi.CreateVolumeResponse, controllerServer *driver.MockControllerServer, createVolumeError error)
	expectErr                 bool
	expectState               controller.ProvisioningState
	expectSelectedNode        string // a specific selected-node of the PVC in the apiserver after the test, same as before if empty
	expectNoProvision         bool   // if true, then ShouldProvision should return false
}
```

hshitomi: I wrote the first three test cases with the functional option pattern.

As you said, the test case setup part in `buildProvisionTestcase()` has become a bit simpler. On the other hand, the functions for the functional option pattern seem to be a bit more complicated than the methods for method chaining. That is why I chose the method chaining. However, I guess the simplicity of the test case setup part might be more important. So, I will use the functional option pattern for the rest of the test cases.

What do you think?

hshitomi:
@mauriciopoppe @msau42 I'm sure you're busy, but I'm looking forward to your feedback. If possible, could you please let me know when you can get to the review?

PR #1126, opened 2023-12-10
Enable prevent-volume-mode-conversion by default following k8s 1.28 release

akalenyu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This is beta since 1.27 so we can safely default to true, to get this new validation logic. https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/3141-prevent-volume-mode-conversion/kep.yaml#L30

Fixes #

```release-note
Enable prevent-volume-mode-conversion feature flag by default.
ACTION REQUIRED: 
Volume mode change will be rejected when creating a PVC from a VolumeSnapshot unless the AllowVolumeModeChange annotation has been set to true. Applications relying on volume mode change when creating a PVC from VolumeSnapshot need to be updated accordingly.
```


akalenyu:
@xing-yang @RaunakShah I think keeping this is fine as `sed` will return 0 but just making sure https://github.com/kubernetes-csi/external-provisioner/blob/680aa9e07657385bf5de168475f20fbe278de2a3/.prow.sh#L39-L40

PR #1152, opened 2024-01-24
[WIP] fix: properly shutdown topology informer on context cancellation

### LINKED ISSUE #1151: Goroutine Leak in Node Topology Informer due to missing informer queue cancellation
**What happened**:

When creating The Capacity Controller there is a [Queue Shutdown happening for its queue](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/capacity/capacity.go#L244). This leads to the shutdown causing the [processNextWorkItem Queue to cancel correctly](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/capacity/capacity.go#L523-L533).

However, the Capacity Controller expects a [Node Topology Informer](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/capacity/capacity.go#L172). This informer is started in the same way and is compromised of the same [processNextWorkItem Queue logic](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/capacity/topology/nodes.go#L223-L227). However, [the Node Topology Controller Queue never gets called for shutdown on context cancellation](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/capacity/topology/nodes.go#L48). 

This leads to the fact that this controller never correctly shuts down and processes remaining work items and can lead to context hangups if waited for correctly.

The Node Topology informer queue should be shut down similarly to the Capacity Informer queue.

**How to reproduce it**:

Instead of [the current code](https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L486):
```go

	    if *enableCapacity {
                  go topologyInformer.RunWorker(ctx)
            }
            // ...
            if capacityController != nil {
                  go capacityController.Run(ctx, int(*capacityThreads))
            }
            if csiClaimController != nil {
                  go csiClaimController.Run(ctx, int(*finalizerThreads))
            }
            provisionController.Run(ctx)
```

Properly run the controllers in a setup with a wait group:
```go
            var wg sync.WaitGroup
	    if *enableCapacity {
	          wg.Add(1)
                  go func() {
		        defer wg.Done()
                        topologyInformer.RunWorker(ctx)
		        logger.Info("topology informer finished shutdown")
                  }
            }
            // ...
          
            if capacityController != nil {
	          wg.Add(1)
                  go func() {
		        defer wg.Done()
                        capacityController.Run(ctx, int(*capacityThreads))
		        logger.Info("capacity controller finished shutdown")
                  }
            }
            if csiClaimController != nil {
	          wg.Add(1)
                  go func() {
		        defer wg.Done()
                        csiClaimController.Run(ctx, int(*finalizerThreads))
		        logger.Info("capacity controller finished shutdown")
                  }
            }

            wg.Add(1)
            go func() {
		 defer wg.Done()
                 provisionController.Run(ctx)
		 logger.Info("provisioner controller finished shutdown")
            }            

	    wg.Wait()
	    
	    logger.Info("all controllers finished shutdown")
```

[instead of running the controllers like here](https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L640-L646). In the new code it is verified that the controllers actually exit, and there is no goroutine leak if the queue is stopped. However to stop the queue we will have to stop the queue from the given capacity controller. This is currently not done. 

The entire provisioner should be verified for goroutine leaks due to this finding as this is usually an easy to spot issue. I only noticed this because I manually started the controllers in a different embedded application and noticed that the controller did not shut down properly.

**Environment**:
- Driver version: Applicable to all versions
- Kubernetes version (use `kubectl version`): Applicable to all k8s versions
- OS (e.g. from /etc/os-release): Applicable to All OS
- Kernel (e.g. `uname -a`): irrelevant
- Install tools: irrelevant
- Others: N/A

jakobmoellerdev:
**What type of PR is this?**
/kind bug

Makes context cancellation clear and resolves the topology informer never shutting down correctly when cancelling its context. Includes shutdown logs to verify this.

Fixes #1151

```release-note
external-provisioner now includes shutdown logs for controllers started.
If consumed as a go dependency, external-provisioner now shuts down the context correctly when stopping the context used for the informers.
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -38,7 +39,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	validation "k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
@@ -127,6 +128,7 @@ var (
 func main() {
 	var config *rest.Config
 	var err error
+	var controllerWg sync.WaitGroup
 
 	flag.Var(utilflag.NewMapStringBool(&featureGates), "feature-gates", "A set of key=value pairs that describe feature gates for alpha/experimental features. "+
 		"Options are:\n"+strings.Join(utilfeature.DefaultFeatureGate.KnownFeatures(), "\n"))
@@ -483,7 +485,13 @@ func main() {
 			klog.Infof("producing CSIStorageCapacity objects with fixed topology segment %s", segment)
 			topologyInformer = topology.NewFixedNodeTopology(&segment)
 		}
-		go topologyInformer.RunWorker(ctx)
+
+		controllerWg.Add(1)
+		go func() {
+			defer controllerWg.Done()
+			topologyInformer.RunWorker(ctx)
+			klog.Info("topology informer stopped")
+		}()
 
 		managedByID := "external-provisioner"
 		if *enableNodeDeployment {
@@ -638,12 +646,31 @@ func main() {
 		}
 
 		if capacityController != nil {
-			go capacityController.Run(ctx, int(*capacityThreads))
+			controllerWg.Add(1)
+			go func() {
+				defer controllerWg.Done()
+				capacityController.Run(ctx, int(*capacityThreads))
+				klog.Info("capacity controller finished shutdown")
+			}()
 		}
 		if csiClaimController != nil {
-			go csiClaimController.Run(ctx, int(*finalizerThreads))
+			controllerWg.Add(1)
+			go func() {
+				defer controllerWg.Done()
+				csiClaimController.Run(ctx, int(*finalizerThreads))
+				klog.Info("csi claim controller finished shutdown")
+			}()
 		}
-		provisionController.Run(ctx)
+
+		controllerWg.Add(1)
+		go func() {
+			defer controllerWg.Done()
+			provisionController.Run(ctx)
+			klog.Info("provisioning controller finished shutdown")
+		}()
+
+		controllerWg.Wait()
+		klog.Info("all controllers finished shutdown")
 	}
 
 	if !*enableLeaderElection {
--- a/pkg/capacity/topology/nodes.go
+++ b/pkg/capacity/topology/nodes.go
@@ -168,6 +168,10 @@ type nodeTopology struct {
 	callbacks []Callback
 }
 
+func (nt *nodeTopology) ShutDown() {
+	nt.queue.ShutDown()
+}
+
 // driverTopologyKeys returns nil if the driver is not running
 // on the node, otherwise at least an empty slice of topology keys.
 func (nt *nodeTopology) driverTopologyKeys(csiNode *storagev1.CSINode) []string {
--- a/pkg/capacity/capacity.go
+++ b/pkg/capacity/capacity.go
@@ -242,6 +242,7 @@ var _ metrics.StableCollector = &Controller{}
 func (c *Controller) Run(ctx context.Context, threadiness int) {
 	klog.Info("Starting Capacity Controller")
 	defer c.queue.ShutDown()
+	defer c.topologyInformer.ShutDown()
 
 	c.prepare(ctx)
 	for i := 0; i < threadiness; i++ {
--- a/pkg/capacity/topology/mock.go
+++ b/pkg/capacity/topology/mock.go
@@ -42,6 +42,8 @@ type Mock struct {
 	callbacks []Callback
 }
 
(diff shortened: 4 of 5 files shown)


jakobmoellerdev:
@pohly mind taking a look and doing a ok-to-test here?

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -638,12 +646,31 @@ func main() {
...
+
+		controllerWg.Add(1)
+		go func() {
+			defer controllerWg.Done()
+			provisionController.Run(ctx)
+			klog.Info("provisioning controller finished shutdown")
+		}()
+
+		controllerWg.Wait()
+		klog.Info("all controllers finished shutdown")
```
pohly, line 673: I don't think any of the changes to `main()` matter because `provisionController.Run(ctx)` is blocking forever until the process exits.

jakobmoellerdev: I think this is only for logging, can revert all of these changes if you dont want this. Just thought it would be proper hygiene.

pkg/capacity/topology/nodes.go
```diff
@@ -168,6 +168,10 @@ type nodeTopology struct {
 	callbacks []Callback
 }
 
+func (nt *nodeTopology) ShutDown() {
+	nt.queue.ShutDown()
```
pohly, line 172: Wouldn't it be better to make the code which creates the queue also shut it down? Then the Informer interface change won't be necessary.

jakobmoellerdev: https://github.com/kubernetes-csi/external-provisioner/blob/0c26e9feb404a4eb807e19ded56819f0e50baa1e/pkg/capacity/topology/nodes.go#L48 the queue is passed externally to the topology informer (from the command). To shut it down from the component that creates it (in cmd), we would still need the interface change I think.

If you mean the code that starts it, This is also external as the Run code just waits for the queue to shutdown: https://github.com/kubernetes-csi/external-provisioner/blob/0c26e9feb404a4eb807e19ded56819f0e50baa1e/pkg/capacity/topology/nodes.go#L208

Maybe I misunderstood you, feel free to elaborate more if you dont think that explains why I did this

pohly: https://github.com/kubernetes-csi/external-provisioner/blob/b377ea40edb757a7c3d8abb7058f701844ebf867/cmd/csi-provisioner/csi-provisioner.go#L474 creates the queue.

IMHO that code should be responsible for shutting it down.

pohly: But in `cmd/csi-provisioner/csi-provisioner.go` that's pointless. If you do the same in your code, does that solve problem without any changes to this repo?

jakobmoellerdev: > https://github.com/kubernetes-csi/external-provisioner/blob/b377ea40edb757a7c3d8abb7058f701844ebf867/cmd/csi-provisioner/csi-provisioner.go#L474
> 
> creates the queue.
> IMHO that code should be responsible for shutting it down.

The queue is never started with this code, so its also not responsible for shutting it down, it just owns the data structure holding the items. We start the informer in the cmd, so we also need to own the context and shut it down from what I see.

> But in cmd/csi-provisioner/csi-provisioner.go that's pointless. If you do the same in your code, does that solve problem without any changes to this repo?

In our embedded code we manually call `topologyInformer.RunWorker(ctx)` so if we expose the Shutdown method, then our code can call the Shutdown of the queue after provisioning controller shutdown and then the topologyInformer will shut down correctly as well.

pohly: There is no explicit starting of the queue, is it?

> In our embedded code we manually call topologyInformer.RunWorker(ctx) so if we expose the Shutdown method, then our code can call the Shutdown of the queue after provisioning controller shutdown and then the topologyInformer will shut down correctly as well.

This still doesn't sound right to me. `topologyInformer.RunWorker` doesn't own any goroutines, therefore `topologyInformer.ShutDown` isn't needed. The caller can call `queue.ShutDown` instead.

Can you show your code?

jakobmoellerdev: RunWorker is itself blocking so it can only reasonably called from code that uses a goroutine for it or exits on process (like you described earlier).
https://github.com/openshift/lvm-operator/blob/main/internal/csi/provisioner.go heres the code we use. Please feel free to give any suggestion how I can workaround this without having to rely on process exit and without adjusting current code.

pohly: How about this?

```
topologyQueue := workqueue.NewRateLimitingQueueWithConfig(rateLimiter, workqueue.RateLimitingQueueConfig{
			Name: "csitopology",
		})
defer topologyQueue.ShutDown()

topologyInformer := topology.NewNodeTopology(
		p.options.DriverName,
		clientset,
		factory.Core().V1().Nodes(),
		factory.Storage().V1().CSINodes(),
		topologyQueue,
	)
```

jakobmoellerdev: Okay so while this did not solve my problem directly you gave me the right path to the idea: https://github.com/openshift/lvm-operator/compare/main...jakobmoellerdev:topology-queue-shutdown?expand=1

This will work in all cases and shutdown right after the controller shuts down to avoid using the queue after shutdown and also because I have a wait group in the code.

I take from your previous comments that the other adjustments are unwanted right? Then I will close. I 100% appreciate you taking the time looking at this but also understand that we dont want to change the current system for just some context stops and logs.

pohly: Good that you figured something out. Yes, let's keep the current code as-is.

PR #1193, opened 2024-04-10
fix: don't pass confusing value to GetCapacityRequest.VolumeCapabilities

zjx20:
**What type of PR is this?**
/kind bug

`external-provisioner` calls `GetCapacity()` with "faked" VolumeCapabilities, but it is confusing for CSI driver implementations to validate the capabilities. For example, the access mode is always `VolumeCapability_AccessMode_UNKNOWN`, but it's unclear for an implementation how to handle an "UNKNOWN" mode. And the access type is always `mount`, it also confuses block-oriented CSI drivers.

The comment says "The standard makes it mandatory to pass something, therefore we pick something rather arbitrarily.", but according to the latest CSI spec, the field is optional. So I think we can simply set an empty value to that field.

https://github.com/container-storage-interface/spec/blob/2696773158b970c65107610d8f3ef4302546eca0/csi.proto#L1050-L1056

Fixes #

```release-note
`GetCapacityRequest.VolumeCapabilities` is now set to `nil` instead of some ambiguous default value when invoking `GetCapacity()`.
```

### DIFF
--- a/pkg/capacity/capacity.go
+++ b/pkg/capacity/capacity.go
@@ -591,17 +591,8 @@ func (c *Controller) syncCapacity(ctx context.Context, item workItem) error {
 
 	req := &csi.GetCapacityRequest{
 		Parameters: sc.Parameters,
-		// The assumption is that the capacity is independent of the
-		// capabilities. The standard makes it mandatory to pass something,
-		// therefore we pick something rather arbitrarily.
-		VolumeCapabilities: []*csi.VolumeCapability{
-			{
-				AccessType: &csi.VolumeCapability_Mount{},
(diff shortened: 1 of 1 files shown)


jsafrane:
@zjx20 the PR looks good, however, it's a behavior change. Please add a release note in the PR description to notify users / CSI driver vendors.

zjx20:
> @zjx20 the PR looks good, however, it's a behavior change. Please add a release note in the PR description to notify users / CSI driver vendors.

Hi @jsafrane, I have updated the PR description, feel free to modify it if it's not good enough.

jsafrane:
Thanks!
/lgtm
/approve

PR #1236, opened 2024-06-21
fix: introduce /v5 module suffix for go.mod

jakobmoellerdev:
changes the go.mod path to add /v5 so that go module is resolved properly

**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Adds the /v5 suffix that was missed on previous releases so that the go.mod resolves properly

Opened as per request on https://kubernetes.slack.com/archives/C09QZFCE5/p1718782253879689

```release-note
action required: go module path was changed to github.com/kubernetes-csi/external-provisioner/v5, adjust accordingly when importing in go
```


jakobmoellerdev:
@msau42 I think https://github.com/kubernetes-csi/external-provisioner/blob/c66e140569df5ef776a1e16812422f63405f2d36/.prow.sh#L3 is problematic because it is both a go import path as well as a directory path and now these 2 diverge with the /v5 bump. I believe that release tooling uses it as a path in https://github.com/kubernetes-csi/external-provisioner/blob/e96448ac2fb9cd15e681bd697c632fc32e580450/release-tools/prow.sh#L1038 and as a go import in https://github.com/kubernetes-csi/external-provisioner/blob/e96448ac2fb9cd15e681bd697c632fc32e580450/release-tools/prow.sh#L977

Any preference on how to fix that?

msau42:
@jakobmoellerdev What do you think of creating a new variable? Edits to release-tools need to be done in https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh? first and than imported here as a git submodule (instructions are in the repo)

jakobmoellerdev:
@msau42 I would need a checkup on the tools PR so that we can unblock that.

jakobmoellerdev:
@msau42 thanks to the adjustment in the release tooling this one is now good to go :+1:

msau42:
/lgtm
/approve

Thanks!

PR #1262, opened 2024-08-20
Bump CSI spec 1.10

jsafrane:
**What type of PR is this?**
/kind cleanup

Bump CSI spec to 1.10.0 and csi-test to 5.3.0.

* gomock is not able to match newly generated Protobuf messages from csi-test, because their private fields differ. Implement a custom message matcher.
* k8s.io/apimachinery can be used to compare API objects, but it panics on the new protobuf messages. Use a generic `cmp.Equal` instead, as suggested in https://protobuf.dev/reference/go/faq/#deepequal

WIP:
* ~the matcher should be probably in csi-tests repo, as it's going to be used by many sidecars.~

```release-note
Updated CSI spec to 1.10.0
```

cc @AndrewSirenko @xing-yang @gnufied


jsafrane:
I filed an issue in the CSI spec repo: https://github.com/container-storage-interface/spec/issues/570
This should not be a minor package bump.

pkg/controller/controller_test.go
```diff
@@ -7046,3 +7048,22 @@ func createFakeCSIPV(volumeHandle string) *v1.PersistentVolume {
 		},
 	}
 }
+
+type ProtobufMatcher struct {
+	// Protobuf messages implement String() function that renders the message as "protobuf style string".
+	// `name:"test-testi"  capacity_range:{required_bytes:100000}  parameters:{key:"fstype"  value:"ext3"}  parameters:{key:"translated"  value:"foo"}`
+	msg proto.Message
+}
+
```
gnufied, line 7057: nit: Since we are defining a new matcher, canonical way of defining this in the context of mock library appears to be defining a function which returns matcher interface. So something like - https://github.com/uber-go/mock/pull/63/files

AndrewSirenko: I can handle this nit when I move the matcher over to csi-test, if we wanted to get this PR in for next provisioner release.

pkg/controller/controller_test.go
```diff
@@ -7046,3 +7048,22 @@ func createFakeCSIPV(volumeHandle string) *v1.PersistentVolume {
 		},
 	}
 }
+
+type ProtobufMatcher struct {
+	// Protobuf messages implement String() function that renders the message as "protobuf style string".
+	// `name:"test-testi"  capacity_range:{required_bytes:100000}  parameters:{key:"fstype"  value:"ext3"}  parameters:{key:"translated"  value:"foo"}`
```
jsafrane, line 7054: this comment is no longer valid

jsafrane: Comment removed

pkg/controller/controller_test.go
jsafrane: I added `Protobuf()` function to return the matcher. Better name is welcome. And indeed, this belongs to csi-test.

AndrewSirenko:
This fixes tests for external-health-monitor and liveness probe when moved to csi-test

- https://github.com/kubernetes-csi/csi-test/pull/553

See tests passing with replaced csi-test module: 
- https://github.com/kubernetes-csi/livenessprobe/pull/296
- https://github.com/kubernetes-csi/external-health-monitor/pull/277

jsafrane:
@AndrewSirenko I updated to csi-test 5.3.1, PTAL

PR #1267, opened 2024-08-27
fix data race on makeCapacity & fakeController

dfajmon:
**What type of PR is this?**
/kind bug

When running tests with `-race` go routine leaks on `capacityCounter` & `utilruntime.ReallyCrash`
Creating mutex for these variables fixes it.

```release-note
NONE
```

### DIFF
--- a/pkg/capacity/capacity_test.go
+++ b/pkg/capacity/capacity_test.go
@@ -26,6 +26,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/capacity/capacity_test.go
```diff
@@ -1608,9 +1612,13 @@ func str2quantity(str string) *resource.Quantity {
 	return &quantity
 }
 
+var counterMutex sync.Mutex
 var capacityCounter int
 
 func makeCapacity(in testCapacity) *storagev1.CSIStorageCapacity {
+	counterMutex.Lock()
+	defer counterMutex.Unlock()
+	// This is a global counter, that goroutines access causing data race
```
gnufied, line 1621: How about something like https://gist.github.com/gnufied/f87cb7c574fc74b7b16797fd2d95ac71 ?

dfajmon: updated, thanks for help

pkg/capacity/capacity_test.go
```diff
@@ -1608,10 +1605,28 @@ func str2quantity(str string) *resource.Quantity {
 	return &quantity
 }
 
-var capacityCounter int
+type capacityCount struct {
+	sync.RWMutex
+	capacityCounter int
+}
```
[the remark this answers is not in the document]
dfajmon: sure, updated

PR #1269, opened 2024-09-06
Fix race in tests

gnufied:
Looks like this was introduced later.


jsafrane:
/test pull-kubernetes-csi-external-provisioner-1-27-on-kubernetes-1-27
/test pull-kubernetes-csi-external-provisioner-1-28-on-kubernetes-1-28
Why those did not start?

Issue #1333, opened 2025-01-20
csi-provisioner: managed-by label is failing with the long-node names.


vamsikrishna-siddu:
**What happened**:
 After the hostpath-provisioner installation the csi-provisioner should create the CSIstoragecapacity objects. But the creation is failing with the following error.
I0120 04:14:26.724089       1 controller.go:873] "Started provisioner controller" component="kubevirt.io.hostpath-provisioner_hostpath-provisioner-csi-smrh4_6264f065-eaed-47aa-90a9-3a2563c965c2"
W0120 04:14:28.020372       1 reflector.go:547] k8s.io/client-go/informers/factory.go:160: failed to list *v1.CSIStorageCapacity: unable to parse requirement: values[0][csi.storage.k8s.io/managed-by]: Invalid value: "-53f40b57-worker1.ocp-virt1-s390x.s390g.lab.eng.rdu2.redhat.com": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
E0120 04:14:28.020406       1 reflector.go:150] k8s.io/client-go/informers/factory.go:160: Failed to watch *v1.CSIStorageCapacity: failed to list *v1.CSIStorageCapacity: unable to parse requirement: values[0][csi.storage.k8s.io/managed-by]: Invalid value: "-53f40b57-worker1.ocp-virt1-s390x.s390g.lab.eng.rdu2.redhat.com": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')

**What you expected to happen**:
Successful creation of CSIStorageCapacity without any error.

**How to reproduce it**:
This issue is coming with the worker nodes have long node names.
worker0.ocp-virt1-s390x.s390g.lab.eng.rdu2.redhat.com
worker1.ocp-virt1-s390x.s390g.lab.eng.rdu2.redhat.com

**Anything else we need to know?**:

**Environment**:
- Driver version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
[...]

PR #1334, opened 2025-01-21
Fix the managed-by-label getting populated with invalid value.

vamsikrishna-siddu:
**What type of PR is this?**
>
> /kind api-change

 /kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

There will be a case when the length of the node name is equals to 53 and base length becomes zero. This will make the prefix value to be empty and the label will be generated starting with '-' which is invalid label.
Ex: nodename is 
worker1.ocp-virt2-s390x.s390g.lab.eng.rdu2.redhat.com
generated label name:
-53f40b57-worker1.ocp-virt1-s390x.s390g.lab.eng.rdu2.redhat.com

This is causing the issue for the hostpath provisioner while generating csistoragecapacity objects.
`W0116 09:21:22.515275       1 reflector.go:547] k8s.io/client-go/informers/factory.go:160: failed to list *v1.CSIStorageCapacity: unable to parse requirement: values[0][csi.storage.k8s.io/managed-by]: Invalid value: "-53f40b57-worker1.ocp-virt1-s390x.s390g.lab.eng.rdu2.redhat.com": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')`

Fixes https://github.com/kubernetes-csi/external-provisioner/issues/1333

```release-note
storage capacity: managed-by label getting invalid label  with a long node name and enableNodeDeployment set to true
```

### DIFF
--- a/cmd/csi-provisioner/util.go
+++ b/cmd/csi-provisioner/util.go
@@ -36,7 +36,7 @@ func getNameWithMaxLength(base, suffix string, maxLength int) string {
 	baseLength := maxLength - 10 /*length of -hash-*/ - len(suffix)
 
 	// if the suffix is too long, ignore it
-	if baseLength < 0 {
+	if baseLength <= 0 {
 		prefix := base[0:min(len(base), max(0, maxLength-9))]
 		// Calculate hash on initial base-suffix string
 		shortName := fmt.Sprintf("%s-%s", prefix, hash(name))
--- a/cmd/csi-provisioner/util_test.go
+++ b/cmd/csi-provisioner/util_test.go
@@ -45,6 +45,10 @@ func TestGetNameWithMaxLength(t *testing.T) {
 			nodeName: fmt.Sprintf("node%s", strings.Repeat("a", 39)),
 		},
 		"very long, ignore suffix": {
+			expected: fmt.Sprintf("%s-%s", externalProvisioner, "a3607ff1"),
+			nodeName: fmt.Sprintf("node%s", strings.Repeat("a", 49)),
+		},
+		"very very long, ignore suffix": {
 			expected: fmt.Sprintf("%s-%s", externalProvisioner, "df38e37f"),
 			nodeName: fmt.Sprintf("node%s", strings.Repeat("a", 63)),
 		},


mowangdk:
Please sign the CLA Authorization first.

vamsikrishna-siddu:
> Please sign the CLA Authorization first.

sure

vamsikrishna-siddu:
Hi @mowangdk I have got the CLA Authorization. Can you please review the pr?

awels:
I will try to take a look at this soon, I have been out sick for a few days and I am finally getting to feeling better.

cmd/csi-provisioner/util.go
```diff
@@ -36,7 +36,7 @@ func getNameWithMaxLength(base, suffix string, maxLength int) string {
 	baseLength := maxLength - 10 /*length of -hash-*/ - len(suffix)
 
 	// if the suffix is too long, ignore it
-	if baseLength < 0 {
+	if baseLength <= 0 {
 		prefix := base[0:min(len(base), max(0, maxLength-9))]
 		// Calculate hash on initial base-suffix string
```
mowangdk, line 41: I think we need to ensure prefix must not equals to ""

mowangdk: And maybe we also need add more ut here to avoid corner cases

vamsikrishna-siddu: Hi @mowangdk ,
with the condition
`if baseLength <= 0`
the prefix will never by empty.
and I have added a unit test for the above issue. With this it will cover all the corner cases.

mowangdk: @vamsikrishna-siddu Did you mean the prefix outside the if condition? The prefix inside the if condition in getNameWithMaxLength, is not guaranteed i think?

cmd/csi-provisioner/util_test.go
```diff
@@ -45,6 +45,10 @@ func TestGetNameWithMaxLength(t *testing.T) {
 			nodeName: fmt.Sprintf("node%s", strings.Repeat("a", 39)),
 		},
 		"very long, ignore suffix": {
+			expected: fmt.Sprintf("%s-%s", externalProvisioner, "a3607ff1"),
+			nodeName: fmt.Sprintf("node%s", strings.Repeat("a", 49)),
```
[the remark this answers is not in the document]
vamsikrishna-siddu: sure

cmd/csi-provisioner/util.go
vamsikrishna-siddu: @mowangdk yeah i meant the prefix outside the if condition. But the prefix inside the if condition is also never become empty. The reason is the base is  always `external-provisioner` and the maxLength is always 63. so the prefix will never be empty. The only case where the prefix will be empty is when we pass the base as empty string. But in code, I dont see any place where we are using this function for any other use case which provides the base as empty string.

cmd/csi-provisioner/util_test.go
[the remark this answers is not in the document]
vamsikrishna-siddu: done

vamsikrishna-siddu:
Hi @mowangdk @jsafrane , I have addressed your review comments. Can you please review it once.

cmd/csi-provisioner/util.go
mowangdk: yeah，I agreed.  But IMHO, it's better not to rely on function parameters because they can change anytime. Instead, parameter validation should be handled within the function itself.

vamsikrishna-siddu: @mowangdk  okay got it. I have done the required changes. Can you please review it once again.

PR #1338, opened 2025-01-29
Set ownerRef of CSIStorageCapacity to node when using node deployment

### LINKED ISSUE #1337: Should CSIStorageCapacity ownerReference use node when node-deployment=true
Today the logic that sets the `ownerReference` for CSIStorageCapacity is here:
https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L462-L479

Where it uses the configured `capacity-ownerref-level` and traverses from the pod (2 for Deployment and 1 for StatefulSet/DaemonSet)

However, when using `node-deployment=true` for deploying the external-provisioner as a DaemonSet for handling node-local volumes, what ends up happening is that the `CSIStorageCapacity` would end up having the DaemonSet as its `ownerReference`

for e.g.
```
apiVersion: storage.k8s.io/v1
kind: CSIStorageCapacity
capacity: 20437544Ki
maximumVolumeSize: 10218772Ki
metadata:
  labels:
    csi.storage.k8s.io/drivername: foo
    csi.storage.k8s.io/managed-by: external-provisioner-node1
  name: csisc-cpzrc
  namespace: kube-system
  ownerReferences:
  - apiVersion: apps/v1
    controller: true
    kind: DaemonSet
    name: foo-csi-controller
    uid: d2c8a05e-6395-42a3-8881-6f407210cfdd
nodeTopology:
  matchLabels:
    kubernetes.io/hostname: node1
storageClassName: foo
```

And as a result if the node gets deleted or consolidated by cluster-autoscaler for e.g. this `CSIStorageCapacity` resource becomes orphaned as it won't be garbage-collected (since the DaemonSet still exists) but will never be used as the node is gone and when cluster-autoscaler brings up a new node, then it likely won't have the same name.

So in the case of using `node-deployment=true` with `enable-capacity=true`, should the CSIStorageCapacity's ownerReference use node resource instead? This way, it will be properly garbage collected when the node no longer exists.

yibozhuang:
**What type of PR is this?**
/kind feature

This change adds logic to set the OwnerReference of the CSIStorageCapacity to the node itself when using node deployment as in this case the external-provisioner will be deployed as a DaemonSet together with the CSI plugin for provisioning node local volumes. Under this case, we should set the OwnerReference of the corresponding CSIStorageCapacity to the node itself so that when the node is gone (e.g. removed by cluster-autoscaler), the CSIStorageCapacity will be garbage-collected and not left behind.

Fixes #1337 

Yes

```release-note
set ownerReference of CSIStorageCapacity to node object when using node-based deployment for provisioning node-local volumes
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -461,19 +461,33 @@ func main() {
 		}
 		var controller *metav1.OwnerReference
 		if *capacityOwnerrefLevel >= 0 {
-			podName := os.Getenv("POD_NAME")
-			if podName == "" {
-				klog.Fatal("need POD_NAME env variable to determine CSIStorageCapacity owner")
-			}
-			var err error
-			controller, err = owner.Lookup(config, namespace, podName,
-				schema.GroupVersionKind{
-					Group:   "",
-					Version: "v1",
-					Kind:    "Pod",
-				}, *capacityOwnerrefLevel)
-			if err != nil {
-				klog.Fatalf("look up owner(s) of pod: %v", err)
+			if *enableNodeDeployment {
+				// When using node deployment, set the OwnerReference of the CSIStorageCapacity to
+				// the node so that it will be garbage collected when the node is deleted.
+				controller, err = owner.Lookup(config, "", node,
+					schema.GroupVersionKind{
+						Group:   "",
+						Version: "v1",
+						Kind:    "Node",
+					}, 0)
+				if err != nil {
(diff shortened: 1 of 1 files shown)


pohly:
Let's discuss the concept first in https://github.com/kubernetes-csi/external-provisioner/issues/1337#issuecomment-2620822092

/hold

yibozhuang:
Closing as we can use `--capacity-ownerref-level=0` to use Pod as the owner of `CSIStorageCapacity` and there is already code that handles this.

PR #1339, opened 2025-01-30
Add functionality to set pv gid annotation via driver volumecontext

dankova22:
**What type of PR is this?**
/kind feature

We would like to add functionality to a CSI driver to be able set stricter permissions on provisioned directories, and still have this be accessible to whatever pod uses this provisioned dir. To do this, we would like to be able to dynamically assign a random gid to a provisioned dir, and add this gid as a supplemental group to any pod using this volume. We need to set the `pv.beta.kubernetes.io/gid` annotation in the PV metadata:  https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#access-control, however this is not exposed in the CreateVolume response of CSI driver.

Therefore, we are requesting this feature of the external-provisioner that will parse `VolumeContext` for a gid key and apply this to the PV metadata if returned from CSI driver. 
notes for your reviewer**:

```release-note
Set PV gid metadata annotation in provision workflow if gid key is returned in volume context from CSI driver
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -54,6 +54,7 @@ import (
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/sig-storage-lib-external-provisioner/v11/controller"
 	"sigs.k8s.io/sig-storage-lib-external-provisioner/v11/util"
+	"sigs.k8s.io/sig-storage-lib-external-provisioner/v11/gidallocator"
 
 	"github.com/kubernetes-csi/csi-lib-utils/connection"
 	"github.com/kubernetes-csi/csi-lib-utils/metrics"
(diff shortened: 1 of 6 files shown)


pkg/controller/controller.go
```diff
@@ -933,6 +937,11 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, annDeletionProvisionerSecretRefNamespace, "")
 	}
 
+	if gid, ok := volumeAttributes[gidCsiKey]; ok {
```
AndrewSirenko, line 940: I think having provisioner reach into volumeAttributes like this violates CSI Spec. 

volumeAttributes comes from CSI CreateVolumeResponse.volume.volume_context, which according to CSI Spec shall be opaque to CO (and therefore csi-provisioner) 

```
// Opaque static properties of the volume. SP MAY use this field to
  // ensure subsequent volume validation and publishing calls have
  // contextual information.
  // The contents of this field SHALL be opaque to a CO.
```

[source](https://github.com/container-storage-interface/spec/blob/98819c45a37a67e0cd466bd02b813faf91af4e45/spec.md?plain=1#L1035-L1050)

PR #1411, opened 2025-09-05
Update kubernetes dependencies to v1.34.0

dfajmon:
**What type of PR is this?**

/kind cleanup

Update the release-tools subtree and bump all k8s dependencies to v1.34.0

cc @jsafrane 

```release-note
Update kubernetes dependencies to v1.34.0
```


go.mod
```diff
@@ -135,82 +113,78 @@ require (
...
 	sigs.k8s.io/structured-merge-diff/v4 v4.7.0 // indirect
-	sigs.k8s.io/yaml v1.4.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
+	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
 replace k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.32.0
 
-replace k8s.io/apimachinery => k8s.io/apimachinery v0.33.0
+replace k8s.io/apimachinery => k8s.io/apimachinery v0.34.0
```
sunnylovestiramisu: Can we remove the replace?

Because of Golang's Minimal Version Selection (MVS), Go will automatically pick the highest version required by any dependency, which is what our replace statement does anyway. So it is safe to remove.

Example:
if module A requires client-go v0.32.0 and module B requires client-go v0.34.0, Go satisfies both by selecting v0.34.0. It assumes that v0.34.0 is backwards-compatible with v0.32.0, which is generally true for minor version changes according to semantic versioning.

dfajmon: Yes, I can remove the replaces. I kept them because they were already there. Otherwise I would do the same.

dfajmon:
resolved in https://github.com/kubernetes-csi/external-provisioner/pull/1414

PR #1413, opened 2025-09-08
Add In Memory Cache of Node/CSINode Info for Long Provisioning to Prevent Volume Leak

sunnylovestiramisu:
**What type of PR is this?**
/kind bug

This is the option 4 discussed in [Leaked Volume Fix in external-provisioner Google doc](https://docs.google.com/document/d/11u_Is5-p9MS2lQWnarEWhEyXLR7a3g5B8rdtNcG52Dk/edit?resourcekey=0-SKG1mIKNzp-aQniZvr0EDA&tab=t.0#heading=h.6kapwl4t599f). 

The short term fix does not solve the issue of csi-provisioner restarting during the long provisioning time, however it is still an improvement. Also the short term fix does not have conflict with other long term fixes if we want to revisit in the future.

https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/issues/152

Making changes in vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v11 to get initial review, I will move the change to sig-storage-lib-external-provisioner and bump up the sig-storage-lib-external-provisioner **major** version once this review is in good shape.

##### Steps
Tested via the following steps.
1. ``kubetest --build --up`` from a k/k repo
2. Add wait time to PDCSI driver similar to [this](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/compare/master...sunnylovestiramisu:gcp-compute-persistent-disk-csi-driver:leakedVolumeNonMasterNode) branch
3. Make the PDCSI driver consuming the external-provisioner change in this PR and deploy PDCSI driver
4. Find the node of the csi-gce-pd-controller, for example
```
k get pod -n gce-pd-csi-driver 
NAME                                    READY   STATUS             RESTARTS        AGE
csi-gce-pd-controller-8dffb78ff-2xzbq   5/5     Running            0               27m
csi-gce-pd-node-54b8l                   2/2     Running            0               27m
csi-gce-pd-node-5fcwm                   2/2     Running            0               27m
csi-gce-pd-node-7g72v                   0/2     CrashLoopBackOff   15 (100s ago)   27m
```
csi-gce-pd-controller-8dffb78ff-2xzbq is on node e2e-test-songsunny-minion-group-kncr, for the following testing pod, specify a nodeSelector that is **NOT** on node e2e-test-songsunny-minion-group-kncr
5. Apply the following objects:
```
kubectl apply -f - <<EOF
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: pd-storage
provisioner: pd.csi.storage.gke.io
volumeBindingMode: WaitForFirstConsumer
allowVolumeExpansion: true
parameters:
  type: pd-balanced
EOF

kubectl apply -f - <<EOF
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pd-pvc
spec:
  storageClassName: pd-storage
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 67Gi
EOF

kubectl apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  containers:
  - name: nginx
    image: nginx
    volumeMounts:
    - name: pd-volume
      mountPath: /data/pd
  nodeSelector:
    kubernetes.io/hostname: e2e-test-songsunny-minion-group-302x
  volumes:
  - name: pd-volume
    persistentVolumeClaim:
      claimName: pd-pvc
      readOnly: false
EOF
```
6. Delete the node e2e-test-songsunny-minion-group-302x, and then the Pod,  and then the PVC
##### Testing Results
1. Deleted the node only, provisioning continued with:
```
successfully created PV xxxxxxx for PVC xxxx and csi volume name xxxxxxxx
```
in the logs. The PVC object is bound to the PV. Deleting PVC is cleaning up the PV object and the actual disk correctly.
3. Deleted node, pod and PVC, provisioning continued with:
```
successfully created PV xxxxxxx for PVC xxxx and csi volume name xxxxxxxx
```

And then a delete volume is called:
```
I0910 22:22:49.848316       1 controller.go:1302] "shouldDelete" PV="pvc-11ed9689-1fcf-4c1d-9ad0-0f5e86969f10"
I0910 22:22:49.848331       1 controller.go:1332] "shouldDelete is true" PV="pvc-11ed9689-1fcf-4c1d-9ad0-0f5e86969f10"
I0910 22:22:49.848338       1 controller.go:1176] "shouldDelete" PV="pvc-11ed9689-1fcf-4c1d-9ad0-0f5e86969f10"
I0910 22:22:49.848432       1 controller.go:1644] "Started" PV="pvc-11ed9689-1fcf-4c1d-9ad0-0f5e86969f10"
I0910 22:22:49.848534       1 controller.go:1328] volume pvc-11ed9689-1fcf-4c1d-9ad0-0f5e86969f10 does not need any deletion secrets
I0910 22:22:49.848634       1 connection.go:264] "GRPC call" method="/csi.v1.Controller/DeleteVolume" request="{\"volume_id\":\"projects/XXXXXX/zones/us-central1-b/disks/pvc-11ed9689-1fcf-4c1d-9ad0-0f5e86969f10\"}"
I0910 22:22:50.354062       1 connection.go:270] "GRPC response" response="{}" err=null
I0910 22:22:50.354105       1 controller.go:1659] "Volume deleted" PV="pvc-11ed9689-1fcf-4c1d-9ad0-0f5e86969f10"
I0910 22:22:50.378980       1 controller.go:1702] "PersistentVolume deleted succeeded" PV="pvc-11ed9689-1fcf-4c1d-9ad0-0f5e86969f10"
```

##### Steps
1. Create a GKE cluster and disable managed PDCSI 
2. Deploy a self managed PDCSI driver with a long wait in CreateVolume as in [this branch](https://github.com/sunnylovestiramisu/gcp-compute-persistent-disk-csi-driver/tree/leakedVolumeTestingGKE)
3. Apply the following objects:
```
kubectl apply -f - <<EOF
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: pd-storage
provisioner: pd.csi.storage.gke.io
volumeBindingMode: Immediate
reclaimPolicy: Delete
allowVolumeExpansion: true
parameters:
  type: pd-balanced
EOF

kubectl apply -f - <<EOF
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pd-pvc
spec:
  storageClassName: pd-storage
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 67Gi
EOF
```
5. Scale down the nodes to 0 in the zone of the disk
6. In the logs verify the PV object is created:
```
successfully created PV xxxxxxx for PVC xxxx and csi volume name xxxxxxxx
```

```release-note
Add In Memory Cache of Node/CSINode Info for Long Provisioning to Prevent Volume Leak
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -408,7 +408,6 @@ func main() {
 		controller.Threadiness(int(*workerThreads)),
 		controller.CreateProvisionedPVLimiter(workqueue.DefaultTypedControllerRateLimiter[string]()),
 		controller.ClaimsInformer(claimInformer),
-		controller.NodesLister(nodeLister),
 		controller.RetryIntervalMax(*retryIntervalMax),
 	}
 
@@ -418,6 +417,10 @@ func main() {
 
 	if supportsMigrationFromInTreePluginName != "" {
 		provisionerOptions = append(provisionerOptions, controller.AdditionalProvisionerNames([]string{supportsMigrationFromInTreePluginName}))
+	}
+	var pvcNodeStore *ctrl.InMemoryStore
+	if ctrl.SupportsTopology(pluginCapabilities) {
+		pvcNodeStore = ctrl.NewInMemoryStore()
 	}
 
 	// Create the provisioner: it implements the Provisioner interface expected by
@@ -448,6 +451,7 @@ func main() {
 		nodeDeployment,
 		*controllerPublishReadOnly,
 		*preventVolumeModeConversion,
+		pvcNodeStore,
 	)
 
 	var capacityController *capacity.Controller
--- a/pkg/controller/cache.go
+++ b/pkg/controller/cache.go
@@ -1,0 +1,132 @@
+package controller
+
+import (
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// TopologyInfo holds the data for NodeLabels and TopologyKeys
+type TopologyInfo struct {
+	NodeLabels     map[string]string
+	TopologyKeys   []string
+	RequisiteTerms []topologyTerm
+}
+
+// TopologyProvider is an interface that defines the behavior for looking up
+// a TopologyInfo object by its pvc UID.
+type TopologyProvider interface {
+	GetByPvcUID(pvcUID types.UID) (*TopologyInfo, error)
+	// The entry is deleted when provision succeeds or returns a final error.
+	Delete(pvcUID types.UID) error
+
+	// Update methods now perform an "upsert" and don't return errors.
+	UpdateNodeLabels(pvcUID types.UID, newLabels map[string]string)
+	UpdateTopologyKeys(pvcUID types.UID, newKeys []string)
+	UpdateRequisiteTerms(pvcUID types.UID, requisiteTerms []topologyTerm)
+}
+
+// InMemoryStore is a concrete implementation of TopologyProvider.
+// It uses an in-memory map for quick lookups.
+type InMemoryStore struct {
+	// The map key is the object's name.
+	data map[types.UID]*TopologyInfo
+	// Adding a mutex for thread-safe access
+	mutex sync.RWMutex
+}
+
+// NewInMemoryStore creates and initializes a new store.
+func NewInMemoryStore() *InMemoryStore {
+	return &InMemoryStore{
+		data: make(map[types.UID]*TopologyInfo),
+	}
+}
+
+// Delete implements the TopologyProvider interface.
+// It uses the built-in delete() function to remove the item from the map.
+func (s *InMemoryStore) Delete(pvcUID types.UID) error {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	// First, check if the key exists to provide a helpful error.
+	_, found := s.data[pvcUID]
+	if !found {
+		return nil
+	}
+	delete(s.data, pvcUID)
+	return nil
+}
+
+// GetByPvcUID implements the TopologyProvider interface.
+func (s *InMemoryStore) GetByPvcUID(pvcUID types.UID) (*TopologyInfo, error) {
+	if s == nil {
+		return nil, fmt.Errorf("pvcNodeStore is nil")
+	}
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+	info, found := s.data[pvcUID]
+	if !found {
+		return nil, fmt.Errorf("topology object with pvcUID '%s' not found", pvcUID)
+	}
+
+	// Return a deep copy to prevent data races
+	infoCopy := &TopologyInfo{}
+	if info.NodeLabels != nil {
+		infoCopy.NodeLabels = make(map[string]string)
+		for k, v := range info.NodeLabels {
+			infoCopy.NodeLabels[k] = v
+		}
+	}
+
+	if info.TopologyKeys != nil {
+		infoCopy.TopologyKeys = make([]string, len(info.TopologyKeys))
+		copy(infoCopy.TopologyKeys, info.TopologyKeys)
+	}
+
+	if info.RequisiteTerms != nil {
+		infoCopy.RequisiteTerms = make([]topologyTerm, len(info.RequisiteTerms))
+		for i, term := range info.RequisiteTerms {
+			newTerm := make(topologyTerm, len(term))
+			copy(newTerm, term)
+			infoCopy.RequisiteTerms[i] = newTerm
+		}
+	}
+
+	return infoCopy, nil
+}
+
+// UpdateNodeLabels finds an object by pvcUID and replaces its NodeLabels.
+func (s *InMemoryStore) UpdateNodeLabels(pvcUID types.UID, newLabels map[string]string) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	info, found := s.data[pvcUID]
+	if !found {
+		s.data[pvcUID] = &TopologyInfo{NodeLabels: newLabels}
+	} else {
+		info.NodeLabels = newLabels
+	}
+}
+
+// UpdateTopologyKeys finds an object by pvcUID and replaces its TopologyKeys.
+func (s *InMemoryStore) UpdateTopologyKeys(pvcUID types.UID, newKeys []string) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	info, found := s.data[pvcUID]
+	if !found {
+		s.data[pvcUID] = &TopologyInfo{TopologyKeys: newKeys}
+	} else {
+		info.TopologyKeys = newKeys
+	}
+}
+
+// UpdateRequisiteTerms finds an object by pvcUID and replaces its RequisiteTerms.
+func (s *InMemoryStore) UpdateRequisiteTerms(pvcUID types.UID, requisiteTerms []topologyTerm) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	info, found := s.data[pvcUID]
+	if !found {
+		s.data[pvcUID] = &TopologyInfo{RequisiteTerms: requisiteTerms}
+	} else {
+		info.RequisiteTerms = requisiteTerms
+	}
+}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -280,6 +280,7 @@ type csiProvisioner struct {
 	nodeDeployment                        *internalNodeDeployment
 	controllerPublishReadOnly             bool
 	preventVolumeModeConversion           bool
+	pvcNodeStore                          TopologyProvider
 }
 
 var (
@@ -361,6 +362,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 	nodeDeployment *NodeDeployment,
 	controllerPublishReadOnly bool,
 	preventVolumeModeConversion bool,
+	pvcNodeStore TopologyProvider,
 ) controller.Provisioner {
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartLogging(klog.Infof)
@@ -396,6 +398,7 @@ func NewCSIProvisioner(client kubernetes.Interface,
 		eventRecorder:                         eventRecorder,
 		controllerPublishReadOnly:             controllerPublishReadOnly,
 		preventVolumeModeConversion:           preventVolumeModeConversion,
+		pvcNodeStore:                          pvcNodeStore,
 	}
 	if nodeDeployment != nil {
 		provisioner.nodeDeployment = &internalNodeDeployment{
@@ -551,7 +554,7 @@ type prepareProvisionResult struct {
 }
 
 // prepareProvision does non-destructive parameter checking and preparations for provisioning a volume.
-func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.PersistentVolumeClaim, sc *storagev1.StorageClass, selectedNode *v1.Node) (*prepareProvisionResult, controller.ProvisioningState, error) {
+func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.PersistentVolumeClaim, sc *storagev1.StorageClass, selectedNodeName string) (*prepareProvisionResult, controller.ProvisioningState, error) {
 	if sc == nil {
 		return nil, controller.ProvisioningFinished, errors.New("storage class was nil")
 	}
@@ -689,13 +692,15 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		requirements, err := GenerateAccessibilityRequirements(
 			p.client,
 			p.driverName,
+			claim.UID,
 			claim.Name,
 			sc.AllowedTopologies,
-			selectedNode,
+			selectedNodeName,
 			p.strictTopology,
 			p.immediateTopology,
 			p.csiNodeLister,
-			p.nodeLister)
+			p.nodeLister,
+			p.pvcNodeStore)
 		if err != nil {
 			return nil, controller.ProvisioningNoChange, fmt.Errorf("error generating accessibility requirements: %v", err)
 		}
@@ -815,7 +820,7 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 		}
 	}
 
-	result, state, err := p.prepareProvision(ctx, claim, options.StorageClass, options.SelectedNode)
+	result, state, err := p.prepareProvision(ctx, claim, options.StorageClass, options.SelectedNodeName)
 	if result == nil {
 		return nil, state, err
 	}
@@ -844,14 +849,18 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 		// even drivers which did not ask for it explicitly might still only look at the first
 		// topology entry and thus succeed after rescheduling.
 		mayReschedule := p.supportsTopology() &&
-			options.SelectedNode != nil
+			len(options.SelectedNodeName) > 0
 		state := checkError(err, mayReschedule)
 		klog.V(5).Infof("CreateVolume failed, supports topology = %v, node selected %v => may reschedule = %v => state = %v: %v",
 			p.supportsTopology(),
-			options.SelectedNode != nil,
+			options.SelectedNodeName,
 			mayReschedule,
 			state,
 			err)
+		// Delete the entry in in memory cache if the error is final
+		if p.supportsTopology() && (state == controller.ProvisioningFinished || state == controller.ProvisioningReschedule) {
+			p.pvcNodeStore.Delete(claim.UID)
+		}
 		return nil, state, err
 	}
 
@@ -973,6 +982,10 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 	}
 
 	klog.V(5).Infof("successfully created PV %+v", pv.Spec.PersistentVolumeSource)
+	// Remove entry from the in memory cache
+	if p.supportsTopology() {
+		p.pvcNodeStore.Delete(claim.UID)
+	}
 	return pv, controller.ProvisioningFinished, nil
 }
 
@@ -1481,7 +1494,7 @@ func (p *csiProvisioner) checkNode(ctx context.Context, claim *v1.PersistentVolu
 		// it must match our own. We can find out by trying to
 		// create accessibility requirements.  If that fails,
 		// we should not become the owner.
-		if len(sc.AllowedTopologies) > 0 {
+		if len(sc.AllowedTopologies) > 0 && p.supportsTopology() {
 			node, err := p.nodeLister.Get(p.nodeDeployment.NodeName)
 			if err != nil {
 				return false, err
@@ -1489,13 +1502,15 @@ func (p *csiProvisioner) checkNode(ctx context.Context, claim *v1.PersistentVolu
 			if _, err := GenerateAccessibilityRequirements(
 				p.client,
 				p.driverName,
+				claim.UID,
 				claim.Name,
 				sc.AllowedTopologies,
-				node,
+				node.Name,
 				p.strictTopology,
 				p.immediateTopology,
 				p.csiNodeLister,
-				p.nodeLister); err != nil {
+				p.nodeLister,
+				p.pvcNodeStore); err != nil {
 				if logger.Enabled() {
 					logger.Infof("%s: ignoring PVC %s/%s, allowed topologies is not compatible: %v", caller, claim.Namespace, claim.Name, err)
 				}
@@ -1550,7 +1565,7 @@ func (p *csiProvisioner) checkCapacity(ctx context.Context, claim *v1.Persistent
 		return false, err
 	}
 
-	result, _, err := p.prepareProvision(ctx, claim, sc, node)
+	result, _, err := p.prepareProvision(ctx, claim, sc, node.Name)
 	if err != nil {
 		return false, err
 	}
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -428,8 +428,12 @@ func TestCreateDriverReturnsInvalidCapacityDuringProvision(t *testing.T) {
 	clientSet := fakeclientset.NewSimpleClientset(clientSetObjects...)
 
 	pluginCaps, controllerCaps := provisionCapabilities()
+	var pvcNodeStore *InMemoryStore = nil
+	if SupportsTopology(pluginCaps) {
+		pvcNodeStore = NewInMemoryStore()
+	}
 	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test",
-		5, csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, nil, false, defaultfsType, nil, true, false)
+		5, csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, nil, false, defaultfsType, nil, true, false, pvcNodeStore)
 
 	// Requested PVC with requestedBytes storage
 	deletePolicy := v1.PersistentVolumeReclaimDelete
@@ -2091,7 +2095,7 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 		"distributed, right node selected": {
 			deploymentNode: "foo",
 			volOpts: controller.ProvisionOptions{
-				SelectedNode: nodeFoo,
+				SelectedNodeName: nodeFoo.Name,
 				StorageClass: &storagev1.StorageClass{
 					ObjectMeta: metav1.ObjectMeta{
 						Name: fakeSCName,
@@ -2147,7 +2151,7 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 		"distributed, wrong node selected": {
 			deploymentNode: "foo",
 			volOpts: controller.ProvisionOptions{
-				SelectedNode: nodeBar,
+				SelectedNodeName: nodeBar.Name,
 				StorageClass: &storagev1.StorageClass{
 					ObjectMeta: metav1.ObjectMeta{
 						Name: fakeSCName,
@@ -2172,7 +2176,7 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 			deploymentNode:   "foo",
 			immediateBinding: true,
 			volOpts: controller.ProvisionOptions{
-				SelectedNode: nodeFoo,
+				SelectedNodeName: nodeFoo.Name,
 				StorageClass: &storagev1.StorageClass{
 					ObjectMeta: metav1.ObjectMeta{
 						Name: fakeSCName,
@@ -2290,8 +2294,9 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 			// the owner (= leaves the selected node annotation unset) because
 			// it would not be able to provision the volume if it was
 			// the owner (generating accessibility requirements would fail).
-			deploymentNode:   "foo",
-			immediateBinding: true,
+			deploymentNode:     "foo",
+			immediateBinding:   true,
+			pluginCapabilities: provisionWithTopologyCapabilities,
 			volOpts: controller.ProvisionOptions{
 				StorageClass: &storagev1.StorageClass{
 					ObjectMeta: metav1.ObjectMeta{
@@ -2562,12 +2567,16 @@ func runFSTypeProvisionTest(t *testing.T, k string, tc provisioningFSTypeTestcas
 	clientSet := fakeclientset.NewSimpleClientset(tc.clientSetObjects...)
 
 	pluginCaps, controllerCaps := provisionCapabilities()
+	var pvcNodeStore *InMemoryStore = nil
+	if SupportsTopology(pluginCaps) {
+		pvcNodeStore = NewInMemoryStore()
+	}
 
 	if tc.skipDefaultFSType {
 		myDefaultfsType = ""
 	}
 	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5, csiConn.conn,
-		nil, provisionDriverName, pluginCaps, controllerCaps, supportsMigrationFromInTreePluginName, false, true, csitrans.New(), nil, nil, nil, nil, nil, nil, false, myDefaultfsType, nil, false, false)
+		nil, provisionDriverName, pluginCaps, controllerCaps, supportsMigrationFromInTreePluginName, false, true, csitrans.New(), nil, nil, nil, nil, nil, nil, false, myDefaultfsType, nil, false, false, pvcNodeStore)
 	out := &csi.CreateVolumeResponse{
 		Volume: &csi.Volume{
 			CapacityBytes: requestedBytes,
@@ -2690,6 +2699,7 @@ func runProvisionTest(t *testing.T, tc provisioningTestcase, requestedBytes int6
 	objects := tc.clientSetObjects
 	var node *v1.Node
 	var csiNode *storagev1.CSINode
+	var pvc *v1.PersistentVolumeClaim
 	if tc.deploymentNode != "" {
 		node = &v1.Node{
 			ObjectMeta: metav1.ObjectMeta{
@@ -2713,7 +2723,8 @@ func runProvisionTest(t *testing.T, tc provisioningTestcase, requestedBytes int6
 				},
 			},
 		}
-		objects = append(objects, node, csiNode)
+		pvc = &v1.PersistentVolumeClaim{}
+		objects = append(objects, node, csiNode, pvc)
 	}
 	if tc.volOpts.PVC != nil {
 		tc.volOpts.PVC = tc.volOpts.PVC.DeepCopy()
@@ -2749,9 +2760,13 @@ func runProvisionTest(t *testing.T, tc provisioningTestcase, requestedBytes int6
 	} else {
 		pluginCaps, controllerCaps = provisionCapabilities()
 	}
+	var pvcNodeStore *InMemoryStore = nil
+	if SupportsTopology(pluginCaps) {
+		pvcNodeStore = NewInMemoryStore()
+	}
 	mycontrollerPublishReadOnly := tc.controllerPublishReadOnly
 	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5, csiConn.conn,
-		nil, provisionDriverName, pluginCaps, controllerCaps, supportsMigrationFromInTreePluginName, false, true, csitrans.New(), scInformer.Lister(), csiNodeInformer.Lister(), nodeInformer.Lister(), nil, nil, nil, tc.withExtraMetadata, defaultfsType, nodeDeployment, mycontrollerPublishReadOnly, false)
+		nil, provisionDriverName, pluginCaps, controllerCaps, supportsMigrationFromInTreePluginName, false, true, csitrans.New(), scInformer.Lister(), csiNodeInformer.Lister(), nodeInformer.Lister(), claimInformer.Lister(), nil, nil, tc.withExtraMetadata, defaultfsType, nodeDeployment, mycontrollerPublishReadOnly, false, pvcNodeStore)
 
 	// Adding objects to the informer ensures that they are consistent with
 	// the fake storage without having to start the informers.
@@ -4538,8 +4553,12 @@ func TestProvisionFromSnapshot(t *testing.T) {
 		utilfeaturetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.CrossNamespaceVolumeDataSource, tc.xnsEnabled)
 
 		pluginCaps, controllerCaps := provisionFromSnapshotCapabilities()
+		var pvcNodeStore *InMemoryStore = nil
+		if SupportsTopology(pluginCaps) {
+			pvcNodeStore = NewInMemoryStore()
+		}
 		csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5, csiConn.conn,
-			client, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, refGrantLister, false, defaultfsType, nil, true, true)
+			client, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, refGrantLister, false, defaultfsType, nil, true, true, pvcNodeStore)
 
 		out := &csi.CreateVolumeResponse{
 			Volume: &csi.Volume{
@@ -4716,10 +4735,14 @@ func TestProvisionWithTopologyEnabled(t *testing.T) {
 			clientSet := fakeclientset.NewSimpleClientset(nodes, csiNodes)
 
 			scLister, csiNodeLister, nodeLister, claimLister, vaLister, stopChan := listers(clientSet)
+			var pvcNodeStore *InMemoryStore = nil
+			if SupportsTopology(pluginCaps) {
+				pvcNodeStore = NewInMemoryStore()
+			}
 			defer close(stopChan)
 
 			csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5,
-				csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), scLister, csiNodeLister, nodeLister, claimLister, vaLister, nil, false, defaultfsType, nil, true, false)
+				csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), scLister, csiNodeLister, nodeLister, claimLister, vaLister, nil, false, defaultfsType, nil, true, false, pvcNodeStore)
 
 			pv, _, err := csiProvisioner.Provision(context.Background(), controller.ProvisionOptions{
 				StorageClass: &storagev1.StorageClass{},
@@ -4810,17 +4833,21 @@ func TestProvisionErrorHandling(t *testing.T) {
 					csiNodes := buildCSINodes(topologyKeys)
 					clientSet := fakeclientset.NewSimpleClientset(nodes, csiNodes)
 					scLister, csiNodeLister, nodeLister, claimLister, vaLister, stopChan := listers(clientSet)
+					var pvcNodeStore *InMemoryStore = nil
+					if SupportsTopology(pluginCaps) {
+						pvcNodeStore = NewInMemoryStore()
+					}
 					defer close(stopChan)
 
 					csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5,
-						csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), scLister, csiNodeLister, nodeLister, claimLister, vaLister, nil, false, defaultfsType, nil, true, false)
+						csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), scLister, csiNodeLister, nodeLister, claimLister, vaLister, nil, false, defaultfsType, nil, true, false, pvcNodeStore)
 
 					options := controller.ProvisionOptions{
 						StorageClass: &storagev1.StorageClass{},
 						PVC:          createFakePVC(requestBytes),
 					}
 					if nodeSelected {
-						options.SelectedNode = &nodes.Items[0]
+						options.SelectedNodeName = nodes.Items[0].Name
 					}
 					pv, state, err := csiProvisioner.Provision(context.Background(), options)
 
@@ -4885,8 +4912,12 @@ func TestProvisionWithTopologyDisabled(t *testing.T) {
 
 	clientSet := fakeclientset.NewSimpleClientset()
 	pluginCaps, controllerCaps := provisionWithTopologyCapabilities()
+	var pvcNodeStore *InMemoryStore = nil
+	if SupportsTopology(pluginCaps) {
+		pvcNodeStore = NewInMemoryStore()
+	}
 	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5,
-		csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, nil, false, defaultfsType, nil, true, false)
+		csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), nil, nil, nil, nil, nil, nil, false, defaultfsType, nil, true, false, pvcNodeStore)
 
 	out := &csi.CreateVolumeResponse{
 		Volume: &csi.Volume{
@@ -4899,13 +4930,9 @@ func TestProvisionWithTopologyDisabled(t *testing.T) {
 	controllerServer.EXPECT().CreateVolume(gomock.Any(), gomock.Any()).Return(out, nil).Times(1)
 
 	pv, _, err := csiProvisioner.Provision(context.Background(), controller.ProvisionOptions{
-		StorageClass: &storagev1.StorageClass{},
-		PVC:          createFakePVC(requestBytes),
-		SelectedNode: &v1.Node{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "some-node",
-			},
-		},
+		StorageClass:     &storagev1.StorageClass{},
+		PVC:              createFakePVC(requestBytes),
+		SelectedNodeName: "some-node",
 	})
 	if err != nil {
 		t.Fatalf("got error from Provision call: %v", err)
@@ -5580,9 +5607,13 @@ func runDeleteTest(t *testing.T, k string, tc deleteTestcase) {
 	}
 
 	pluginCaps, controllerCaps := provisionCapabilities()
+	var pvcNodeStore *InMemoryStore = nil
+	if SupportsTopology(pluginCaps) {
+		pvcNodeStore = NewInMemoryStore()
+	}
 	scLister, _, _, _, vaLister, _ := listers(clientSet)
 	csiProvisioner := NewCSIProvisioner(clientSet, 5*time.Second, "test-provisioner", "test", 5,
-		csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), scLister, nil, nil, nil, vaLister, nil, false, defaultfsType, nodeDeployment, true, false)
+		csiConn.conn, nil, driverName, pluginCaps, controllerCaps, "", false, true, csitrans.New(), scLister, nil, nil, nil, vaLister, nil, false, defaultfsType, nodeDeployment, true, false, pvcNodeStore)
 
 	err = csiProvisioner.Delete(context.Background(), tc.persistentVolume)
 	if tc.expectErr && err == nil {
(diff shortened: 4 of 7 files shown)


vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v11/controller/controller.go
```diff
@@ -1486,31 +1486,19 @@ func (ctrl *ProvisionController) provisionClaimOperation(ctx context.Context, cl
 		return ProvisioningFinished, errStopProvision
 	}
 
-	var selectedNode *v1.Node
+	var selectedNodeName string
```
sunnylovestiramisu, line 1468: @jsafrane Should we just make a new major version instead of making it an option for other provisioners?

pkg/controller/controller.go
```diff
@@ -844,14 +849,17 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
...
+			len(options.SelectedNodeName) > 0
 		state := checkError(err, mayReschedule)
 		klog.V(5).Infof("CreateVolume failed, supports topology = %v, node selected %v => may reschedule = %v => state = %v: %v",
 			p.supportsTopology(),
-			options.SelectedNode != nil,
+			options.SelectedNodeName,
 			mayReschedule,
 			state,
 			err)
+		if IsFinalError(err) {
```
sunnylovestiramisu: Check state instead, if the state is ProvisioningReschedule or ProvisioningFinished(failed with a final error), we remove the cache.

pkg/controller/cache.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"fmt"
+	"sync"
+)
+
+// TopologyInfo holds the data for NodeLabels and TopologyKeys
+type TopologyInfo struct {
+	NodeLabels       map[string]string
+	TopologyKeys     []string
+	SelectedNodeName string
```
sunnylovestiramisu: SelectedNodeName can be removed because to delete the cache for final error or reschedule.

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -417,7 +417,7 @@ func main() {
 	if supportsMigrationFromInTreePluginName != "" {
 		provisionerOptions = append(provisionerOptions, controller.AdditionalProvisionerNames([]string{supportsMigrationFromInTreePluginName}))
 	}
-
+	pvcNodeStore := ctrl.NewInMemoryStore()
```
msau42: If the driver doesn't support topology, then I think the cache is not needed.

pkg/controller/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	UpdateSelectedNodeName(name string, newName string)
+}
+
+// InMemoryStore is a concrete implementation of TopologyProvider.
+// It uses an in-memory map for quick lookups.
+type InMemoryStore struct {
+	// The map key is the object's name.
+	data map[string]*TopologyInfo
+	// Adding a mutex for thread-safe access
+	mutex sync.Mutex
```
msau42: Do you want to use RWLock instead?

pkg/controller/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	NodeLabels       map[string]string
+	TopologyKeys     []string
+	SelectedNodeName string
+}
+
+// TopologyProvider is an interface that defines the behavior for looking up
+// a TopologyInfo object by its name.
+// The name is pvcNamespace/pvcName
+type TopologyProvider interface {
+	GetByName(name string) (*TopologyInfo, error)
```
msau42: To make it very clear what the key is, what if we make it `v1.PersistentVolumeClaim*` instead of a string?

pkg/controller/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	Delete(name string) error
```
msau42: Should `Add` also be part of the interface?

pkg/controller/cache.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Delete implements the TopologyProvider interface.
+// It uses the built-in delete() function to remove the item from the map.
+func (s *InMemoryStore) Delete(name string) error {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	// First, check if the key exists to provide a helpful error.
+	_, found := s.data[name]
+	if !found {
+		return fmt.Errorf("cannot delete: object with name '%s' not found", name)
```
msau42: Do we want to error if it doesn't exist? I don't think the caller will do anything other than ignore the error. If you do want to return an error, I would make an explicit error type for NotFoundError, so that the caller can distinguish it from other errors.

pkg/controller/cache.go
```diff
@@ -0,0 +1,113 @@
...
+
+// TopologyProvider is an interface that defines the behavior for looking up
+// a TopologyInfo object by its name.
+// The name is pvcNamespace/pvcName
+type TopologyProvider interface {
+	GetByName(name string) (*TopologyInfo, error)
+	Delete(name string) error
+
+	// Update methods now perform an "upsert" and don't return errors.
+	UpdateNodeLabels(name string, newLabels map[string]string)
```
msau42: Would we ever update Node labels independently from TopologyKey? If not, I think we can combine it under one call.

pkg/controller/controller.go
```diff
@@ -1484,18 +1494,26 @@ func (p *csiProvisioner) checkNode(ctx context.Context, claim *v1.PersistentVolu
 		if len(sc.AllowedTopologies) > 0 {
 			node, err := p.nodeLister.Get(p.nodeDeployment.NodeName)
 			if err != nil {
-				return false, err
+				if !apierrors.IsNotFound(err) {
```
msau42: this `checkNode` function is only called if this csi driver is a "local volume" csi driver that has a provisioner running on every node.

So I think you can remove the entire `Get` call here and jsut pass `p.nodeDeployment.NodeName` directly into `GenerateAccessibilityRequirements`

pkg/controller/topology.go
```diff
@@ -452,6 +477,28 @@ func flatten(allowedTopologies []v1.TopologySelectorTerm) []topologyTerm {
 	return finalTerms
 }
 
+func getTopologyKeysFromCache(pvcNodeStore *InMemoryStore, pvcKey string, nodeName string) ([]string, error) {
+	cacheInfo, err := pvcNodeStore.GetByName(pvcKey)
+	if err != nil {
+		return nil, err
+	}
+	if len(cacheInfo.TopologyKeys) == 0 {
+		return nil, fmt.Errorf("no topology key found in the CSINode in memory cache %s", nodeName)
```
msau42: Do we need a special error message just for no keys/label from cache? We already have a warning message that we're going to lookup the cache.

pkg/controller/topology.go
```diff
@@ -157,12 +160,15 @@ func GenerateAccessibilityRequirements(
 	)
 
 	// 1. Get CSINode for the selected node
-	if selectedNode != nil {
-		selectedCSINode, err = getSelectedCSINode(csiNodeLister, selectedNode)
+	var topologyKeys []string
+	if len(selectedNodeName) > 0 {
```
msau42, line 191: This logic is very hard to follow with the cache lookups and updates being interspersed with condition checks.  I think it would be easier to read if we wrapped the informer/cache lookup logic in one call, and kept the core `GenerateAccessibility` function clean with just business logic. Something like:

```
func topologyLookup(selectedNodeName) {

csiNode, err1 := getSelectedCSINode(csiNodeLister, selectedNodeName)
node,err2 := getNode(nodeLister, selectedNodeName)

if csiNode == nil || node  == nil {
  topologyInfo = cache.getTopologyInfo(pvc)
} else {
  topologyInfo = { getTopologyKeys(csiNode), getNodeLabels(node, topologyKeys)}
  cache.addToplogyInfo(topologyInfo)
}

return topologyInfo
}
```

vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v11/controller/controller.go
```diff
@@ -1486,31 +1486,19 @@ func (ctrl *ProvisionController) provisionClaimOperation(ctx context.Context, cl
...
-				ctx2 := klog.NewContext(ctx, logger)
-				return ctrl.provisionVolumeErrorHandling(ctx2, ProvisioningReschedule, err, claim)
-			}
-			err = fmt.Errorf("failed to get target node: %v", err)
-			ctrl.eventRecorder.Event(claim, v1.EventTypeWarning, "ProvisioningFailed", err.Error())
-			return ProvisioningNoChange, err
-		}
+		selectedNodeName = nodeName
+	} else {
+		logger.Info("No selected node annotation")
```
msau42: This will spam for drivers that don't support topology or use Immediate binding

pkg/controller/topology.go
```diff
@@ -461,8 +508,69 @@ func getTopologyKeys(csiNode *storagev1.CSINode, driverName string) []string {
...
+				if !ok {
+					return nil, nil, true
+				}
+				term = append(term, topologySegment{key, v})
+			}
+			term.sort()
+			return term, node.Labels, false
+		}
+	} else {
+		// nodeLister is nil, read from the cache directly
```
msau42: Is this possible? I think nodeLister is required.

pkg/controller/topology.go
```diff
@@ -461,8 +508,69 @@ func getTopologyKeys(csiNode *storagev1.CSINode, driverName string) []string {
...
+		var err error
+		nodeLabels, err = getNodeLabelsFromCache(pvcNodeStore, pvcKeyForStore, nodeName)
+		if err != nil {
+			klog.Warningf("no node labels for node %s found in memory cache", nodeName)
+			return nil, nil, true
+		}
+		for _, key := range topologyKeys {
```
msau42: This logic looks repeated 3 times in this function. It would be better to share it across all 3 branches.

msau42:
For your e2e test case steps, can you also test deleting the PVC object along with the Pod and Node to trigger the leak?  I guess the leak is triggered by 2 scenarios:

1. PVC is deleted along with Pod and Node
2. PVC is not deleted, but Pod and Node are deleted. AND the scheduler retries in a different zone.

Can you verify you can repro the issue with both those scenarios, and that the fix works in both those scenarios?

I'm trying to think if AllowedTopologies or Immediate binding mode would be impacted.  AllowedTopologies shouldn't be impacted because we grab the topology directly from it. I think Immediate binding could be impacted in a scenario like:
1. Start provisioning in zone A.
2. During provisioning, scale down all the Nodes in zone A.
3. Future retries will either error if no zones are found, or it will retry with a zone B, leaking the volume being provisioned in zone A.

pkg/controller/topology.go
msau42: We can also consider looking up cache first, then if it doesn't exist in cache, then lookup in informers. I think switching that logic will solve the immediate binding scenario I outlined in https://github.com/kubernetes-csi/external-provisioner/pull/1413#issuecomment-3268571209

msau42:
Also in the testing steps, with the fix, verify that the PV object is created and then deleted when the disk gets deleted.

pkg/controller/cache.go
```diff
@@ -0,0 +1,113 @@
...
+// TopologyInfo holds the data for NodeLabels and TopologyKeys
+type TopologyInfo struct {
+	NodeLabels       map[string]string
+	TopologyKeys     []string
+	SelectedNodeName string
+}
+
+// TopologyProvider is an interface that defines the behavior for looking up
+// a TopologyInfo object by its name.
+// The name is pvcNamespace/pvcName
```
jsafrane: IMHO the key should be PVC UID.

Consider this scenario:
1. User creates a PVC
2. The provisioning takes a long time, i.e. many timeouts + retries.
3. User deletes the PVC and creates a new one, with the same name, but different StorageClass.
4. The original PVC is still being provisioned. Despite it has been deleted from the API server, the provisioner waits for a final error.
5. The new PVC should be provisioned too.
-> Two PVCs have the same name and are being provisioned, but they have a different StorageClass and thus different topology cache.

PVC UID will be unique for these PVCs.

pkg/controller/cache.go
jsafrane: +1 for returning no error.

vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v11/controller/controller.go
jsafrane: This is a breaking change, so yes, it will require a new major version.
I don't see where you make it an option, this PR looks very backward incompatible. And I am not against.

pkg/controller/controller.go
jsafrane: If I understand it correctly, as long as the PVC exists and the provisioner lib calls Provision(), the external-provisioner re-loads the topology from the API server, adds it to the cache and deletes it from the cache on every final error.

It's actually a smart idea that simplifies the cache cleanup. It's just not very obvious from the cache name and its comments. I somehow expected it caches the topology _forever_ and deletes it after PVC is deleted. A comment on the right place (the `struct` definition?) would help.

pkg/controller/topology.go
jsafrane: I would prefer checking the informers first, the node might have gotten new labels as the CSI driver just got installed there.

vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v11/controller/controller.go
msau42: The alternative, if we think non-csi provisioners still want the old behavior, is to add the new changes as a feature gate. I'm not aware of any non-csi provisioners that support topology though and they would still run into the topology bug if we keep the old way.

So I am also in favor of just making a new major.

pkg/controller/topology.go
sunnylovestiramisu: The original logic in the lib is:
```
if ctrl.nodeLister != nil {
	selectedNode, err = ctrl.nodeLister.Get(nodeName)
} else {
	selectedNode, err = ctrl.client.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{}) // TODO (verult) cache Nodes
}
```
So I was thinking nodeLister can be nil? I am not sure when will it occur though.

cmd/csi-provisioner/csi-provisioner.go
sunnylovestiramisu: So we need to add Topology check for all the cache code.

pkg/controller/cache.go
sunnylovestiramisu: And PVC UID guarantees uniqueness as well.

pkg/controller/cache.go
sunnylovestiramisu: What happens if PVC is deleted? Also currently we only pass in pvcName and pvcNamespace in the GenerateAccessibilityRequirements. Can changing to a pointer to PVC cause problems?

pkg/controller/cache.go
sunnylovestiramisu: Oh yes, but currently there is no code using Add. Should we just delete the function.

pkg/controller/cache.go
sunnylovestiramisu: Do we want to just log a warning and not returning error?

pkg/controller/topology.go
msau42: In the csi-provisioner it is mandatory as long as the plugin supports topology: https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L339

pkg/controller/cache.go
sunnylovestiramisu: Shared read access should be fine, I can change to RWLock

pkg/controller/cache.go
sunnylovestiramisu: Removed Add func

pkg/controller/topology.go
sunnylovestiramisu: Okay then we can remove the nodeLister nil check

pkg/controller/topology.go
sunnylovestiramisu: Removed one for the nodeLister is nil code. However the rest two, the first one read nodeLabels from Cache, the other one read nodeLabels from the node. I think we can make them to a processTerm func.

pkg/controller/topology.go
sunnylovestiramisu: We do not need the special error message but we should return an error. Maybe errors.New("")?

vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v11/controller/controller.go
sunnylovestiramisu: Okay so we just do not log anything

pkg/controller/cache.go
```diff
@@ -50,22 +51,24 @@ func (s *InMemoryStore) Add(name string, info *TopologyInfo) {
 
 // Delete implements the TopologyProvider interface.
 // It uses the built-in delete() function to remove the item from the map.
+// The entry is deleted when provision succeeds or returns a final error.
```
msau42: This is tue regardless of implementation. I would probably put it as a comment in the interface

pkg/controller/controller.go
```diff
@@ -692,7 +692,7 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		requirements, err := GenerateAccessibilityRequirements(
 			p.client,
 			p.driverName,
-			claim.Namespace,
+			string(claim.UID),
```
msau42: do we need to convert the key to a string? Can we use the direct type as the key?

pkg/controller/topology.go
```diff
@@ -105,6 +106,78 @@ func SupportsTopology(pluginCapabilities rpc.PluginCapabilitySet) bool {
 		utilfeature.DefaultFeatureGate.Enabled(features.Topology)
 }
 
+func topologyLookup(
```
msau42: This is not quite what I meant about the wrapper function.

I think it would be simpler to understand the code if things like obj lookup (whether from informer or cache) is abstracted from business logic (how to behave when there are no keys, or strictTopology mode is set, etc).

So I think a better structure for `GenerateAccessiblityRequirements` would be something like:
```
func GenerateAccessibilityRequirements(...) (...) {

  // This returns the information from informer or cache, the caller doesn't care
   topologyInfo = topologyLookup(selectedNode)

  // Nothing in the logic after this should care if the information came from the cache or informer

   if len(topologyKeys) == 0) {
     // business logic to handle it
   }

  if strictTopology {
    // business logic to handle it
  }

  // everything else
}

```

pkg/controller/cache.go
```diff
@@ -0,0 +1,104 @@
...
+type TopologyInfo struct {
+	NodeLabels   map[string]string
+	TopologyKeys []string
+}
+
+// TopologyProvider is an interface that defines the behavior for looking up
+// a TopologyInfo object by its name.
+// The name is pvc UID
+type TopologyProvider interface {
+	Add(name string, info *TopologyInfo)
```
jsafrane: nit: I would rename all `name` parameters to `pvcUID` - everyone then can see what the key is without reading comments on the interface. And I would consider `types.UID` as the type instead of `string`

pkg/controller/topology.go
sunnylovestiramisu: I do not feel too comfortable to mix a sizable refactor with a feature change in one PR and release(the small cls best practices, higher risks of rollback). Or if you think the current test coverage we have already is sufficient to catch significant issues of refactoring, we can make it.

pkg/controller/cache.go
sunnylovestiramisu: There is a case that csinode is missing but node is still present, so for current logic, we do update them separately.

pkg/controller/controller.go
sunnylovestiramisu: Yeah we can use it as a direct type.

pkg/controller/cache.go
sunnylovestiramisu: Done

vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v11/controller/controller.go
sunnylovestiramisu: Raised PR for the lib: https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/194

sunnylovestiramisu:
> Also in the testing steps, with the fix, verify that the PV object is created and then deleted when the disk gets deleted.

This is verified in the logs, it says PV object successfully created. And:

- If PVC is not deleted, it will show it is bound to a PV(and disk)
- If PVC is deleted, in the provisioner logs, it shows a DeleteVolume call deleted the PV(and disk)

pkg/controller/cache.go
```diff
@@ -0,0 +1,107 @@
...
+
+// NewInMemoryStore creates and initializes a new store.
+func NewInMemoryStore() *InMemoryStore {
+	return &InMemoryStore{
+		data: make(map[types.UID]*TopologyInfo),
+	}
+}
+
+// Add is a helper function to populate our store with data.
+func (s *InMemoryStore) Add(pvcUID types.UID, info *TopologyInfo) {
```
jsafrane: This function is not called anywhere in the controller.

pkg/controller/cache.go
jsafrane, line 1: I tested this PR in a single-node cluster and I deleted the only Node + CSINode in the cluster while my CSI driver was returning `DeadlineExceeded` to all `CreateVolume` calls.

The cache had the deleted node, but still, provisioning ended with `No topology keys found on any node` here, because there was no CSINode in the whole cluster:
https://github.com/kubernetes-csi/external-provisioner/blob/be69ca2e5789a30f014d2c7a9f6b20e9d846f9b0/pkg/controller/topology.go#L319-L332

Should we expect the cluster has at least one working Node + CSINode? I admit it's very niche case. Maybe when someone scales their cluster to 0 workers?

sunnylovestiramisu: Yeah this is the immediate binding mode I need to fix next, like we will cache the topologyKeys, and use it when if len(topologyKeys) == 0

sunnylovestiramisu: I tested the latest version of the code and I got:
```
I0912 18:16:17.663128       1 controller.go:968] successfully created PV pvc-a064b742-1f17-4f08-8c36-2d465ceae5ab for PVC pd-pvc and csi volume name projects/XXXXXXX/zones/us-central1-a/disks/pvc-a064b742-1f17-4f08-8c36-2d465ceae5ab
```
for the immediate binding case, when the nodes are scaled down to 0 in the zone of the PV is still creating.

sunnylovestiramisu: But now it is still leaking volume. It creates two disks with the exact same name in two different zones. PVC is bound to the disk created later in the new zone.

pkg/controller/topology.go
```diff
@@ -367,12 +367,18 @@ func aggregateTopologies(
 		}
 
 		if len(topologyKeys) == 0 {
+			// Try retrieving from in memory cache
+			topologyKeys, err = getTopologyKeysFromCache(pvcNodeStore, pvcKeyForStore)
```
sunnylovestiramisu: @msau42 This is the change for Immediate

pkg/controller/topology.go
msau42: I will propose a commit with my changes because I think it will be easier for me to explain how I think the new code could be structured to make it simpler to read and understand. It should not be a big refactor.

pkg/controller/cache.go
msau42: We discussed offline and have a strong suspicion of where the fix for immediate binding needs to go

sunnylovestiramisu: Okay I found the issue, it does not exist anymore.

pkg/controller/topology.go
msau42: Here's my branch with a commit on top that restructures the new changes; https://github.com/kubernetes-csi/external-provisioner/compare/master...msau42:external-provisioner:leakedVolumeCacheAlt

The thing that I prefer about this approach, is that when you review the changes to topology.go, none of the core business logic has been changed. The only things that have been replaced are the "getTopologyKeys" and the "getNodeTopology" calls and the logic to retrieve the information from informer vs cache have been abstracted from the caller. This makes the change easier to review because the core logic hasn't changed so the reviewer only needs to reason about the new "get" calls.

pkg/controller/topology.go
```diff
@@ -317,10 +359,17 @@ func aggregateTopologies(
 		})
 
 		// Pick the first node with topology keys
-		for _, csiNode := range csiNodes {
-			topologyKeys = getTopologyKeys(csiNode, driverName)
-			if topologyKeys != nil {
-				break
+		topologyKeys, err = getTopologyKeysFromCache(pvcNodeStore, pvcKeyForStore)
```
msau42, line 344: It would be more efficient to check this first before we List + sort all the csinodes

pkg/controller/topology_test.go
```diff
@@ -1456,6 +1494,183 @@ func TestPreferredTopologies(t *testing.T) {
...
+				{Segments: map[string]string{"com.example.csi/zone": "zone1", "com.example.csi/rack": "rackA"}},
+				{Segments: map[string]string{"com.example.csi/zone": "zone2", "com.example.csi/rack": "rackB"}},
+			},
+			expectedPreferred: []*csi.Topology{
+				{Segments: map[string]string{"com.example.csi/zone": "zone1", "com.example.csi/rack": "rackA"}},
+				{Segments: map[string]string{"com.example.csi/zone": "zone2", "com.example.csi/rack": "rackB"}},
+			},
+		},
+		{
+			name:              "delayed-binding, strict-with-allowed-topology",
```
msau42: what about strict without allowed topologies?

pkg/controller/topology_test.go
```diff
@@ -1456,6 +1494,183 @@ func TestPreferredTopologies(t *testing.T) {
...
+		name              string
+		strictTopology    bool
+		immediateTopology bool
+		allowedTopologies []v1.TopologySelectorTerm
+		selectedNode      string
+		expectedRequisite []*csi.Topology
+		expectedPreferred []*csi.Topology
+	}{
+		{
+			name:              "delayed-binding, non-strict",
```
msau42: what about non-strict with allowed topologies?

pkg/controller/topology_test.go
```diff
@@ -1456,6 +1494,183 @@ func TestPreferredTopologies(t *testing.T) {
...
+			selectedNode: "node-0",
+			expectedRequisite: []*csi.Topology{
+				{Segments: map[string]string{"com.example.csi/zone": "zone1", "com.example.csi/rack": "rackA"}},
+			},
+			expectedPreferred: []*csi.Topology{
+				{Segments: map[string]string{"com.example.csi/zone": "zone1", "com.example.csi/rack": "rackA"}},
+			},
+		},
+		{
+			name:              "immediate-no-selected-node, non-strict",
```
msau42: what about non-strict with allowed topologies?

pkg/controller/topology_test.go
```diff
@@ -1456,6 +1494,183 @@ func TestPreferredTopologies(t *testing.T) {
...
+				{Segments: map[string]string{"com.example.csi/zone": "zone1", "com.example.csi/rack": "rackA"}},
+				{Segments: map[string]string{"com.example.csi/zone": "zone2", "com.example.csi/rack": "rackB"}},
+			},
+			expectedPreferred: []*csi.Topology{
+				{Segments: map[string]string{"com.example.csi/zone": "zone1", "com.example.csi/rack": "rackA"}},
+				{Segments: map[string]string{"com.example.csi/zone": "zone2", "com.example.csi/rack": "rackB"}},
+			},
+		},
+		{
+			name:              "immediate-no-selected-node, strict-with-allowed-topology",
```
msau42: what about strict without allowed topologies?

pkg/controller/topology_test.go
```diff
@@ -1493,6 +1494,251 @@ func TestPreferredTopologies(t *testing.T) {
...
+			},
+		},
+		{
+			name:              "delayed-binding, non-strict, with-allowed-topology",
+			strictTopology:    false,
+			immediateTopology: false,
+			allowedTopologies: []v1.TopologySelectorTerm{
+				{
+					MatchLabelExpressions: []v1.TopologySelectorLabelRequirement{
+						{Key: "com.example.csi/zone", Values: []string{"zone1"}},
```
msau42, line 1561: for the allowed topology test cases, it would be better to use values that are different from what we store in the cache so that we can be sure we're not using the cache values.

sunnylovestiramisu: But this is to test the node has been deleted? How can we pass in a different set of allowedTopologies if it is for the same CreateVolume retry?

pkg/controller/cache.go
msau42: Looks like this can be removed.

pkg/controller/controller.go
```diff
@@ -844,14 +849,18 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
...
 		klog.V(5).Infof("CreateVolume failed, supports topology = %v, node selected %v => may reschedule = %v => state = %v: %v",
 			p.supportsTopology(),
-			options.SelectedNode != nil,
+			options.SelectedNodeName,
 			mayReschedule,
 			state,
 			err)
+		// Delete the entry in in memory cache if the error is final
+		if state == controller.ProvisioningFinished || state == controller.ProvisioningReschedule {
+			p.pvcNodeStore.Delete(claim.UID)
```
msau42, line 862: Does this need to check for `pvcNodeStore == nil`? Are we missing a test case?

pkg/controller/controller.go
```diff
@@ -973,6 +982,8 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 	}
 
 	klog.V(5).Infof("successfully created PV %+v", pv.Spec.PersistentVolumeSource)
+	// Remove entry from the in memory cache
+	p.pvcNodeStore.Delete(claim.UID)
```
msau42: Does this need to check for `pvcNodeStore == nil`? Are we missing a test case?

pkg/controller/topology.go
```diff
@@ -452,6 +507,50 @@ func flatten(allowedTopologies []v1.TopologySelectorTerm) []topologyTerm {
...
+	return cacheInfo.RequisiteTerms, nil
+}
+
+func getPreferredTermsFromCache(pvcNodeStore TopologyProvider, pvcKey types.UID) ([]topologyTerm, error) {
+	cacheInfo, err := pvcNodeStore.GetByPvcUID(pvcKey)
+	if err != nil {
+		return nil, err
+	}
+	if len(cacheInfo.PreferredTerms) == 0 {
+		return nil, nil
```
msau42, line 523: These functions return nil error when nothing is found, but the previous two return an error. It should be consistent across all methods. My preference is nil error is returned.

pkg/controller/topology.go
```diff
@@ -256,19 +285,23 @@ func GenerateAccessibilityRequirements(
 			// In case of strict topology, preferred = requisite
 			preferredTerms = requisiteTerms
 		} else {
-			for i, t := range requisiteTerms {
-				if t.subset(selectedTopology) {
-					preferredTerms = append(requisiteTerms[i:], requisiteTerms[:i]...)
-					break
+			preferredTerms, err = getPreferredTermsFromCache(pvcNodeStore, pvcUID)
```
msau42: Add a comment why we need to read from cache first because it's not obvious

pkg/controller/topology.go
```diff
@@ -279,16 +312,18 @@ func GenerateAccessibilityRequirements(
...
+		}
 		// We don't want to fallback and provision in the wrong topology if there's some temporary
 		// error with the API server.
-		return nil, fmt.Errorf("error getting CSINode for selected node %q: %v", selectedNode.Name, err)
-	}
-	if selectedCSINode == nil {
-		return nil, fmt.Errorf("CSINode for selected node %q not found", selectedNode.Name)
+
+		// Try to get from cache
+		return nil, fmt.Errorf("error getting CSINode for selected node %q: %v", selectedNodeName, err)
```
msau42: I don't know if we need to wrap the error. It seems like the caller is already wrapping the error.

pkg/controller/topology.go
```diff
@@ -297,30 +332,40 @@ func getSelectedCSINode(
...
+	pvcNodeStore TopologyProvider,
+	strictTopology bool) ([]topologyTerm, error) {
 	// 1. Determine topologyKeys to use for aggregation
 	var topologyKeys []string
-	if selectedCSINode == nil {
+	var err error
+	if len(selectedNodeName) == 0 {
 		// Immediate binding
-		csiNodes, err := csiNodeLister.List(labels.Everything())
+		topologyKeys, err = getTopologyKeysFromCache(pvcNodeStore, pvcKeyForStore)
```
msau42, line 344: Add a comment why we need to read from cache first.

pkg/controller/topology.go
```diff
@@ -344,7 +388,8 @@ func aggregateTopologies(
 			//
 			// Returning an error in provisioning will cause the scheduler to retry and potentially
 			// (but not guaranteed) pick a different node.
-			return nil, fmt.Errorf("no topology key found on CSINode %s", selectedCSINode.Name)
+			topologyKeys, err = getTopologyKeysFromCache(pvcNodeStore, pvcKeyForStore)
```
msau42: `selectedNodeTopologyKeys` already should have read from cache. We shouldn't need to read from cache again.

pkg/controller/topology.go
```diff
@@ -361,21 +406,31 @@ func aggregateTopologies(
...
-		return nil, fmt.Errorf("topologyKeys %v were not found on any nodes", topologyKeys)
+		for _, node := range nodes {
+			term, _ := getTopologyFromNode(node, topologyKeys)
+			if len(term) > 0 {
+				terms = append(terms, term)
+			}
+		}
+
+		if len(terms) == 0 {
+			// Try again to read from cache
```
msau42: this comment doesn't seem reflect the code

pkg/controller/topology.go
```diff
@@ -361,21 +406,31 @@ func aggregateTopologies(
 	if err != nil {
 		return nil, err
 	}
-	nodes, err := nodeLister.List(selector)
+	var terms []topologyTerm
+	terms, err = getRequisiteTermsFromCache(pvcNodeStore, pvcKeyForStore)
```
msau42, line 406: In `aggregateTopologies`, we read from cache first and save requisite/preferred terms based on that.

But for topology keys and node labels with selectedNode, we read from informer first, then cache.

I'm concerned this could result in some subtle consistency issue. I think we should be consistent and read from cache first in all scenarios. I believe @jsafrane  had a concern that topology could change on a retry, but IMO if we have already started an operation, we need to retry with the same arguments. Only if it returns final error, we clear the cache and retry again with potentially different arguments

pkg/controller/topology.go
```diff
@@ -452,6 +507,50 @@ func flatten(allowedTopologies []v1.TopologySelectorTerm) []topologyTerm {
 	return finalTerms
 }
 
+func getTopologyKeysFromCache(pvcNodeStore TopologyProvider, pvcKey types.UID) ([]string, error) {
```
msau42, line 506: Do we need these wrappers? It seems like callers could directly call 'GetByPvcUID` and reference the field they need.

pkg/controller/topology.go
```diff
@@ -461,8 +560,48 @@ func getTopologyKeys(csiNode *storagev1.CSINode, driverName string) []string {
...
+func getTopologyFromNodeName(nodeName string, topologyKeys []string, nodeLister corelisters.NodeLister, pvcKeyForStore types.UID, pvcNodeStore TopologyProvider) (term topologyTerm, selectedNodeLabels map[string]string, isMissingKey bool) {
+	// Get Node Here
+	if nodeLister != nil {
+		node, err := nodeLister.Get(nodeName)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				// Read from the cache
+				nodeLabels, err := getNodeLabelsFromCache(pvcNodeStore, pvcKeyForStore)
+				if err != nil || len(nodeLabels) == 0 {
+					klog.Warningf("no node labels for node %s found in memory cache", nodeName)
```
msau42: Do we need this error message? The caller is already returning an error for this scenario right?

pkg/controller/topology_test.go
msau42: Another idea I have to be able to distinguish the test cases between strict and allowedTopology better is to have 3 topologies in the cluster. Then:

1) Strict expects only the first topology to be returned
2) AllowedTopology could restrict to 2 topologies (and not necessarily the first 2)
3) non-strict, non-allowed topology retruns all 3 topologies

pkg/controller/topology.go
jsafrane: I agree it's inconsistent. I think using cache first would work, if we clear the cache on a final error and we assume the drivers return INVALID_ARGUMENT (= final error) when there is something wrong with topology.

Use case: an user has mis-labeled nodes, the driver returns errors on CreateVolume, the user fixes the labels, and they expect the provisioner picks new API objects. It will work so if the driver returns INVALID_ARGUMENT and not DEADLINE_EXCEEDED or so :-)

pkg/controller/topology.go
sunnylovestiramisu: GetByPvcUID returns an error if not found. It does not return an error if the list is empty. 

But overall the wrapper functions make the code easier to read and provide specificity. topology.go doesn't need to be concerned with the TopologyInfo struct or the raw pvcNodeStore interface.

And we do not really have a use case to access multiple fields from the cache at one Get call.

pkg/controller/topology.go
sunnylovestiramisu: If we do not return an error, the current logic is not correct. 

topologyKeys, err = getTopologyKeysFromCache(pvcNodeStore, pvcKeyForStore)
		if err != nil {
			...
            // Get from listing objects
}

sunnylovestiramisu: We will need to change it to if err != nil || len(topologyKeys) == 0

pkg/controller/topology.go
sunnylovestiramisu: It does not return an error, it returns nil, nil, true so the error message is lost

pkg/controller/topology_test.go
sunnylovestiramisu: For 2, I restrict to 2 because i was intentionally deleting the first node to make the node "deleted".

pkg/controller/topology.go
msau42: In general, I think the callers shoud distinguish between temporary API error vs not found.

So returning nil error for not found is ideal so callers can check for len(x) == 0 separately.

pkg/controller/topology.go
msau42: What I mean is that the caller of `getTopologyFromNodeName()` is also logging a similar error, so this gets logged twice.

sunnylovestiramisu: Removed

pkg/controller/controller_test.go
```diff
@@ -426,10 +426,11 @@ func TestCreateDriverReturnsInvalidCapacityDuringProvision(t *testing.T) {
 
 	var clientSetObjects []runtime.Object
 	clientSet := fakeclientset.NewSimpleClientset(clientSetObjects...)
+	pvcNodeStore := NewInMemoryStore()
```
msau42: I think it's better to only provision the store if topology is enabled. Then we could have caught the issue earlier when we were not doing the nil check.

pkg/controller/topology.go
```diff
@@ -104,6 +106,31 @@ func SupportsTopology(pluginCapabilities rpc.PluginCapabilitySet) bool {
 		utilfeature.DefaultFeatureGate.Enabled(features.Topology)
 }
 
+func topologyKeysLookup(
+	csiNodeLister storagelistersv1.CSINodeLister,
+	selectedNodeName string,
+	driverName string,
+	pvcNodeStore TopologyProvider,
+	pvcUID types.UID) (topologyKeys []string, err error) {
+	selectedCSINode, err := getSelectedCSINode(csiNodeLister, selectedNodeName)
```
msau42, line 121: from discussion in https://github.com/kubernetes-csi/external-provisioner/pull/1413/files#r2377144860, we should read topology keys and node labels from cache first.

pkg/controller/topology.go
```diff
@@ -238,37 +268,56 @@ func GenerateAccessibilityRequirements(
...
+	// but should not be considered for placement.
+	requisiteTerms = filterEmptyTerms(requisiteTerms)
 	requirement.Requisite = toCSITopology(requisiteTerms)
 
 	// 3. Generate CSI Preferred Terms
 	var preferredTerms []topologyTerm
-	if selectedCSINode == nil {
+	if len(selectedNodeName) == 0 {
 		// Immediate binding, we fallback to statefulset spreading hash for backwards compatibility.
+		preferredTerms, err = getPreferredTermsFromCache(pvcNodeStore, pvcUID)
```
msau42: Do we actually need to cache preferred terms? Preferred is calculated from requisite, and shouldn't require looking up csinode/node info.

pkg/controller/topology.go
```diff
@@ -238,37 +268,56 @@ func GenerateAccessibilityRequirements(
 	requisiteTerms = slices.CompactFunc(requisiteTerms, slices.Equal)
 	// TODO (verult) reduce subset duplicate terms (advanced reduction)
 
+	// Filter out empty topology terms to prevent issues with caching and spreading.
+	// An empty term can be generated from an empty TopologySelectorTerm,
+	// but should not be considered for placement.
+	requisiteTerms = filterEmptyTerms(requisiteTerms)
```
msau42: `toCSITopology` already skips empty terms. Do we still need `filterEmptyTerms` here?

pkg/controller/topology.go
```diff
@@ -238,37 +268,56 @@ func GenerateAccessibilityRequirements(
 	requisiteTerms = slices.CompactFunc(requisiteTerms, slices.Equal)
 	// TODO (verult) reduce subset duplicate terms (advanced reduction)
 
+	// Filter out empty topology terms to prevent issues with caching and spreading.
```
msau42: Also do you know where the empty terms are being generated? Is it from `aggregateTopologies`?

pkg/controller/topology_test.go
```diff
@@ -1456,6 +1494,305 @@ func TestPreferredTopologies(t *testing.T) {
...
 
+func TestProvisionWithDeletedNodeFromCache(t *testing.T) {
+	nodeLabels := []map[string]string{
+		{"com.example.csi/zone": "zone1", "com.example.csi/rack": "rackA"},
+		{"com.example.csi/zone": "zone2", "com.example.csi/rack": "rackB"},
+		{"com.example.csi/zone": "zone3", "com.example.csi/rack": "rackC"},
+	}
+	topologyKeys := []map[string][]string{
+		{testDriverName: {"com.example.csi/zone", "com.example.csi/rack"}},
+		{testDriverName: {"com.example.csi/zone", "com.example.csi/rack"}},
```
msau42, line 1505: I think you can delete the 2nd entry since the driver name and keys are the same.

pkg/controller/topology.go
```diff
@@ -297,30 +343,41 @@ func getSelectedCSINode(
...
-	selectedCSINode *storagev1.CSINode,
+	selectedNodeName string,
+	selectedNodeTopologyKeys []string,
 	csiNodeLister storagelistersv1.CSINodeLister,
-	nodeLister corelisters.NodeLister) ([]topologyTerm, error) {
-
+	nodeLister corelisters.NodeLister,
+	pvcKeyForStore types.UID,
+	pvcNodeStore TopologyProvider,
+	strictTopology bool) ([]topologyTerm, error) {
```
msau42: strictTopology doesn't seem to be used.

pkg/controller/topology.go
sunnylovestiramisu: I completely have no idea, the only thing I can think of the some race at modifying some values in a slice itself. Because it only happens when I tested it end to end, unit tests also just solid passing. And the empty slice does not have a specific step that added into the slice. Sometimes it is appended to the start of the slice, sometimes it is add the end:
```
// Attached to the beginning
I0926 21:30:29.536700       1 topology.go:262] ======== Generate CSI Requisite Terms [[] [{topology.gke.io/zone us-central1-a}] [{topology.gke.io/zone us-central1-b}] [{topology.gke.io/zone us-central1-c}]] ========
I0926 21:30:29.536783       1 topology.go:323] ======== Delayed binding preferredTerms [[{topology.gke.io/zone us-central1-a}] [{topology.gke.io/zone us-central1-b}] [{topology.gke.io/zone us-central1-c}]] ========

// Attached to the end
I0926 18:52:05.297685       1 topology.go:262] ======== Generate CSI Requisite Terms [[{topology.gke.io/zone us-central1-a}] [{topology.gke.io/zone us-central1-b}] [{topology.gke.io/zone us-central1-c}] [{topology.gke.io/zone us-central1-f}] [{topology.gke.io/zone us-central1-a}]] ========
I0926 18:52:05.297747       1 topology.go:317] ======== Delayed binding preferredTerms [[{topology.gke.io/zone us-central1-b}] [{topology.gke.io/zone us-central1-c}] [{topology.gke.io/zone us-central1-f}] []] ========

// Attached to the end 
I0926 18:52:06.298084       1 topology.go:262] ======== Generate CSI Requisite Terms [[{topology.gke.io/zone us-central1-a}] [{topology.gke.io/zone us-central1-b}] [{topology.gke.io/zone us-central1-c}] [{topology.gke.io/zone us-central1-f}] []] ========
I0926 18:52:06.298181       1 topology.go:317] ======== Delayed binding preferredTerms [[{topology.gke.io/zone us-central1-a}] [{topology.gke.io/zone us-central1-b}] [{topology.gke.io/zone us-central1-c}] [{topology.gke.io/zone us-central1-f}]] ========

// Attached to the end
I0926 18:52:06.298084       1 topology.go:262] ======== Generate CSI Requisite Terms [[{topology.gke.io/zone us-central1-a}] [{topology.gke.io/zone us-central1-b}] [{topology.gke.io/zone us-central1-c}] [{topology.gke.io/zone us-central1-f}] []] ========
I0926 18:52:06.298181       1 topology.go:317] ======== Delayed binding preferredTerms [[{topology.gke.io/zone us-central1-a}] [{topology.gke.io/zone us-central1-b}] [{topology.gke.io/zone us-central1-c}] [{topology.gke.io/zone us-central1-f}]] ========
```

That is why I took an approach that tries to eliminate it at generation.

pkg/controller/topology.go
sunnylovestiramisu: That is something I do not quite understand, it seems not filtering out the empty terms when I did the test end to end. And I added log examples to [this comment](https://github.com/kubernetes-csi/external-provisioner/pull/1413#discussion_r2383750102).

pkg/controller/topology_test.go
sunnylovestiramisu: This is for test setup. So it sets up com.example.csi/zone and com.example.csi/rack for different nodes.

In the actual flow the topologyKeys is only "com.example.csi/zone", "com.example.csi/rack" once.

pkg/controller/topology.go
sunnylovestiramisu: Removed

pkg/controller/topology.go
sunnylovestiramisu: Removed

pkg/controller/topology.go
sunnylovestiramisu: Removed

pkg/controller/controller.go
```diff
@@ -1481,21 +1494,23 @@ func (p *csiProvisioner) checkNode(ctx context.Context, claim *v1.PersistentVolu
 		// it must match our own. We can find out by trying to
 		// create accessibility requirements.  If that fails,
 		// we should not become the owner.
-		if len(sc.AllowedTopologies) > 0 {
+		if len(sc.AllowedTopologies) > 0 && p.supportsTopology() {
```
sunnylovestiramisu, line 1497: @jsafrane For distributed local volumes, it should support topologies correct?

pkg/controller/cache.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+	info, found := s.data[pvcUID]
+	if !found {
+		// To prevent printing UID in logs, we return a standard error
+		return nil, fmt.Errorf("topology object with given pvcUID not found")
+	}
+
+	// Return a deep copy to prevent data races
```
msau42, line 72: There is a `reflect.DeepCopy` you could use: https://pkg.go.dev/golang.design/x/reflect

sunnylovestiramisu: From unit test, this one is setting the struct to an empty slice instead of nil after copying, that is not the behavior we want?

Or for future PRs about this object, we need to somehow make sure everyone is checking len(info.XXX) > 0 instead of info.XXX != nil

pkg/controller/cache.go
```diff
@@ -69,9 +68,34 @@ func (s *InMemoryStore) GetByPvcUID(pvcUID types.UID) (*TopologyInfo, error) {
 	defer s.mutex.RUnlock()
 	info, found := s.data[pvcUID]
 	if !found {
-		return nil, fmt.Errorf("topology object with pvcUID '%s' not found", pvcUID)
+		// To prevent printing UID in logs, we return a standard error
+		return nil, fmt.Errorf("topology object with given pvcUID not found")
```
msau42: what's wrong with printing pvcUID? Also I thought we are returning nil error when the entry is not found so that we can distinguish between an error and not found?

pkg/controller/cache.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Delete implements the TopologyProvider interface.
+// It uses the built-in delete() function to remove the item from the map.
+func (s *InMemoryStore) Delete(pvcUID types.UID) error {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	// First, check if the key exists to provide a helpful error.
+	_, found := s.data[pvcUID]
+	if !found {
+		klog.Warningf("cannot delete: object with pvcUID '%s' not found", pvcUID)
```
msau42: do we need this? If something returns an error early, it's possible that nothing gets added to the cache

pkg/controller/topology.go
```diff
@@ -279,16 +312,13 @@ func GenerateAccessibilityRequirements(
...
+	selectedNodeName string) (*storagev1.CSINode, error) {
 
-	selectedCSINode, err := csiNodeLister.Get(selectedNode.Name)
+	selectedCSINode, err := csiNodeLister.Get(selectedNodeName)
 	if err != nil {
 		// We don't want to fallback and provision in the wrong topology if there's some temporary
 		// error with the API server.
-		return nil, fmt.Errorf("error getting CSINode for selected node %q: %v", selectedNode.Name, err)
-	}
-	if selectedCSINode == nil {
```
msau42, line 290: Should we keep this check to retain original behavior?

pkg/controller/topology.go
```diff
@@ -461,8 +543,49 @@ func getTopologyKeys(csiNode *storagev1.CSINode, driverName string) []string {
...
+}
+
+func getTopologyFromNodeName(nodeName string, topologyKeys []string, nodeLister corelisters.NodeLister, pvcKeyForStore types.UID, pvcNodeStore TopologyProvider) (term topologyTerm, selectedNodeLabels map[string]string, isMissingKey bool) {
+	// Get Node Here
+	if nodeLister != nil {
+		node, err := nodeLister.Get(nodeName)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				// Read from the cache
+				nodeLabels, err := getNodeLabelsFromCache(pvcNodeStore, pvcKeyForStore)
```
msau42: Should this also read from cache first?

pkg/controller/cache.go
sunnylovestiramisu: Do we need signal at all if there is something wrong with the cache? Like somehow the cache is not written properly so it is missing from deletion.

pkg/controller/topology.go
sunnylovestiramisu: The Get call does not return nil, it returns an error is the csinode is not there? I do not know why it is here to begin with.

pkg/controller/topology.go
sunnylovestiramisu: Swapped.

pkg/controller/cache.go
sunnylovestiramisu: Removed

pkg/controller/topology.go
sunnylovestiramisu: Hmm actually there may still be a chance that the csinode as the value in the map can be nil, I was mistaken it for en non-exist key. Added the logic back.

msau42:
/approve

Can you squash your commits?

PR #1417, opened 2025-09-22
Restart csi sidecars faster

rhrmo:
**What type of PR is this?**

/kind feature

This PR makes external-provisioner restart faster when it is terminating because of sigterm. This is accomplished by releasing leader election lease on sigterm and then it does not have to wait for lease to expire but it proceeds with restart right away. This is a PR from series of PRs that fix https://github.com/kubernetes-csi/external-attacher/issues/609

Fixes #

```release-note
NONE
```


cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -661,13 +687,35 @@ func main() {
...
+				}()
+			}
+			if csiClaimController != nil {
+				wg.Add(1)
+				go func() {
+					defer wg.Done()
+					csiClaimController.Run(controllerCtx, int(*finalizerThreads), &wg)
+				}()
+			}
+			provisionController.ControllerWaitGroup(&wg)
```
jsafrane, line 706: Ouch, this should have been an `option` in `NewProvisionController()`. Too late to change this, I already tagged the lib-external-provisioner with this function.
Note to self: review external-provisioner PRs before merging things to the provisioner lib.

PR #1427, opened 2025-10-21
Bump CSI spec to 1.12

jsafrane:
**What type of PR is this?**
/kind feature

Update github.com/container-storage-interface/spec to v1.12.0, which brings GA support of ControllerModifyVolume.

```release-note
Updates CSI spec to v1.12.0, which brings GA support of ControllerModifyVolume.
```


jsafrane:
cc @xing-yang @msau42

xing-yang:
It looks like we need to update csi-test.

PR #1435, opened 2025-10-30
capacity: fix duplicate topology

huww98:
**What type of PR is this?**

/kind bug

When the controller starts, 2 sync() call will run simultaneously, one from HasSynced(), another from processNextWorkItem(). Each will produce an instance for the same topology segment, and pass it to callbacks.

This will result in duplicated entries in capacities map, resulting in: either
- Two CSIStorageCapacity object get created for the same topology, or
- The same CSIStorageCapacity object get assigned to two keys in capacities map. When one of them is updated, the other one will hold an outdated object and all subsequent update will fail with conflict.

Fixes #

```release-note
Fixed possible duplicated CSIStorageCapacity and constantly failing update request.
```

### DIFF
--- a/pkg/capacity/topology/nodes.go
+++ b/pkg/capacity/topology/nodes.go
@@ -21,6 +21,7 @@ import (
 	"reflect"
 	"sort"
 	"sync"
+	"sync/atomic"
 
 	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
@@ -159,6 +160,7 @@ type nodeTopology struct {
 	nodeInformer    coreinformersv1.NodeInformer
 	csiNodeInformer storageinformersv1.CSINodeInformer
 	queue           workqueue.TypedRateLimitingInterface[string]
+	hasSynced       atomic.Bool
 
 	mutex sync.Mutex
 	// segments hold a list of all currently known topology segments.
@@ -209,13 +211,19 @@ func (nt *nodeTopology) RunWorker(ctx context.Context) {
 	}
 }
 
+func (nt *nodeTopology) upstreamInformersSynced() bool {
+	return nt.nodeInformer.Informer().HasSynced() &&
+		nt.csiNodeInformer.Informer().HasSynced()
+}
+
 func (nt *nodeTopology) HasSynced() bool {
-	if nt.nodeInformer.Informer().HasSynced() &&
-		nt.csiNodeInformer.Informer().HasSynced() {
-		// Now that both informers are up-to-date, use that
-		// information to update our own view of the world.
-		nt.sync(context.Background())
+	if nt.hasSynced.Load() {
 		return true
+	}
+	if nt.upstreamInformersSynced() {
+		// Now that both informers are up-to-date,
+		// trigger a sync to update the list of topology segments.
+		nt.queue.Add("")
 	}
 	return false
 }
@@ -231,6 +239,11 @@ func (nt *nodeTopology) processNextWorkItem(ctx context.Context) bool {
 }
 
 func (nt *nodeTopology) sync(_ context.Context) {
+	if !nt.hasSynced.Load() && nt.upstreamInformersSynced() {
+		// We are not yet synced, but the upstream informers are.
+		// we will become synced when this function returns
+		defer nt.hasSynced.Store(true)
+	}
 	// For all nodes on which the driver is registered, collect the topology key/value pairs
 	// and sort them by key name to make the result deterministic. Skip all segments that have
 	// been seen before.
--- a/pkg/capacity/topology/nodes_test.go
+++ b/pkg/capacity/topology/nodes_test.go
@@ -566,6 +566,41 @@ func TestNodeTopology(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


pohly:
> When the controller starts, 2 sync() call will run simultaneously, one from HasSynced(), another from processNextWorkItem(). Each will produce an instance for the same topology segment, and pass it to callbacks.

New segments get produced in `sync`, right? So whenever two `sync` calls are executed in parallel, we have this problem. I agree that this is faulty. What isn't clear to me is the proposed solution.

Suppose there are two work queue items in the queue at a time when `nt.hasSynced` is still false. Both get processed in parallel. Don't we still have the problem?

Two solutions:
- only run one worker
- serialize the code which generates new segment pointers (not exactly sure though how long the mutex must be held for that)

huww98:
> only run one worker

Yes, I think this controller is designed to only run one worker

https://github.com/kubernetes-csi/external-provisioner/blob/b84b08f87be64ee63e059a71831d35d4569743cb/pkg/capacity/topology/nodes.go#L204-L210

It is not configuable, and we will only start one worker goroutine now.

pkg/capacity/topology/nodes.go
```diff
@@ -209,14 +211,20 @@ func (nt *nodeTopology) RunWorker(ctx context.Context) {
...
 func (nt *nodeTopology) HasSynced() bool {
-	if nt.nodeInformer.Informer().HasSynced() &&
-		nt.csiNodeInformer.Informer().HasSynced() {
-		// Now that both informers are up-to-date, use that
-		// information to update our own view of the world.
-		nt.sync(context.Background())
+	if nt.hasSynced.Load() {
 		return true
 	}
+	if nt.upstreamSynced() {
```
sunnylovestiramisu: Nit: can we rename it to upstreamInformersSynced to be clearer?

pkg/capacity/topology/nodes.go
```diff
@@ -209,14 +211,20 @@ func (nt *nodeTopology) RunWorker(ctx context.Context) {
...
+		// Now that both informers are up-to-date,
+		// trigger a sync to update the list of topology segments.
+		nt.queue.Add("")
```
sunnylovestiramisu, line 226: Nit: I think this is a common way to trigger a sync, but it is kinda difficult to debug with logging? Can the key be something more explict like "full-sync" or "reconcile-all"?

But if it is a common pattern among our sidecars, ignore this comment.

pkg/capacity/topology/nodes.go
huww98: Sure, done.

pohly:
> Yes, I think this controller is designed to only run one worker

`RunWorker` executes one worker, but could be invoked more than once. Where is it called?

huww98:
> > Yes, I think this controller is designed to only run one worker
> 
> `RunWorker` executes one worker, but could be invoked more than once. Where is it called?

https://github.com/kubernetes-csi/external-provisioner/blob/b84b08f87be64ee63e059a71831d35d4569743cb/cmd/csi-provisioner/csi-provisioner.go#L524

Here, only once.

pohly:
It wasn't designed to be run only once, that's just only how it's currently being done. But as that apparently is sufficient, the fix can be pretty simple:
- document that RunWorker must only be called once
- in RunWorker, block waiting for informer sync
- populate one work queue item
- run the for loop

Wouldn't that solve the problem without all of the complicated proposed back-and-forth between event handlers and sync loop?

huww98:
> * in RunWorker, block waiting for informer sync

Then we will lost the ability of syncing partial data from upstream controllers. Not sure if this is good. Given the current implementation, I think incremental sync is not faster than a full sync. But this may delay the first topology being passed to callbacks.

And together with this, I think we need to move `go topologyInformer.RunWorker(ctx)` after we start other informers, or we will poll not-started informers infinitely if we are not the leader.  

> Wouldn't that solve the problem without all of the complicated proposed back-and-forth between event handlers and sync loop?

We still need a `hasSynced atomic.Bool` to tell outside that we have finished at least one loop.
IMO, your proposal can be simpler, but not very much. I can give it a try.

pohly:
> Then we will lost the ability of syncing partial data from upstream controllers. Not sure if this is good.

It's normal that controllers wait for a full cache sync before starting their work. It depends a bit on the controller whether it makes sense to start earlier.

> We still need a hasSynced atomic.Bool to tell outside that we have finished at least one loop.

Is someone checking that? I don't remember.

huww98:
https://github.com/kubernetes-csi/external-provisioner/blob/b84b08f87be64ee63e059a71831d35d4569743cb/pkg/capacity/capacity.go#L276

checked here, via `c.topologyInformer.HasSynced`.

So It seems fine to only sync after upstream synced, because the controller is still waiting for sync.

huww98:
@pohly Please take a look at #1450. Implemented your proposed fix.

pohly:
/remove-lifecycle rotten

I need to check whether all comments were addressed.

pohly:
/close

In favor of https://github.com/kubernetes-csi/external-provisioner/pull/1450.

PR #1438, opened 2025-11-06
[controller] Reschedule provisioning if node is missing

### LINKED ISSUE #1437: PVC stuck in Pending when CSINode is deleted during provionning phase
**What happened**:

I got a PVC stuck in an infinite retry loop during Provisioning which I think is caused by the changes introduced in https://github.com/kubernetes-csi/external-provisioner/pull/1413

The PVC had the following events:
```
Events:
  Type     Reason                Age                    From                                                                                          Message
  ----     ------                ----                   ----                                                                                          -------
  Normal   WaitForFirstConsumer  4h3m (x2 over 4h3m)    persistentvolume-controller                                                                   waiting for first consumer to be created before binding
  Normal   ExternalProvisioning  177m (x263 over 4h3m)  persistentvolume-controller                                                                   Waiting for a volume to be created either by the external provisioner 'ebs.csi.aws.com' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
  Normal   ExternalProvisioning  68m (x421 over 173m)   persistentvolume-controller                                                                   Waiting for a volume to be created either by the external provisioner 'ebs.csi.aws.com' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
  Warning  ProvisioningFailed    64m (x56 over 4h3m)    ebs.csi.aws.com_aws-ebs-csi-controller-7b7f86db84-9t2m6_11663768-627d-419a-93ad-24f794b77af6  error generating accessibility requirements: no topology key found for node ip-10-150-80-32.us-west-2.compute.internal
  Normal   Provisioning          4m34s (x72 over 4h3m)  ebs.csi.aws.com_aws-ebs-csi-controller-7b7f86db84-9t2m6_11663768-627d-419a-93ad-24f794b77af6  External provisioner is provisioning volume for claim "cass-ukv-xpq-multi-step-execution/server-data-ukv-xpq-multi-step-execution-dc1-1c19-ra-sts-0"
  Warning  ProvisioningFailed    4m34s (x16 over 59m)   ebs.csi.aws.com_aws-ebs-csi-controller-7b7f86db84-9t2m6_11663768-627d-419a-93ad-24f794b77af6  error generating accessibility requirements: failed to get selected CSINode ip-10-150-80-32.us-west-2.compute.internal: csinode.storage.k8s.io "ip-10-150-80-32.us-west-2.compute.internal" not found
  Normal   ExternalProvisioning  63s (x261 over 66m)    persistentvolume-controller                                                                   Waiting for a volume to be created either by the external provisioner 'ebs.csi.aws.com' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
```

From the audit logs I gathered the following timeline:
1. a CSI node pod is restarted on a node and the new pod fails to start (reason unrelated to the bug) 

Image

2. at the same time, the CSINode object gets cleared of the driver because the CSI node is unable to register itself  (expected)

Image

3. Multiple hours later, a pod gets scheduled onto the node and the CSI provisioner attempts to provision the PV but fails due to the still missing topology (expected)

Image

4. A few hours later the node object get deleted alongside the CSINode (expected)

Image

5. events for the PVC provisioning become `ebs.csi.aws.com_aws-ebs-csi-controller-7b7f86db84-9t2m6_11663768-627d-419a-93ad-24f794b77af6  error generating accessibility requirements: failed to get selected CSINode ip-10-150-80-32.us-west-2.compute.internal: csinode.storage.k8s.io "ip-10-150-80-32.us-west-2.compute.internal" not found` due to the missing CSINode and they never change (**unexpected**)

6. I recovered the situation by manually deleting the `selected-node` annotation to let the Pod+PVC be rescheduled somewhere else

Once the CSINode and node objects have been deleted and removed from the cluster, I expect the CSI provisioner to recover the situation and delete the `selected-node` annotation by itself like it used to do before [this change](https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/194/files#diff-3c5bb5f48211873c58fcba055dcae2ac7b1958969219e06e1508d76d485dace7L1495-L1498) 

**How to reproduce it**:

I don't have an exact reproducer but I think you can reproduce by:
1. Force a long volume provisioning (for instance by removing the CSI node pod so that it gets deregistered by the kubelet and there is no more topology keys)
2. Schedule a pod with a PVC on the node with the missing topology
3. Delete Node and CSINode objects
4. The PVC should now be stuck until something removes the `selected-node` annotation

**Environment**:
- Driver version: aws-ebs-csi v1.52.0 with csi-provisioner v6.0.0
- Kubernetes version (use `kubectl version`): v1.33
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Fricounet:
**What type of PR is this?**

/kind bug

GenerateAccessibilityRequirements tries to get the Node and CSINode objects but if they are missing (because they were deleted), then the provisioning will fail with ProvisioningNoChange which means that it will potentially be retried forever if the node never comes back because nothing is removing the selected-node annotation anymore. This commit makes it so that Not Found api errors are properly caught and when it's the case, ProvisioningReschedule is returned to tell the scheduler to try a new node. This matches the previous implementation in the external-provisioner lib (https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/194/files#diff-3c5bb5f48211873c58fcba055dcae2ac7b1958969219e06e1508d76d485dace7L1496-L1498)

Fixes #1437 

```release-note
Fixed infinite retry loop during provisioning if node was deleted in the meantime.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -701,7 +701,10 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 			p.csiNodeLister,
 			p.nodeLister,
 			p.pvcNodeStore)
-		if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The node or CSINode object can't be found, ask the scheduler for a reschedule
+			return nil, controller.ProvisioningReschedule, err
+		} else if err != nil {
 			return nil, controller.ProvisioningNoChange, fmt.Errorf("error generating accessibility requirements: %v", err)
 		}
 		req.AccessibilityRequirements = requirements
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -2503,6 +2503,29 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 			expectErr:   true,
 			expectState: controller.ProvisioningFinished,
 		},
+		"fail with selected node but node doesn't exist": {
+			pluginCapabilities: provisionWithTopologyCapabilities,
+			volOpts: controller.ProvisionOptions{
+				SelectedNodeName: nodeBar.Name,
+				StorageClass: &storagev1.StorageClass{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: fakeSCName,
+					},
+					ReclaimPolicy: &deletePolicy,
+					Parameters: map[string]string{
+						"fstype": "ext3",
+					},
+				},
+				PVName: "test-name",
+				PVC: func() *v1.PersistentVolumeClaim {
+					claim := createFakePVC(requestedBytes)
+					claim.Annotations[annSelectedNode] = nodeBar.Name
+					return claim
+				}(),
+			},
+			expectErr:   true,
+			expectState: controller.ProvisioningReschedule,
+		},
 	}
 }
 


pkg/controller/controller.go
```diff
@@ -701,7 +701,10 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 			p.csiNodeLister,
 			p.nodeLister,
 			p.pvcNodeStore)
-		if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The node or CSINode object can't be found, ask the scheduler for a reschedule
+			return nil, controller.ProvisioningReschedule, err
```
sunnylovestiramisu, line 706: Does the GenerateAccessibilityRequirements properly return NotFound error now？

sunnylovestiramisu: Because if the cache still has the csinode info, the GenerateAccessibilityRequirements will not even return a NotFound error.

sunnylovestiramisu:
Did you test it end to end that the fix works as expected?

pkg/controller/controller.go
sunnylovestiramisu: Under what condition the node never comes back will retry forever? Because if it is a long provisioning, the PV object will be eventually created or failed with a final error, so it should not be forever?

sunnylovestiramisu:
I am wondering if you encountered the case that: 
1. CreateVolume is called the first time but the cache of node/csinode is not filled yet 
2. The node is deleted 
3. CreateVolume throws the NotFound error of node or csinode
4. This case it does not count as a final error so it retries forever

pkg/controller/controller_test.go
```diff
@@ -2503,6 +2503,29 @@ func provisionTestcases() (int64, map[string]provisioningTestcase) {
 			expectErr:   true,
 			expectState: controller.ProvisioningFinished,
 		},
+		"fail with selected node but node doesn't exist": {
```
sunnylovestiramisu, line 2506: Does this test fail before you add change to controller.go?

pkg/controller/controller.go
```diff
@@ -701,7 +701,10 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 			p.csiNodeLister,
 			p.nodeLister,
 			p.pvcNodeStore)
-		if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The node or CSINode object can't be found, ask the scheduler for a reschedule
```
msau42, line 705: are there any other paths where NotFound is returned?

pkg/controller/controller_test.go
Fricounet: it does. From the current master branch:
```
$ go test ./pkg/controller -v -test.run "TestProvision/fail_with_selected_node_but_node_doesn't_exist"
=== RUN   TestProvision
=== RUN   TestProvision/fail_with_selected_node_but_node_doesn't_exist
W1112 09:23:06.869182 2120813 controller.go:648] "fstype" is deprecated and will be removed in a future release, please use "csi.storage.k8s.io/fstype" instead
    controller_test.go:2820: expected ProvisioningState Reschedule, got NoChange
--- FAIL: TestProvision (0.01s)
    --- FAIL: TestProvision/fail_with_selected_node_but_node_doesn't_exist (0.01s)
```

pkg/controller/controller.go
Fricounet: I don't think so. Based on the inputs of [GenerateAccessibilityRequirements](https://github.com/kubernetes-csi/external-provisioner/blob/1b71152a5b3463d9c13157c488e54a00e7bedc9c/pkg/controller/topology.go#L168-L179), I see 4 things that could potentially lead to this kind of errors:
- the `kubeClient` if it's used to make any kind of api request. But looking at the code, we can see that this variable is actually completely unused (maybe we should consider removing it?)
- the `pvcNodeStore` object though the [TopologyProvider](https://github.com/kubernetes-csi/external-provisioner/blob/1b71152a5b3463d9c13157c488e54a00e7bedc9c/pkg/controller/cache.go#L69) implementation doesn't return an apierrors.NotFound error
- the `csiNodeLister` and the `nodeLister` are the two remaining ones and will return a proper apierror.NotFound error when the object is not present in the informer which is what we want

sunnylovestiramisu:
Should we also cherry-pick the change v6.x.x?

Fricounet:
up to you, I don't particularly need this to be backported on my end

PR #1448, opened 2025-11-25
Allow provisioning to proceed to prevent leaking resources

### LINKED ISSUE #1449: Leaking resources when pending PVC is deleted while being created from snapshot that is also being deleted
**What happened**:
Volume is leaked when Creating PVC from VolumeSnapshot times out and the user starts to delete PVC and delete VolumeSnapshot.

Create volume from snapshot times out, user tries to delete PVC, and delete VolumeSnapshot. CSI driver returns non-final error code. However, provisioner never tries to delete the volume when it is finally created.

In the external-provisioner logs, I see “I1125 17:27:23.655056       1 controller.go:1082] Temporary error received, adding PVC 2844879f-0f2f-4d6d-bf0d-e02ae676b8e7 to claims in progress”. So after the CSI driver returns a non-final error, the PVC is added to “Claims in Progress” so that it will be retried.

However it failed because the VolumeSnapshot DataSource is being deleted: “E1125 17:27:23.666986       1 controller.go:957] error syncing claim "2844879f-0f2f-4d6d-bf0d-e02ae676b8e7": failed to provision volume with StorageClass "wcpglobal-storage-profile": error getting handle for DataSource Type VolumeSnapshot by Name test-snapshot: snapshot test-snapshot is currently being deleted”.

As a result, CSI driver never gets a call to retry. Volume is finally getting created but it is leaked because there isn't a PVC/PV that points to it.

Although VolumeSnapshot has a deletionTimestamp, it has a finalizer that prevents it from being deleted when it is used as a source to create PVC. The finalizer is eventually removed when the volume is created.  It takes a while but VolumeSnapshot will be deleted eventually.

Note: If I only delete the pending PVC, not deleting the VolumeSnapshot, the volume will be deleted after it is being created successfully. However, if creating PVC from VolumeSnapshot takes a long time, it is typical for a backup application to delete PVC and then also delete VolumeSnapshot.

Volume and VolumeSnapshot should be deleted.

The external-provisioner should continue to retry provisioning the PVC by calling the CSI driver even if VolumeSnapshot is being deleted. CSI driver should be idempotent and it will handle retries.

**How to reproduce it**:

**Environment**:
- Driver version:
- Kubernetes version (use `kubectl version`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

xing-yang:
**What type of PR is this?**
/kind bug

The external-provisioner should continue to retry provisioning the PVC by calling the CSI driver even if VolumeSnapshot is being deleted. CSI driver should be idempotent and it will handle retries.

Fixes #1449

```release-note
Allow provisioning to proceed when snapshot is being deleted to prevent leaking volumes and snapshots.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -150,7 +150,8 @@ const (
 
 	snapshotNotBound = "snapshot %s not bound"
 
-	pvcCloneFinalizer = "provisioner.storage.kubernetes.io/cloning-protection"
+	pvcCloneFinalizer                 = "provisioner.storage.kubernetes.io/cloning-protection"
+	snapshotSourceProtectionFinalizer = "snapshot.storage.kubernetes.io/volumesnapshot-as-source-protection"
 
 	annAllowVolumeModeChange = "snapshot.storage.kubernetes.io/allow-volume-mode-change"
 )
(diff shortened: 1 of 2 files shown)


jsafrane:
I can see it makes sense to _continue_ provisioning when it started with a valid snapshot. But does it make sense to _start_ it with a snapshot that is being deleted? Do we have a good way how to distinguish those two?

For volume cloning, the provisioner adds a finalizer to PVC that is being cloned. Should it add a similar finalizer to VolumeSnapshot that is being restored? We could then easily see if provisioning started before a VolumeSnapshot got DeletionTimestamp:
* If the finalizer exists, it must have been added *before* the snapshot got DeletionTimestamp (the API server checks that) and therefore the provisioning started and the provisioner should continue. Once it succeeds, remove the finalizer.
* If the finalizer does not exist, try to add it. The API server will refuse to add a new finalizer to an object with DeletionTimestamp. Provisioning did not start, reject it with an error.

xing-yang:
That's a good point. We do have a finalizer added to VolumeSnapshot being used as a source to create PVC. I can add a check for that to differentiate an already started provisioning from a new provisioning case.

pkg/controller/controller.go
```diff
@@ -150,7 +150,8 @@ const (
 
 	snapshotNotBound = "snapshot %s not bound"
 
-	pvcCloneFinalizer = "provisioner.storage.kubernetes.io/cloning-protection"
+	pvcCloneFinalizer                 = "provisioner.storage.kubernetes.io/cloning-protection"
```
sunnylovestiramisu, line 153: Did we discuss the approach between finalizer and in memory cache or persistent storage?

xing-yang: External-provisioner already keeps retrying PVC provisioning. See logs below.
```
I1125 17:27:23.655056 1 controller.go:1082] Temporary error received, adding PVC 2844879f-0f2f-4d6d-bf0d-e02ae676b8e7 to claims in progress”.
```
Here the CSI driver returns a non-final error, the PVC is added to “Claims in Progress” to be retried even if PVC itself is deleted.

This check for finalizer is just to differentiate between a new provisioning request vs a provisioning retry. If it is a new provisioning request (the finalizer is not added yet), the external provisioner will fail the request and won't call the CSI driver. If it is a retry (finalizer is already added), the external provisioner will call CSI driver to retry. This allows the CSI driver to complete the volume provisioning and return success. After that, the volume will be deleted if the deletionTimestamp was added to PVC earlier.

PR #1450, opened 2025-11-26
capacity: fix duplicate topology (attempt #2)

huww98:
When the controller starts, 2 sync() call will run simultaneously, one from HasSynced(), another from processNextWorkItem(). Each will produce an instance for the same topology segment, and pass it to callbacks.

This will result in duplicated entries in capacities map, resulting in: either
- Two CSIStorageCapacity object get created for the same topology, or
- The same CSIStorageCapacity object get assigned to two keys in capacities map. When one of them is updated, the other one will hold an outdated object and all subsequent update will fail with conflict.

**What type of PR is this?**

/kind bug

>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

Please see also #1435

```release-note
Fixed possible duplicated CSIStorageCapacity and constantly failing update request.
```

/cc @pohly

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -449,6 +449,7 @@ func main() {
 	)
 
 	var capacityController *capacity.Controller
+	var topologyInformer topology.Informer
 	if *enableCapacity {
 		// Publishing storage capacity information uses its own client
 		// with separate rate limiting.
@@ -483,7 +484,6 @@ func main() {
 			klog.Infof("using %s/%s %s as owner of CSIStorageCapacity objects", controller.APIVersion, controller.Kind, controller.Name)
 		}
 
-		var topologyInformer topology.Informer
 		if nodeDeployment == nil {
 			topologyRateLimiter := workqueue.NewTypedItemExponentialFailureRateLimiter[string](*retryIntervalStart, *retryIntervalMax)
 			topologyInformer = topology.NewNodeTopology(
@@ -503,7 +503,6 @@ func main() {
 			klog.Infof("producing CSIStorageCapacity objects with fixed topology segment %s", segment)
 			topologyInformer = topology.NewFixedNodeTopology(&segment)
 		}
-		go topologyInformer.RunWorker(ctx)
 
 		managedByID := "external-provisioner"
 		if *enableNodeDeployment {
@@ -662,9 +661,12 @@ func main() {
 
 		factory.Start(ctx.Done())
 		if factoryForNamespace != nil {
-			// Starting is enough, the capacity controller will
+			// Starting is enough, the capacityController and topologyInformer will
 			// wait for sync.
 			factoryForNamespace.Start(ctx.Done())
+		}
+		if topologyInformer != nil {
+			go topologyInformer.RunWorker(ctx)
 		}
 		cacheSyncResult := factory.WaitForCacheSync(ctx.Done())
 		for _, v := range cacheSyncResult {
--- a/pkg/capacity/topology/nodes.go
+++ b/pkg/capacity/topology/nodes.go
@@ -21,6 +21,7 @@ import (
 	"reflect"
 	"sort"
 	"sync"
+	"sync/atomic"
 
 	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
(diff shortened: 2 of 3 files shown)


pkg/capacity/topology/nodes_test.go
```diff
@@ -566,6 +567,39 @@ func TestNodeTopology(t *testing.T) {
 	}
 }
 
+func TestHasSynced(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		client := fakeclientset.NewSimpleClientset()
+		informerFactory := informers.NewSharedInformerFactory(client, 1*time.Hour)
```
huww98: For why not setting resync period to 0, see https://github.com/kubernetes/kubernetes/pull/133500

pkg/capacity/topology/nodes.go
```diff
@@ -205,19 +207,22 @@ func (nt *nodeTopology) RunWorker(ctx context.Context) {
...
 
+	if !cache.WaitForCacheSync(ctx.Done(),
+		nt.nodeInformer.Informer().HasSynced, nt.csiNodeInformer.Informer().HasSynced) {
+		return
+	}
+
+	go func() {
+		<-ctx.Done()
+		nt.queue.ShutDown()
+	}()
```
huww98, line 224: synctest checks for leaked goroutines. So I have to clean it up.

huww98:
We need go 1.25 to use the synctest package :(

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -680,10 +679,13 @@ func main() {
 
 		factory.Start(ctx.Done())
 		if factoryForNamespace != nil {
-			// Starting is enough, the capacity controller will
+			// Starting is enough, the capacityController and topologyInformer will
 			// wait for sync.
 			factoryForNamespace.Start(ctx.Done())
 		}
+		if topologyInformer != nil {
+			go topologyInformer.RunWorker(ctx)
```
pohly, line 669: Please add documentation which explains that RunWorker may only be called once per instance and why.

pkg/capacity/topology/nodes_test.go
pohly: I don't see the connection to https://github.com/kubernetes/kubernetes/pull/133500. Can you explain here why a resync period of one hour is useful?

pohly: I'm guessing it's because without https://github.com/kubernetes/kubernetes/pull/133500, a zero resync period would cause reads from `neverExitWatch`, which isn't valid in a synctest. But it would be nice to explicitly state that.

pohly:
/remove-lifecycle rotten

@huww98: shall we finish this PR?

See:
- https://github.com/kubernetes-csi/external-provisioner/pull/1450#discussion_r2566099389
- https://github.com/kubernetes-csi/external-provisioner/pull/1450#discussion_r2619056309

huww98:
Yes, I was waiting for go 1.25 update for synctest. Now we have it, let me rebase and resolve the comments.

pkg/capacity/topology/nodes_test.go
huww98: Now we have Kubernetes v0.35 included in the go.mod. We actually can set resync period to 0. So no need special comment now.

cmd/csi-provisioner/csi-provisioner.go
huww98: Added a comment on `nodeTopology.RunWorker`. PTAL again!

cmd/csi-provisioner/csi-provisioner.go
```diff
@@ -662,10 +661,13 @@ func main() {
 
 		factory.Start(ctx.Done())
 		if factoryForNamespace != nil {
-			// Starting is enough, the capacity controller will
+			// Starting is enough, the capacityController and topologyInformer will
 			// wait for sync.
 			factoryForNamespace.Start(ctx.Done())
 		}
+		if topologyInformer != nil {
```
pohly, line 668: Why is important that `topologyInformer` gets started here vs. where it was started before?

If it's important, then let's add a comment explainining why. If it's not important, then let's not change it.

huww98: All other informers are started here. It is nature to start it here.
I think we should only start `topologyInformer` after `factoryForNamespace.Start`, to avoid `cache.WaitForCacheSync` waiting for not-started informers. Wasting a little CPU when waiting for leader election.

There is already a comment just before:
```go
			// Starting is enough, the capacityController and topologyInformer will
			// wait for sync.
```
Do you think this is enough?

pohly: I think both approaches would be fine and in general I prefer to avoid drive-by enhancements that aren't related to what a PR is primarily trying to do (in this case "fix duplicate topology"). It causes churn and makes reviews harder.

But we can keep it.

pohly (commented):
/lgtm
/approve

PR #1458, opened 2025-12-13
Add a finalizer on VolumeSnapshot as Source

xing-yang:
**What type of PR is this?**
/kind bug

In this PR https://github.com/kubernetes-csi/external-provisioner/pull/1448, we added a check for this finalizer "snapshot.storage.kubernetes.io/volumesnapshot-as-source-protection" which is added and removed by external-snapshotter to prevent a VolumeSnapshot from being deleted while it is being used as a source to create PVC. However, there is a race condition. When a user deletes a pending PVC, the PVC is removed from the informer cache by [the external-snapshotter logic](https://github.com/kubernetes-csi/external-snapshotter/blob/7dd22e0a905756e095d2989591d40cc2f44a9e1d/pkg/common-controller/snapshot_controller.go#L951), even though the volume is still being created from snapshot in the backend and the external-provsioner is still processing it. The external-snapshotter has no knowledge of whether the provisioner is still processing the volume. So it is not in the best position to handle the life cycle of this kind of finalizer.

In this PR, we add a new finalizer "provisioner.storage.kubernetes.io/volumesnapshot-as-source-protection". It is added and removed by external-provisioner.
* When Added: The provisioner.storage.kubernetes.io/volumesnapshot-as-source-protection finalizer is added to the VolumeSnapshot when the provisioner starts provisioning a PVC from that snapshot (before calling the CSI driver's CreateVolume).
* When Removed: The finalizer is removed when the provisioner completes provisioning (either successfully creating the PV or encountering a final error), as indicated by returning ProvisioningFinished state from the Provision method.

The original finalizer managed by external-snapshotter is deprecated and will be removed in a future release. When removing the old finalizer, we should add logic in external-snapshotter to always remove it. This way user won't get stuck with a finalizer not being removed by any component.

Fixes #

```release-note
Add provisioner.storage.kubernetes.io/volumesnapshot-as-source-protection finalizer on VolumeSnapshot as Source. Add rbac rules to watch/update volumesnapshots.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1071,14 +1071,22 @@ func (p *csiProvisioner) setSnapshotFinalizer(ctx context.Context, dataSource *v
 	}
 
 	snapshotClone := snapshot.DeepCopy()
-	if !checkFinalizer(snapshotClone, snapshotSourceProtectionFinalizer) {
-		snapshotClone.Finalizers = append(snapshotClone.Finalizers, snapshotSourceProtectionFinalizer)
-		_, err := p.snapshotClient.SnapshotV1().VolumeSnapshots(snapshotClone.Namespace).Update(ctx, snapshotClone, metav1.UpdateOptions{})
-		if err != nil {
-			return err
-		}
-		klog.V(3).Infof("Added finalizer %s to snapshot %s/%s", snapshotSourceProtectionFinalizer, snapshotClone.Namespace, snapshotClone.Name)
+	if checkFinalizer(snapshotClone, snapshotSourceProtectionFinalizer) {
+		return nil
 	}
+
+	snapshotClone.Finalizers = append(snapshotClone.Finalizers, snapshotSourceProtectionFinalizer)
+	_, err = p.snapshotClient.SnapshotV1().VolumeSnapshots(snapshotClone.Namespace).Update(ctx, snapshotClone, metav1.UpdateOptions{})
+	if err != nil {
+		// If we don't have permission to update VolumeSnapshots, log at info level and continue.
+		// This allows the provisioner to work without the new RBAC permissions for backwards compatibility.
+		if apierrors.IsForbidden(err) {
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -1045,6 +1064,75 @@ func (p *csiProvisioner) setCloneFinalizer(ctx context.Context, dataSource *v1.O
...
 }
 
+func (p *csiProvisioner) setSnapshotFinalizer(ctx context.Context, dataSource *v1.ObjectReference) error {
+	snapshot, err := p.snapshotClient.SnapshotV1().VolumeSnapshots(dataSource.Namespace).Get(ctx, dataSource.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	snapshotClone := snapshot.DeepCopy()
+	if !checkFinalizer(snapshotClone, snapshotSourceProtectionFinalizer) {
```
jsafrane: Reduce nesting and return early:
```go
if checkFinalizer(snapshotClone, snapshotSourceProtectionFinalizer) {
    return
}
```

pkg/controller/controller.go
```diff
@@ -1045,6 +1064,75 @@ func (p *csiProvisioner) setCloneFinalizer(ctx context.Context, dataSource *v1.O
...
+		snapshotClone.Finalizers = append(snapshotClone.Finalizers, snapshotSourceProtectionFinalizer)
+		_, err := p.snapshotClient.SnapshotV1().VolumeSnapshots(snapshotClone.Namespace).Update(ctx, snapshotClone, metav1.UpdateOptions{})
+		if err != nil {
+			// If we don't have permission to update VolumeSnapshots, log a warning and continue.
+			// This allows the provisioner to work without the new RBAC permissions for backwards compatibility.
+			if apierrors.IsForbidden(err) {
+				klog.Warningf("Unable to add finalizer to snapshot %s/%s due to missing RBAC permissions (needs 'update' on volumesnapshots): %v. Provisioning will continue without snapshot protection. Please update RBAC to include 'update' verb for volumesnapshots.", snapshotClone.Namespace, snapshotClone.Name, err)
```
jsafrane: This may spam the log. It is OK in `master` branch, where we want people to update the RBACs. But I'd log it at lower level if we're going to backport it.

pkg/controller/controller.go

pkg/controller/controller.go

xing-yang:
Comments are addressed. Thanks.

PR #1481, opened 2026-03-02
Support added for architecture riscv64 (linux)

Opvolger:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

Now that riscv64 support is available for the distroless Docker image, it's possible to add support for riscv64.

See pull request: GoogleContainerTools/distroless#2001

I tested this combination with my riscv64 Kubernetes cluster (k0s with smb-provider). See: https://www.youtube.com/watch?v=WERPNNyzqqs

Riscv64 support can now be slowly added to Docker images that relied on "gcr.io/distroless/static:latest".

Fixes #

```release-note
Support added for architecture riscv64 (linux)
```

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -2,6 +2,7 @@
 
 CSI_PROW_SIDECAR_E2E_IMPORT_PATH="github.com/kubernetes-csi/external-provisioner/v5/test/e2e"
 CSI_PROW_SIDECAR_E2E_PATH="github.com/kubernetes-csi/external-provisioner/test/e2e"
+CSI_PROW_BUILD_PLATFORMS="linux amd64 amd64; linux ppc64le ppc64le -ppc64le; linux s390x s390x -s390x; linux arm arm -arm; linux arm64 arm64 -arm64; linux arm arm/v7 -armv7; linux riscv64 riscv64 -riscv64; windows amd64 amd64 .exe nanoserver:1809 servercore:ltsc2019; windows amd64 amd64 .exe nanoserver:ltsc2022 servercore:ltsc2022"
 
 . release-tools/prow.sh
 


.prow.sh
```diff
@@ -2,6 +2,7 @@
 
 CSI_PROW_SIDECAR_E2E_IMPORT_PATH="github.com/kubernetes-csi/external-provisioner/v5/test/e2e"
 CSI_PROW_SIDECAR_E2E_PATH="github.com/kubernetes-csi/external-provisioner/test/e2e"
+CSI_PROW_BUILD_PLATFORMS="linux amd64 amd64; linux ppc64le ppc64le -ppc64le; linux s390x s390x -s390x; linux arm arm -arm; linux arm64 arm64 -arm64; linux arm arm/v7 -armv7; linux riscv64 riscv64 -riscv64; windows amd64 amd64 .exe nanoserver:1809 servercore:ltsc2019; windows amd64 amd64 .exe nanoserver:ltsc2022 servercore:ltsc2022"
```
jsafrane, line 5: This belongs into https://github.com/kubernetes-csi/csi-release-tools/blob/de06a09a7a68bc8da0a275094ade1f5ea7a1a995/prow.sh#L81. As result, all CSI sidecars will be built for riscv64 (once the release-tools are propagated to all repos, which happens with every golang bump for CVEs...)

Still, the question is, how are we going to support this platform? In the CSI community, we have a lot of Linux + amd64 expertise, with some arm / arm64 when needed. We have no riscv64 to test with.

Opvolger: The problem with all the build is, that the debian base image (used in some CSI projects) has no riscv64 version (yet). See https://github.com/kubernetes/release/pull/4303 and https://github.com/kubernetes/kubernetes/pull/116686

PR #1497, opened 2026-04-14
Bump suffix to v6 in external provisioner

chimanjain:
Bump suffix to v6 in external provisioner as it is still pointing to `v5`.

Because of it, in the `pkg.go.dev`, the latest version shown is still `v5.3.0`.
externa-provisionerv5

```release-note
Updated the package import path to github.com/kubernetes-csi/external-provisioner/v6
```


jsafrane:
/ok-to-test
@chimanjain please sign the CLA, we cannot merge without it

Issue #1507, opened 2026-05-08
PVCs based on snapshots can get stuck deleting due to `volumesnapshot-as-source-protection` finalizer


ConnorJC3:
**What happened**:

The `provisioner.storage.kubernetes.io/volumesnapshot-as-source-protection` finalizer added by the external-provisioner to VolumeSnapshots during provisioning-from-snapshot can become permanently orphaned, preventing the VolumeSnapshot (and potentially its namespace) from being deleted.

The finalizer is added at the start of provisioning and removed on success, but there is no mechanism to remove it if provisioning fails permanently or the provisioner restarts at an inopportune time. Specifically:

1. The provisioner adds the finalizer to the source VolumeSnapshot when a PVC with `dataSource` referencing the snapshot begins provisioning.
2. If provisioning fails with a non-recoverable error (e.g., the PVC is deleted mid-provisioning, the volume is deleted externally, or the StorageClass is removed), `removeSnapshotFinalizer` is never called because it only exists on the success path.
3. If the provisioner crashes or restarts between `setSnapshotFinalizer` and the completion of provisioning, the finalizer is never removed.
4. The VolumeSnapshot remains stuck in `Terminating` state indefinitely if deletion is attempted.

This also can happen if the sidecar fails to remove the finalizer. Currently, the `external-provisioner` only attempts removal a single time. If this removal fails for any reason (e.g. temporary API server failure, rate limit, networking outage, etc.) the finalizer will become stuck forever.

When this happens, the PVC will get stuck deleting due to the finalizer forever. The only current workaround is to manually patch out the finalizer with `kubectl edit` or `kubectl patch`.

**What you expected to happen**:

Kubernetes controllers like the `external-provisioner` should reconcile the existing state on the cluster. In specific, the finalizer should be removed asyncronously when it incorrectly gets stuck on a PVC or otherwise fails to remove.

The existing code has a comment that claims the finalizer "will be cleaned up later" if it fails to remove, but that isn't actually true today:
https://github.com/kubernetes-csi/external-provisioner/blob/867c1dcf5fc69bb643fd791b3d0abdfa22214ada/pkg/controller/controller.go#L1045-L1048

**How to reproduce it**:

1. Create a VolumeSnapshot from an existing PVC
2. Create a new PVC with `dataSource` referencing that VolumeSnapshot
[...]

PR #1512, opened 2026-05-20
Remove finalizer from volume snapshot in error cases during volume provision workflow

vdkotkar:
**What type of PR is this?**
/kind bug

If we create a volume from VolumeSnapshot and CreateVolume returns error, then finalizer from VolumeSnapshot won't be removed and snapshot will not be deleted. Fixed this issue by removing finalizer from volume snapshot in error cases during volume provision workflow.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -867,6 +867,13 @@ func (p *csiProvisioner) Provision(ctx context.Context, options controller.Provi
 
 	result, state, err := p.prepareProvision(ctx, claim, options.StorageClass, options.SelectedNodeName)
 	if result == nil {
+		// prepareProvision may have added the snapshot finalizer before
+		// hitting a subsequent error. Clean it up when provisioning
+		// will not be retried (ProvisioningFinished / ProvisioningReschedule).
+		if state != controller.ProvisioningNoChange &&
+			state != controller.ProvisioningInBackground {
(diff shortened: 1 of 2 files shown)


xing-yang:
Can you fix the unit test failure?

vdkotkar:
> Can you fix the unit test failure?

Fixed formatting error.

jsafrane:
> If we create a volume from VolumeSnapshot and CreateVolume returns error, then finalizer from VolumeSnapshot won't be removed and snapshot will not be deleted

That's only a subset of issues we have with the snapshot finalizer. See https://github.com/kubernetes-csi/external-provisioner/issues/1507 for more of them.

Especially the provisioner restart cannot be fixed by random finalizer removals in the `Provision()` path. `Provision()` may not be called after restart at all, yet some snapshots may need their finalizers removed, because a PVC was deleted while the provisioner was down.

I think it would be more reliable to add a separate controller that removes the finalizer from snapshots that are not referenced by any PVC or the PVC was bound (i.e. `spec.persistentVolumeName` is not empty). It could be very similar to `CloningProtectionController`.

xing-yang:
Here's a fix from @ConnorJC3 that adds a SnapshotFinalizerController: https://github.com/kubernetes-csi/external-provisioner/pull/1519

We can close this PR.  Thanks @vdkotkar for the effort.

PR #1519, opened 2026-05-28
Add SnapshotFinalizerController to clean up orphaned snapshot source-protection finalizers

### LINKED ISSUE #1507: PVCs based on snapshots can get stuck deleting due to `volumesnapshot-as-source-protection` finalizer

ConnorJC3:
**What type of PR is this?**

/kind feature

Adds a `SnapshotFinalizerController` that periodically sweeps all VolumeSnapshots and removes orphaned `provisioner.storage.kubernetes.io/volumesnapshot-as-source-protection` finalizers when no Pending PVC references the snapshot.

The inline `removeSnapshotFinalizer` in the provisioning path handles the happy case, but the finalizer can become permanently orphaned in several scenarios:
- Provisioner crashes between setting and removing the finalizer
- Duplicate provisioning race where the second attempt fails permanently
- PVC is deleted while provisioning is still in progress

An orphaned finalizer blocks snapshot deletion indefinitely. This controller ensures we actively reconcile broken states by sweeping a local VolumeSnapshot informer cache every 5 minutes and cleaning up any finalizer where no Pending PVC still references the snapshot.

The controller is only started when the driver reports `CREATE_DELETE_SNAPSHOT` capability, and gracefully handles missing RBAC permissions (the `list`/`watch`/`update` verbs on volumesnapshots remain optional for backwards compatibility).

Fixes #1507

- I've tested this with both the EBS CSI Driver and hostpath CSI Driver.
- The sweep uses an in-memory informer cache . I'm not married to this implementation and could change from informer to live `List` calls but I figured this solution was more scalable.
- The snapshot sweep controller is enabled by default, configurable with the flag `--snapshot-orphan-sweep-interval` (zero disables the sweeper). Default is every 5 minutes which seems to be standard in other CSI sidecars.

```release-note
Add controller to clean up orphaned snapshot source-protection finalizers that can block snapshot deletion after provisioner crashes or PVC deletions.
```

### DIFF
--- a/pkg/controller/snapshot_finalizer_controller.go
+++ b/pkg/controller/snapshot_finalizer_controller.go
@@ -1,0 +1,234 @@
+package controller
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/kubernetes-csi/csi-lib-utils/rpc"
+	"github.com/kubernetes-csi/external-provisioner/v6/pkg/features"
+	v1 "k8s.io/api/core/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/watch"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	crdv1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
+	snapclientset "github.com/kubernetes-csi/external-snapshotter/client/v8/clientset/versioned"
+)
+
+// SnapshotFinalizerController periodically sweeps all VolumeSnapshots (via a
+// local informer cache) looking for the provisioner's source-protection
+// finalizer. When it finds one, it checks whether any Pending PVC still
+// references that snapshot. If not, the finalizer is orphaned and is removed.
+//
+// This handles the case where the finalizer becomes orphaned due to:
+//   - Provisioner crash between setting and removing the finalizer
+//   - Duplicate provisioning race where the second attempt fails permanently
+//   - PVC deletion while provisioning is still in progress
+type SnapshotFinalizerController struct {
+	snapshotClient   snapclientset.Interface
+	snapshotInformer cache.SharedInformer
+	claimLister      corelisters.PersistentVolumeClaimLister
+
+	resyncInterval time.Duration
+}
+
+// NewSnapshotFinalizerController creates a new controller for removing orphaned
+// snapshot source-protection finalizers after provisioning completes or fails.
+// Returns nil if the snapshot client is nil or the driver lacks snapshot capability.
+func NewSnapshotFinalizerController(
+	snapshotClient snapclientset.Interface,
+	claimLister corelisters.PersistentVolumeClaimLister,
+	controllerCapabilities rpc.ControllerCapabilitySet,
+	sweepInterval time.Duration,
+) *SnapshotFinalizerController {
+	if snapshotClient == nil {
+		return nil
+	}
+	if !controllerCapabilities[csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT] {
+		return nil
+	}
+
+	snapshotInformer := cache.NewSharedInformer(
+		&cache.ListWatch{
+			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+				return snapshotClient.SnapshotV1().VolumeSnapshots("").List(context.Background(), options)
+			},
+			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+				return snapshotClient.SnapshotV1().VolumeSnapshots("").Watch(context.Background(), options)
+			},
+		},
+		&crdv1.VolumeSnapshot{},
+		10*time.Minute,
+	)
+
+	return &SnapshotFinalizerController{
+		snapshotClient:   snapshotClient,
+		snapshotInformer: snapshotInformer,
+		claimLister:      claimLister,
+		resyncInterval:   sweepInterval,
+	}
+}
+
+// Run starts the controller's periodic sweep loop.
+func (c *SnapshotFinalizerController) Run(ctx context.Context, wg *sync.WaitGroup) {
+	klog.Info("Starting SnapshotFinalizerProtection controller")
+	defer utilruntime.HandleCrash()
+
+	// Preflight: verify we have list/watch permissions on VolumeSnapshots.
+	// These are optional RBAC verbs, so exit gracefully if forbidden.
+	if _, err := c.snapshotClient.SnapshotV1().VolumeSnapshots("").List(ctx, metav1.ListOptions{Limit: 1}); err != nil {
+		if apierrs.IsForbidden(err) {
+			klog.V(3).Infof("SnapshotFinalizerProtection: disabled, missing list permission on volumesnapshots: %v", err)
+			return
+		}
+	}
+
+	go c.snapshotInformer.Run(ctx.Done())
+
+	if !cache.WaitForCacheSync(ctx.Done(), c.snapshotInformer.HasSynced) {
+		klog.Error("SnapshotFinalizerProtection: failed to sync informer caches")
+		return
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/controller/snapshot_finalizer_controller.go
```diff
@@ -0,0 +1,207 @@
...
+		if f != snapshotSourceProtectionFinalizer {
+			finalizers = append(finalizers, f)
+		}
+	}
+
+	clone := snapshot.DeepCopy()
+	clone.Finalizers = finalizers
+	if _, err := c.snapshotClient.SnapshotV1().VolumeSnapshots(namespace).Update(ctx, clone, metav1.UpdateOptions{}); err != nil {
+		if apierrs.IsNotFound(err) || apierrs.IsConflict(err) {
+			return
```
jsafrane, line 192: `V(4)` log could be useful.

pkg/controller/snapshot_finalizer_controller.go
```diff
@@ -0,0 +1,207 @@
...
+// and checks whether any still carry the provisioner's finalizer without a
+// corresponding Pending PVC. This catches cases where PVC deletion events were
+// missed (e.g., controller was not running when the PVC was deleted).
+func (c *SnapshotFinalizerController) sweepOrphanedFinalizers(ctx context.Context) {
+	snapshots := c.snapshotInformer.GetStore().List()
+
+	// List all PVCs cluster-wide once from the in-memory cache. This covers
+	// cross-namespace DataSourceRef references where the PVC and snapshot are
+	// in different namespaces.
+	allPVCs, err := c.claimLister.List(labels.Everything())
```
jsafrane: I am a bit afraid about checking all snapshots with a finalizer with all PVCs. Can we pre-filter PVCs that are restoring a snapshot and are Pending? Or even add an indexer to `claimLister` to list unbound PVCs that restore a snapshot in a given namespace quickly?

pkg/controller/snapshot_finalizer_controller.go
```diff
@@ -0,0 +1,207 @@
...
+
+func isSnapshotSourcePVC(pvc *v1.PersistentVolumeClaim) bool {
+	if pvc.Spec.DataSource != nil && pvc.Spec.DataSource.Kind == snapshotKind {
+		return true
+	}
+	if pvc.Spec.DataSourceRef != nil && pvc.Spec.DataSourceRef.Kind == snapshotKind {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
ConnorJC3: Fixed in next rev

pkg/controller/snapshot_finalizer_controller.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	if pvc.Spec.DataSourceRef != nil && pvc.Spec.DataSourceRef.Kind == snapshotKind {
+		ns := pvc.Namespace
+		if pvc.Spec.DataSourceRef.Namespace != nil && len(*pvc.Spec.DataSourceRef.Namespace) > 0 {
+			ns = *pvc.Spec.DataSourceRef.Namespace
+		}
+		return ns, pvc.Spec.DataSourceRef.Name
+	}
+	return "", ""
+}
```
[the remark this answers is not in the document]
ConnorJC3: Ditto above, also fixed

pkg/controller/snapshot_finalizer_controller.go
```diff
@@ -0,0 +1,207 @@
...
+			wait.Until(func() { c.sweepOrphanedFinalizers(ctx) }, c.resyncInterval, ctx.Done())
+		}()
+	} else {
+		go wait.Until(func() { c.sweepOrphanedFinalizers(ctx) }, c.resyncInterval, ctx.Done())
+	}
+
+	klog.Info("Started SnapshotFinalizerProtection controller")
+	<-ctx.Done()
+	klog.Info("Shutting down SnapshotFinalizerProtection controller")
+}
```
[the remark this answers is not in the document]
ConnorJC3: I've gone ahead and added a preflight check but I am a bit worried about the case where the RBAC permissions are added after startup due to some weird deployment ordering by the user. So if anybody has a better idea I'm all ears

pkg/controller/snapshot_finalizer_controller.go
ConnorJC3: I decided to just build a map, to pre-filter using, because adding an index has to be done before the informer starts, which happens in `main()`, so it would be a lot more complicated to pre-know whether or not it was needed. That said if we strongly prefer that I can look into it.

ConnorJC3:
Ok, all above feedback should hopefully be addressed now :)

pkg/controller/snapshot_finalizer_controller.go
```diff
@@ -0,0 +1,233 @@
...
+// that are currently referenced by at least one PVC whose provisioning has not
+// yet completed (Phase is Pending or empty).
+func (c *SnapshotFinalizerController) buildPendingSnapshotIndex() map[string]bool {
+	allPVCs, err := c.claimLister.List(labels.Everything())
+	if err != nil {
+		klog.V(3).Infof("SnapshotFinalizerProtection: failed to list PVCs: %v", err)
+		return nil
+	}
+
+	pending := make(map[string]bool)
```
jsafrane: I know I am annoying, but if the only thing we need from `pending` is a bool flag, then a `Set` would have a better semantics. 
```suggestion
	pending := sets.New[string]()
```

And 

```
func (c *SnapshotFinalizerController) buildPendingSnapshotIndex() sets.Set[string] {
```
etc.

ConnorJC3: All good I honestly forgot we even had a set library in `k8s.io`, new rev coming shortly after I retest it :)

ConnorJC3:
Should be ready for re-review again now

PR #1531, opened 2026-06-12
Add namespace label/annotation values to CreateVolume parameters

eopeter:
## **What type of PR is this?**
/kind feature

## **What this PR does / why we need it**:
Adds two flags, `--extra-create-metadata-namespace-labels` and `--extra-create-metadata-namespace-annotations`. When either is set, the provisioner fetches the PVC's namespace and forwards the configured label / annotation values to the driver on `CreateVolume` requests as parameters, under the `csi.storage.k8s.io/namespace/labels/<key>` and `csi.storage.k8s.io/namespace/annotations/<key>` prefixes respectively.

This lets CSI drivers consume namespace metadata (for example a tenant or pool identifier recorded on the namespace) without needing namespace read access of their own — the provisioner already runs with the necessary scope. Only the explicitly configured keys are forwarded, and only when present on the namespace.

Both flags default off and operate independently of `--extra-create-metadata`.

### **Usage example**:
A platform records a quota-pool identifier on each namespace and wants the storage backend to account new volumes against that pool at provision time, without granting the CSI driver cluster-wide namespace read access.

  Configure the provisioner to forward the relevant keys:
  
  ```yaml
  # csi-provisioner container args
  - --extra-create-metadata-namespace-annotations=example.com/quota-pool-id
  - --extra-create-metadata-namespace-labels=example.com/tier
```

Grant the provisioner's ServiceAccount read access to namespaces:

 ```yaml
  - apiGroups: [""]
    resources: ["namespaces"]
    verbs: ["get"]
```

Given a namespace:

 ```yaml
  apiVersion: v1
  kind: Namespace
  metadata:
    name: team-a
    labels:
      example.com/tier: gold
    annotations:
      example.com/quota-pool-id: pool-1234
```

A PVC created in team-a results in a CreateVolumeRequest whose parameters include:

```
csi.storage.k8s.io/namespace/annotations/example.com/quota-pool-id: pool-1234
csi.storage.k8s.io/namespace/labels/example.com/tier:               gold
```

The driver reads these parameters and forwards the pool id / tier to its backend. Keys configured but absent on the namespace are simply not added.

## **Which issue(s) this PR fixes**:

Fixes #

## **Special notes for your reviewer**:
- When either flag is set, the provisioner requires `get` on `namespaces`; a lookup failure reschedules the provision (`ProvisioningNoChange`) rather than silently dropping metadata.
- Injected keys use the existing `csi.storage.k8s.io/` parameter convention and are added after StorageClass parameter stripping, so they reach the driver.
- Only the configured keys present on the namespace are forwarded; absent keys are skipped.

## **Does this PR introduce a user-facing change?**:

```release-note
Added --extra-create-metadata-namespace-labels and --extra-create-metadata-namespace-annotations flags to forward configured PVC-namespace label/annotation values to drivers as CreateVolume parameters (under the csi.storage.k8s.io/namespace/labels/ and csi.storage.k8s.io/namespace/annotations/ prefixes). When set, the provisioner requires get access to namespaces.
```

### DIFF
--- a/cmd/csi-provisioner/csi-provisioner.go
+++ b/cmd/csi-provisioner/csi-provisioner.go
@@ -92,7 +92,10 @@ var (
 	strictTopology      = flag.Bool("strict-topology", false, "Late binding: pass only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.")
 	immediateTopology   = flag.Bool("immediate-topology", true, "Immediate binding: pass aggregated cluster topologies for all nodes where the CSI driver is available (enabled, the default) or no topology requirements (if disabled).")
 	extraCreateMetadata = flag.Bool("extra-create-metadata", false, "If set, add pv/pvc metadata to plugin create requests as parameters.")
-	enableProfile       = flag.Bool("enable-pprof", false, "Enable pprof profiling on the TCP network address specified by --http-endpoint. The HTTP path is `/debug/pprof/`.")
+
+	extraCreateMetadataNamespaceLabels      = flag.StringSlice("extra-create-metadata-namespace-labels", nil, "A list of PVC-namespace label keys whose values are added to plugin create requests as parameters under the csi.storage.k8s.io/namespace/labels/ prefix. Requires get access to namespaces. Works independently of --extra-create-metadata.")
+	extraCreateMetadataNamespaceAnnotations = flag.StringSlice("extra-create-metadata-namespace-annotations", nil, "A list of PVC-namespace annotation keys whose values are added to plugin create requests as parameters under the csi.storage.k8s.io/namespace/annotations/ prefix. Requires get access to namespaces. Works independently of --extra-create-metadata.")
+	enableProfile                           = flag.Bool("enable-pprof", false, "Enable pprof profiling on the TCP network address specified by --http-endpoint. The HTTP path is `/debug/pprof/`.")
 
 	defaultFSType = flag.String("default-fstype", "", "The default filesystem type of the volume to provision when fstype is unspecified in the StorageClass. If the default is not set and fstype is unset in the StorageClass, then no fstype will be set")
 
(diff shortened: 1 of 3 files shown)


jsafrane:
@eopeter what CSI driver would use these parameters? Do you have a concrete one that are you aiming at?

eopeter:
> what CSI driver would use these parameters? Do you have a concrete one that are you aiming at?

 Yes - the immediate consumer is an internal/proprietary CSI driver for a multi-tenant network block-storage backend. In that environment, a namespace belongs to a tenant / quota-pool on the backend, and that pool identity is recorded on the Namespace object (as a label/annotation) by the platform that provisions namespaces. The driver needs that identifier at `CreateVolume` time so it can place and account the volume against the correct pool/tenant on the backend.
  
Reasons the driver can't get this today:

- It's a per-namespace value, not per-StorageClass - one StorageClass serves many tenants, so it can't live in StorageClass parameters.
- The existing `--extra-create-metadata` gives the driver the PVC's name/namespace, but not the namespace's labels/annotations where the tenant identity actually lives.
- The alternative: having the driver itself read the Namespace - means granting the CSI controller cluster-wide get namespaces RBAC. The provisioner already runs with the scope to do this, so doing it here (and passing the values as parameters) keeps that privilege in one place, consistent with how `--extra-create-metadata` already forwards object metadata.

I kept the feature fully generic: the operator configures which label/annotation keys to forward, and they arrive under the existing csi.storage.k8s.io/namespace/{labels,annotations}/<key> parameter convention. So it's not specific to our driver - any driver backing a multi-tenant system that records tenant/project/quota identity on namespaces (a common pattern) can consume it.

jsafrane:
I can see it has an use case and it does not break any security barrier (at least I can't see it). I am a bit worried about adding a functionality that no public driver actually uses.

Anyway, it will need a bit more than this PR to get it merged:
* Please update README.md with new cmdline options.
* Does it need a similar change for VolumeAttributesClass?
* Add e2e tests to kubernetes/kubernetes, where the CSI sidecar tests actually reside. Since there is no CSI driver we can test with, an e2e test makes sure the feature stays working.
  * There is a [framework](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/csimock/) to deploy csi-driver-hostpath with hooks that allow an e2e test actually get a CSI call from the sidecar (or kubelet) and check it contains the right arguments. Please test that the provisioner translated the namespace annotation to a CreateVolume call correctly (with various cmdline options and namespace annotations).

Please submit a work-in-progress PR with the e2e test, feel free to use your personal external-provisioner image with this PR compiled. The test PR will need to wait for an official release of the external-provisioner with this PR, which will take few months.

Sorry about such huge requirements, but I'd like to have all PRs at least somewhat ready, so we can continue easily in case you disappear somewhere in the middle.

PR #1538, opened 2026-07-04
Filter capacity work items by storage class allowed topologies

### LINKED ISSUE #1162: Reduce CentralCapacityController workque depth by filtering allowed topologies in storage class
**Is your feature request related to a problem?/Why is this needed**

See `onSCAddOrUpdate` implementation in  https://github.com/kubernetes-csi/external-provisioner/blob/v4.0.0/pkg/capacity/capacity.go#L378-L384
```
	segments := c.topologyInformer.List()

	c.capacitiesLock.Lock()
	defer c.capacitiesLock.Unlock()
	for _, segment := range segments {
		c.addWorkItem(segment, sc)
	}
```

Currently the capacity controller creates a workqueue item for every `segment x storageClass` combination, some of which are redundant.

**Describe the solution you'd like in detail**

Storage classes can specify allowed topologies https://kubernetes.io/docs/concepts/storage/storage-classes/#allowed-topologies so not all combinations of `segment x storageClass` is valid here. An example would be certain types of storage (and hence storage class) is only available in selected topology segments.

We can reduce potential workqueue cardinality by filtering out impossible `segment x storageClass` combinations. This allows quicker `CSIStorageCapacity` object updates.

**Describe alternatives you've considered**

**Additional context**

DPS0340:
**What type of PR is this?**
/kind feature

The central capacity controller creates a work item (and with it a `CSIStorageCapacity` object) for every combination of topology segment and storage class, even when the storage class's `allowedTopologies` rule out that segment. This PR filters out those impossible combinations when a storage class is added/updated and when new topology segments appear, reducing work queue depth and the number of `CSIStorageCapacity` objects.

The check is intentionally conservative: a combination is only skipped when a topology selector term contains a requirement for one of the segment's keys with a value that is not allowed. Requirements using label keys which are not part of the segment are assumed to match, because the CSI driver's topology keys are not necessarily the same as the node label keys used in `allowedTopologies` — this avoids regressions for such setups.

When the `allowedTopologies` of an existing storage class are changed, work items that have become impossible are removed so that their `CSIStorageCapacity` objects get deleted.

Fixes #1162

- `onSCDelete` intentionally stays unfiltered: removing a work item that was never created is a cheap no-op, and this keeps deletion robust.
- New unit test `TestSegmentCompatibleWithStorageClass` covers the matching semantics (OR of terms, AND of requirements, unknown label keys), and `TestCapacityController` gained cases for initial filtering, filtering of newly added segments, non-segment label keys, and narrowing `allowedTopologies` on an update.

```release-note
The central storage capacity controller no longer creates CSIStorageCapacity objects for topology segments that are ruled out by a storage class's allowedTopologies.
```

### DIFF
--- a/pkg/capacity/capacity.go
+++ b/pkg/capacity/capacity.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"slices"
 	"sort"
 	"sync"
 	"time"
@@ -368,6 +369,10 @@ func (c *Controller) onTopologyChanges(added []*topology.Segment, removed []*top
 			continue
 		}
 		for _, segment := range added {
+			if !segmentCompatibleWithStorageClass(segment, sc) {
+				klog.V(5).Infof("Capacity Controller: skipping segment %s for storage class %s, not compatible with its allowed topologies", segment.SimpleString(), sc.Name)
+				continue
+			}
 			c.addWorkItem(segment, sc)
 		}
 		for _, segment := range removed {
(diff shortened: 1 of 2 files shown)


DPS0340:
@jsafrane thanks for the `/ok-to-test`. All jobs came back green afterwards:

```
unit                                    success
1-32 / 1-33 / 1-34 / 1-35 on k8s        success (4/4)
distributed on k8s 1-26 / 1-35          success (2/2)
tide                                    Not mergeable. Needs approved, lgtm labels.
```

So the only thing left is a review. @pohly, would you be able to take a look, or point me at whoever owns `pkg/capacity` these days? Not trying to rush anyone — just that tide is waiting on labels rather than on CI, and I would rather ask than let it sit.

I also confirmed it is still current: no conflicts against today's `master` (`git merge-tree` reports zero conflict markers), and `go test ./pkg/capacity/...` passes locally on the merge result. Happy to rebase if you would prefer a fresher base.

## The one decision worth a reviewer's attention

Everything else in the diff follows from this: **an unknown label key counts as a match, not as a mismatch.**

A segment is skipped only when a topology selector term has a requirement for one of *that segment's* keys with a value the segment does not have. If `allowedTopologies` mentions a key the segment does not carry at all, the term is treated as satisfied.

That is deliberately the conservative direction. CSI driver topology keys are not necessarily the same as the node label keys used in `allowedTopologies`, so treating an unknown key as a mismatch would silently stop producing `CSIStorageCapacity` objects for setups that work today. Treating it as a match preserves the current behaviour and only removes combinations that are provably impossible.

`TestSegmentCompatibleWithStorageClass` pins that semantics directly (OR across terms, AND within a term, unknown keys match), and `TestCapacityController` covers the paths where it matters: initial filtering, newly appearing segments, non-segment label keys, and narrowing `allowedTopologies` on an existing storage class so stale objects get deleted.

One asymmetry that is intentional and easy to read as an oversight: `onSCDelete` stays unfiltered. Removing a work item that was never created is a cheap no-op, and keeping deletion unconditional means a filtering bug can never leave an orphaned `CSIStorageCapacity` behind.

pkg/capacity/capacity.go
```diff
@@ -393,10 +398,69 @@ func (c *Controller) onSCAddOrUpdate(sc *storagev1.StorageClass) {
...
+	// The terms are ORed, so one compatible term is enough.
+	for _, term := range sc.AllowedTopologies {
+		if segmentCompatibleWithTerm(segment, term) {
+			return true
+		}
+	}
+	return false
+}
+
+func segmentCompatibleWithTerm(segment *topology.Segment, term v1.TopologySelectorTerm) bool {
```
jsafrane, line 440: `TopologySelectorTerm` says `A null or empty topology selector term matches no objects.`

In the code below, a term with no `MatchLabelExpressions` will skip `for` and it will go straight to `return true`. Which is IMO wrong. No `topology.Segment` can satisfy term with empty `term.MatchLabelExpressions`.

\+ please add an unit test for it.

jsafrane (commented):
Sorry it took so long. There is one issue, the PR looks pretty good otherwise.

PR #1552, opened 2026-07-29
Add snapshot topology support for Immediate binding (KEP-5943)

mdzraf:
**What type of PR is this?**

/kind feature

This PR introduces the changes necessary in the provisioner to be able to check for Volume Snapshot Topology and see if there is any intersection between the snapshot's `NodeAffinity` and `StorageClass.AllowedTopologies`.  This functionality is gated behind the `VolumeSnapshotTopology` feature gate.

This is part of the implementation for https://github.com/kubernetes/enhancements/issues/5943

/hold 

As noted in the TODO comment, the current implementation references a custom fork of the external snapshotter and will be switched to the released version of it once [the PR to introduce topology for volume snapshots is merged](https://github.com/kubernetes-csi/external-snapshotter/pull/1463).

```release-note
Introduce VolumeSnapshotTopology Feature Gate Along with Functionality to Check for Topology Intersection
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -695,12 +695,14 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 		},
 	}
 
+	var snapshotNodeAffinity []v1.TopologySelectorTerm
 	if dataSource != nil && (rc.clone || rc.snapshot) {
-		volumeContentSource, err := p.getVolumeContentSource(ctx, claim, sc, dataSource)
+		volumeContentSource, nodeAffinity, err := p.getVolumeContentSource(ctx, claim, sc, dataSource)
 		if err != nil {
 			return nil, controller.ProvisioningNoChange, fmt.Errorf("error getting handle for DataSource Type %s by Name %s: %v", dataSource.Kind, dataSource.Name, err)
 		}
 		req.VolumeContentSource = volumeContentSource
+		snapshotNodeAffinity = nodeAffinity
 	}
 
 	if dataSource != nil && rc.clone {
@@ -737,6 +739,19 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
 			return nil, controller.ProvisioningNoChange, fmt.Errorf("error generating accessibility requirements: %v", err)
 		}
 		req.AccessibilityRequirements = requirements
+
+		if utilfeature.DefaultFeatureGate.Enabled(features.VolumeSnapshotTopology) &&
+			rc.snapshot && selectedNodeName == "" && len(snapshotNodeAffinity) > 0 {
+			intersected := intersectSnapshotTopology(sc.AllowedTopologies, snapshotNodeAffinity)
+			if len(intersected) == 0 {
+				return nil, controller.ProvisioningFinished, fmt.Errorf("no compatible topology: snapshot %s NodeAffinity does not intersect StorageClass %s AllowedTopologies, so the volume cannot be provisioned where the snapshot is accessible", dataSource.Name, sc.Name)
+			}
+			req.AccessibilityRequirements = &csi.TopologyRequirement{
+				Requisite: intersected,
+				Preferred: intersected,
+			}
+			klog.V(2).Infof("prepareProvision: constrained snapshot-sourced PVC %s to %d topology terms from the intersection of snapshot NodeAffinity and StorageClass.AllowedTopologies", claim.Name, len(intersected))
+		}
 	}
 
 	// Resolve provision secret credentials.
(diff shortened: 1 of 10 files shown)


pkg/controller/controller.go
```diff
@@ -737,6 +739,19 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
...
+		if utilfeature.DefaultFeatureGate.Enabled(features.VolumeSnapshotTopology) &&
+			rc.snapshot && selectedNodeName == "" && len(snapshotNodeAffinity) > 0 {
+			intersected := intersectSnapshotTopology(sc.AllowedTopologies, snapshotNodeAffinity)
+			if len(intersected) == 0 {
+				return nil, controller.ProvisioningFinished, fmt.Errorf("no compatible topology: snapshot %s NodeAffinity does not intersect StorageClass %s AllowedTopologies, so the volume cannot be provisioned where the snapshot is accessible", dataSource.Name, sc.Name)
+			}
+			req.AccessibilityRequirements = &csi.TopologyRequirement{
+				Requisite: intersected,
+				Preferred: intersected,
+			}
```
xing-yang: This silently drops all the results generated by GenerateAccessibilityRequirements().

pkg/controller/controller.go
xing-yang, line 747: This will trigger a reschedule even in cases where the snapshot-topology intersection could have resolved things locally without needing that lookup.

pkg/controller/topology.go
```diff
@@ -692,6 +692,61 @@ func toCSITopology(terms []topologyTerm) []*csi.Topology {
...
+//     and a label expression's values are ORed. We flatten both inputs to the
+//     same disjunctive-normal-form (a list of fully-specified topologyTerms),
+//     then keep the flattened StorageClass terms that are a superset of (i.e.
+//     satisfy) at least one flattened snapshot term.
+//   - If the snapshot has no NodeAffinity, the StorageClass terms are returned
+//     unchanged (no additional constraint from the snapshot).
+//   - If the StorageClass has no AllowedTopologies, the snapshot terms are
+//     returned (the snapshot is the only constraint).
+//   - A nil, empty result means the two constraints are incompatible; callers
+//     should treat this as a fatal provisioning error.
```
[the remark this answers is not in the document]
mdzraf: Fixed in new rev

pkg/controller/controller.go
```diff
@@ -737,6 +739,19 @@ func (p *csiProvisioner) prepareProvision(ctx context.Context, claim *v1.Persist
...
+			klog.V(2).Infof("prepareProvision: constrained snapshot-sourced PVC %s to %d topology terms from the intersection of snapshot NodeAffinity and StorageClass.AllowedTopologies", claim.Name, len(intersected))
+		}
```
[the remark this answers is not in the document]
mdzraf: Addressed in new rev by adding `TestProvisionFromSnapshotWithTopology` tests

pkg/controller/controller.go
mdzraf: Good catch, fixed in new rev,  intersection is now fed into `GenerateAccessibilityRequirements` as its allowedTopologies input, so `GenerateAccessibilityRequirements` aggregation and preferred-ordering results are preserved and not dropped.

pkg/controller/controller.go
mdzraf: In the new rev an empty intersection fails with `ProvisioningFinished` so we never reach this line for that case. The `ProvisioningReschedule` here remains the pre-existing path for when `GenerateAccessibilityRequirements` can't find the node/CSINode object (apierrors.IsNotFound), which is unchanged by this PR.

pkg/controller/snapshot_topology_test.go
```diff
@@ -0,0 +1,133 @@
...
+			expected: []v1.TopologySelectorTerm{
+				singleZoneTerm("us-west-2a"),
+			},
+		},
+		"disjoint zones yield empty intersection": {
+			scTopology:   []v1.TopologySelectorTerm{zoneTerm("us-west-2d")},
+			snapTopology: []v1.TopologySelectorTerm{zoneTerm("us-west-2a", "us-west-2b", "us-west-2c")},
+			expected:     nil,
+		},
+		"snapshot zone within StorageClass region keeps the more specific snapshot term": {
```
jsafrane, line 89: Add a test with disjunct topology:
* SC wants: zone: us-east-2a  (user explicitly wants just a single zone)
* snapshot wants: region: us-east-2 (the snapshots are regional)

The final topology should require both - region: us-east-2 && zone: us-east-2a.

jsafrane, line 89: Add a test with "impossible" topology.
* SC wants: zone: us-west-1c (user explicitly wants just a single zone)
* snapshot wants: region: us-east-2 (the snapshots are regional)

The external provisioner should merge them in the same way as above - `region: us-west-1c && zone: us-east-2a`, because it has no knowledge about the real topology. It's IMO the driver's job to reject it.

pkg/controller/topology.go
```diff
@@ -692,6 +692,88 @@ func toCSITopology(terms []topologyTerm) []*csi.Topology {
...
+func intersectSnapshotTopology(scTopology, snapTopology []v1.TopologySelectorTerm) []v1.TopologySelectorTerm {
+	scTerms := flatten(scTopology)
+	snapTerms := flatten(snapTopology)
+
+	if len(snapTerms) == 0 {
+		return topologyTermsToSelectorTerms(scTerms)
+	}
+	if len(scTerms) == 0 {
+		return topologyTermsToSelectorTerms(snapTerms)
+	}
```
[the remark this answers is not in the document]
jsafrane: This looks like a valid comment to me.

pkg/controller/topology.go
```diff
@@ -692,6 +692,88 @@ func toCSITopology(terms []topologyTerm) []*csi.Topology {
...
+			// accessible from. Keep the more-specific StorageClass term.
+			if snapTerm.subset(scTerm) {
+				intersected = append(intersected, scTerm)
+			} else if scTerm.subset(snapTerm) {
+				// The snapshot term is more specific (e.g. SC allows a whole
+				// region, snapshot pins a zone within it). Keep the snapshot term.
+				intersected = append(intersected, snapTerm)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
jsafrane: Again, pretty good catch

pkg/controller/snapshot_topology_test.go
```diff
@@ -0,0 +1,133 @@
...
+				t.Errorf("intersectSnapshotTopology() mismatch (-got +want):\n%s",
+					cmp.Diff(got, tc.expected))
+			}
+		})
+	}
+}
+
+// TestIntersectSnapshotTopologyEmptyIsFatalSignal documents that an empty (but
+// non-nil) result is the signal callers use to fail provisioning fast when the
+// snapshot and StorageClass topologies are incompatible.
```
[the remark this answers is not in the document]
jsafrane: It's maybe valid, but much less concerning than the two AI comments above.

pkg/controller/snapshot_topology_test.go
mdzraf: Added in new rev ✅

mdzraf: Added in new rev ✅

